PR #7, opened 2016-10-19
fix to work in python 3

thdeltei:
in python 3, visual_default.keys() return a dict_keys object which does not have the 'remove' method. .iteritems() does not exist in Python3 anymore.


maartenbreddels:
thanks, I really need to get this tested on travis!

PR #31, opened 2017-07-10
Update installing.rst

rieder:
I would default to installing vaex as user, not as root. Modified docs accordingly.


maartenbreddels:
I'm not sure, maybe the wording isn't right. I think using anaconda or virtualenv should be the way to do it anyway, but maybe this is a better default, I'll merge it :)

rieder:
Maybe anaconda/virtualenv is better, but installing as root is definitely worse ;). Thanks for merging.

PR #35, opened 2017-08-22
Bug in coordinate conversion

JovanVeljanoski:
Hi,

I believe there is a bug in the coordinate transformations cartesian_to_spherical. It was a simple sign in the calculation of delta. Please verify again just to be sure.

Thank you,
JV.


maartenbreddels:
Thanks Jovan, I'd like to see some unittests before I merge this, but I'm happy to do this myself.

PR #54, opened 2017-12-15
new: added ds.select_circle/ellipse

JovanVeljanoski:
Added two new selection methods: one for circular and one for elliptical regions.

### DIFF
--- a/packages/vaex-core/vaex/test/dataset.py
+++ b/packages/vaex-core/vaex/test/dataset.py
@@ -2161,6 +2161,31 @@ class TestDataset(unittest.TestCase):
 		self.assertAlmostEqual(sumx, 0+1+2)
 		self.assertAlmostEqual(sumy, 0+1+4)
 
+	def test_select_circle(self):
+		# Generate test data
+		x = np.arange(-2,3)
+		x, y = np.meshgrid(x,x)
+
+		# add to dataset
+		ds = vx.from_arrays(x=x, y=y)
+		# do the selection
+		ds.select_circle('x', 'y', 0.5, 0.5, 1, name='circ')
+
+		# assert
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Could you add unittests for both? I'm moving to use py.test, so I'd say add tests/select_circle_test.py and include a test for the ellipse and the circle, sth like:
```
ds = vaex.from_scaler(x=1, y=2)
ds.select_circle(ds.x, ds.y, 2, 2)
assert ds.count(selection=True) == 1
```

packages/vaex-core/vaex/test/dataset.py
```diff
@@ -2161,6 +2161,31 @@ def test_lasso(self):
...
 
+	def test_select_circle(self):
+		# Generate test data
+		x = np.arange(-2,3)
+		x, y = np.meshgrid(x,x)
+
+		# add to dataset
+		ds = vx.from_arrays(x=x, y=y)
+		# do the selection
+		ds.select_circle('x', 'y', 0.5, 0.5, 1, name='circ')
```
maartenbreddels: pls use self.dataset, there are deliberate 'quircks' in them, such as missing data etc. It also has x and y.

packages/vaex-core/vaex/test/dataset.py
```diff
@@ -2161,6 +2161,31 @@ def test_lasso(self):
...
+
+		# assert
+		np.testing.assert_equal(4, ds.selected_length('circ'))
+
+	def test_select_circle(self):
```
maartenbreddels: The previous won't get executed since you've used the same name.

maartenbreddels:
travis osx failed for 1 machine, I think it's good to go!

PR #55, opened 2017-12-15
Added coordinate transformations: galactic to equatorial.

JovanVeljanoski:
Added galactic (l,b) to equatorial  (ra, dec) coordinate transformations.


maartenbreddels:
thanks 👍

PR #56, opened 2018-01-03
execfile workaround for python 3

dougalsutherland:
The `~/.vaex/vaex_import.py` script doesn't work in python 3, because `execfile` is gone. This replaces it with something that will work in either 2 or 3.


maartenbreddels:
Thanks again!

PR #72, opened 2018-04-05
Update colormaps.py

rieder:
spectral -> Spectral to make it work with recent matplotlib (not sure when this changed, but lowercase makes it fail for me)


maartenbreddels:
Thanks!

PR #94, opened 2018-06-02
Adding support to plot contours on top of the 2D histograms vaex produces.

JovanVeljanoski:
Using pylab.contour or pylab.contourf, adding support for displaying contours onto the 2D histograms vaex produces. 

I have tried to keep to how other vaex functions have been coded up. 

I hope this helps. 
(no unittests yet, but tested locally to some extent).

### DIFF
--- a/packages/vaex-viz/vaex/viz/contour.py
+++ b/packages/vaex-viz/vaex/viz/contour.py
@@ -1,0 +1,90 @@
+import re
+import vaex
+import numpy as np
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Excellent! Can you make it more like plot2d vector with the delayed stuff? That makes it possible to do many plots in parallel, like in 1 pass over the data.

maartenbreddels:
Thanks!

PR #99, opened 2018-06-29
Concat

JovanVeljanoski:
Added a vaex.concat method that concatenates a list of datasets. It utilizes the dataset.concat method for this. 

Also there is a unittest included.


maartenbreddels:
Since pandas does something similar, we'll stick to this api. Thanks

PR #100, opened 2018-07-16
Add a unit to a h5py dataset 

DougRzz:
Added a comment in the documentation which shows an example of adding an Astropy unit to a h5py dataset


maartenbreddels:
Thanks @DougRzz , however I went with #154 which is closer to my (unmerged) solution in #139.

DougRzz:
Yes, thats good. i noticed that better solution and i couldn't work out how to remove it on github. Glad it is sorted.

PR #116, opened 2018-11-01
Arrow/pyarrow support

maartenbreddels:
This allows converting from and to arrow Tables
```
ds = vaex.example()
table = ds.to_arrow_table()
ds_2 = vaex.from_arrow_table(table) # no memory copies
ds.export('test.arrow')
```

Reading should be fast (using memory mapping)
```
ds2 = vaex.open('test.arrow')
ds2.plot(ds2.x, ds2.y)
```

### DIFF
--- a/packages/vaex-arrow/vaex_arrow/convert.py
+++ b/packages/vaex-arrow/vaex_arrow/convert.py
@@ -1,0 +1,34 @@
+import pyarrow
+import numpy as np
+from .dataset import DatasetArrow
+
+def arrow_array_from_numpy_array(array):
+    dtype = array.dtype
+    mask = None
+    if np.ma.isMaskedArray(array):
+        mask = array.mask
+    if dtype.kind == 'S':
+        type = pyarrow.binary(dtype.itemsize)
+        arrow_array = pyarrow.array(array, type, mask=mask)
+    else:
+        if dtype.isnative:
+            arrow_array = pyarrow.array(array, mask=mask)
+        else:
+            # TODO: we copy here, but I guess we should not... or give some warning
+            arrow_array = pyarrow.array(array.astype(dtype.newbyteorder('=')), mask=mask)
+    return arrow_array
+
+
+def arrow_table_from_vaex_dataset(ds, column_names=None, selection=None, strings=True, virtual=False):
+    """Implementation of Dataset.to_arrow_table"""
+    names = []
+    arrays = []
+    for name, array in ds.to_items(column_names=column_names, selection=selection, strings=strings, virtual=virtual):
+        names.append(name)
+        arrays.append(arrow_array_from_numpy_array(array))
+    # import IPython
+    # IPython.embed()
+    return pyarrow.Table.from_arrays(arrays, names)
+
+def vaex_dataset_from_arrow_table(table):
+    return DatasetArrow(table=table)
--- a/packages/vaex-arrow/vaex_arrow/convert.py
+++ b/packages/vaex-arrow/vaex_arrow/convert.py
@@ -1,6 +1,5 @@
 import pyarrow
 import numpy as np
-from .dataset import DatasetArrow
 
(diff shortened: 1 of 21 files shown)


packages/vaex-arrow/vaex_arrow/convert.py
```diff
@@ -0,0 +1,34 @@
+import pyarrow
+import numpy as np
+from .dataset import DatasetArrow
+
+def arrow_array_from_numpy_array(array):
+    dtype = array.dtype
+    mask = None
+    if np.ma.isMaskedArray(array):
```
xhochy: We should support this natively in Arrow. Can you open a JIRA there?

packages/vaex-arrow/vaex_arrow/dataset.py
```diff
@@ -0,0 +1,81 @@
...
+            buffers = arrow_array.buffers()
+            assert len(buffers) == 2
+            mask = None
+            bitmap_buffer = buffers[0]
+            data_buffer = buffers[1]
+            if bitmap_buffer is not None:
+                # arrow uses a bitmap https://github.com/apache/arrow/blob/master/format/Layout.md
+                bitmap = np.frombuffer(bitmap_buffer, np.uint8, len(bitmap_buffer))
+                # we do have to change the ordering of the bits
+                mask = 1-np.unpackbits(bitmap).reshape((len(bitmap),8))[:,::-1].reshape(-1)[:len(arrow_array)]
```
xhochy: Is this just changing the ordering or also doing the conversion from bit to byte mask?

packages/vaex-arrow/vaex_arrow/dataset.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _load(self):
+        source = pa.memory_map(self.path)
+        reader = pa.open_stream(source)
+        table = pa.Table.from_batches([b for b in reader])
+        self._load_table(table)
+    
+    def _load_table(self, table):
+        self._length_unfiltered =  self._length_original = table.num_rows
+        for col in table.columns:
```
xhochy: This function converts arrow arrays to numpy masked arrays or does it more?

packages/vaex-arrow/vaex_arrow/convert.py
maartenbreddels: I think so too, I wanted to move fast, but I think many of these features should go into pyarrow.

packages/vaex-arrow/vaex_arrow/dataset.py
maartenbreddels: Yeah, arrow seems to use this layout https://github.com/apache/arrow/blob/master/format/Layout.md#null-bitmaps which is different than numpy's (un)packbits. Would be nice if pyarrow provides a function for this (or numpy gets extended)

packages/vaex-arrow/vaex_arrow/convert.py
xhochy: Feel free to just hammer at us with JIRAs. We're happy for them; we're even happier when we get PRs for them but just JIRAs are also helpful to see what the missing pieces are that people actually need.

packages/vaex-arrow/vaex_arrow/dataset.py
maartenbreddels: Yes, basically does that. I need to refactor it a bit, move some of the code to convert.py. But again, it would be nice of pyarrow does all of this.

maartenbreddels:
Thanks

> 👨‍🚀astronomical to see this!

👍

maartenbreddels:
Ref to Jira issues resulted from this:
 * https://issues.apache.org/jira/browse/ARROW-3685
 * https://issues.apache.org/jira/browse/ARROW-3686
 * https://issues.apache.org/jira/browse/ARROW-3669

Issue #113, opened 2018-10-29
fillna crashes


datascienceit:
```
import vaex as vx
ds = vx.open('../datasets/train.csv')
ds['Cabin'] = ds['Cabin'].fillna('NA')
ds['Cabin']
```

Result:
```
<vaex.expression.Expression(expressions='Cabin')> instance at 0x11454beb8 values=[Error evaluating: NameError("name 'NA' is not defined",)] 
```

[dataset](https://www.kaggle.com/c/titanic/data)

maartenbreddels:
It is exposing a bug, you can try this workaround for now:
```
ds = vaex.open('../train.csv')
ds.columns['Cabin'] = ds.columns['Cabin'].astype(str)
ds['Cabin'] = ds.apply(lambda x: 'missing' if x == 'nan' else x, ['Cabin'])
```

PR #121, opened 2018-11-16
Fillna test

JovanVeljanoski:
Updating the test for fillna which exposes issue #113 (happens for columns of dtype 'object' only).


maartenbreddels:
Awesome!

PR #123, opened 2018-11-17
Add .values properties to Dataset

JovanVeljanoski:
When multiple expressions are considered, a decision needs to be made how missing values will be treated.

I propose either missing values to be converted to `np.nan`, or the whole resulting `nd.array` to be case to a masked array.

### DIFF
--- a/tests/values_test.py
+++ b/tests/values_test.py
@@ -10,6 +10,7 @@ def test_values(ds_local):
     assert ds['obj'].values.data.tolist() == ds.evaluate('obj').data.tolist()
     assert ds[['x', 'y']].values.tolist() == np.array([ds.evaluate('x'), ds.evaluate('y')]).T.tolist()
     assert ds[['x', 'y']].values.shape == (len(ds), 2)
+    assert ds[['m']].values[7][0] == 77777.0
     assert ds[['m','x']].values[7][0] == 77777
     # The missing values are included. This may not be the correct behaviour
     assert ds[['x', 'y', 'nm']].values.tolist(), np.array([ds.evaluate('x'), ds.evaluate('y'), ds.evaluate('nm')]).T.tolist()


tests/values_test.py
```diff
@@ -0,0 +1,18 @@
+from common import *
+
+
+def test_values(ds_local):
+    ds = ds_local
+
+    np.testing.assert_array_equal(ds['x'].values, np.array(ds.evaluate('x')).T)
+    np.testing.assert_array_equal(ds['name'].values, np.array(ds.evaluate('name')).T)
+
```
maartenbreddels: This already works right, and no need to transpose

tests/values_test.py
```diff
@@ -0,0 +1,18 @@
...
+    np.testing.assert_array_equal(ds['obj'].values.mask, ds.evaluate('obj').mask)
+    ind = ds['obj'].values.data == ds.evaluate('obj').data
```
maartenbreddels: Ind? You mean mask?

tests/values_test.py
```diff
@@ -0,0 +1,18 @@
...
+    assert np.concatenate((ind[:5], ind[6:])).all()
```
maartenbreddels: What does this test?

tests/values_test.py
```diff
@@ -0,0 +1,18 @@
...
+    assert np.isnan(ds['obj'].values.data[5])
+
+    np.testing.assert_array_equal(ds[['x', 'y']].values, np.array(ds.evaluate('x, y')).T)
```
maartenbreddels: This use of evaluate is not supported, and accidentally works

tests/values_test.py
JovanVeljanoski: Yeah, sorry, this test was written later, and I was just copy pasting code. I will improve it in the next iteration.

tests/values_test.py
JovanVeljanoski: `ind` was meant to be "index" which i use for the next test, but maybe `mask` is indeed a better name.

JovanVeljanoski:
A potential issue may come when columns of different dtypes are exposed as a single array. I guess in that case the dtype of the resulting array should be `np.object`.

Ok I will try to implement this and let's see how it goes.

maartenbreddels:
Not object, check that method I mentioned, it will find the best dtype using numpy, or float64, the idea is to pass it to sklearn or similar packages, so I think this is behaviour is kind of expected

JovanVeljanoski:
I attempted to implement this. Please note that if any of the columns have masked values, with the current implementation they will be propagated to the resulting ndarray. 

Is this the expected or required behaviour?

packages/vaex-core/vaex/dataset.py
```diff
@@ -5754,3 +5754,7 @@ def add_column(self, name, data):
         super(DatasetArrays, self).add_column(name, data)
         self._length_unfiltered = int(round(self._length_original * self._active_fraction))
         # self.set_active_fraction(self._active_fraction)
+
+    @property
+    def values(self):
+        return np.array([self[column].values for column in self.column_names]).T
```
maartenbreddels: Np.array(zelf) is what I meant.
This will call dataset.__array__

tests/values_test.py
```diff
@@ -0,0 +1,23 @@
...
+    assert np.concatenate((ind[:5], ind[6:])).all()
+    assert np.isnan(ds['obj'].values.data[5])
+
+    np.testing.assert_array_equal(ds[['x', 'y']].values, np.array([ds.evaluate('x'), ds.evaluate('y')]).T)
+    # The missing values are included. This may not be the correct behaviour
+    np.testing.assert_array_equal(ds[['x', 'y', 'nm']].values, np.array([ds.evaluate('x'), ds.evaluate('y'), ds.evaluate('nm')]).T)
+
+def test_object_column_values(ds_local):
+    ds = ds_local
+    with pytest.raises(Exception) as e_info:
```
maartenbreddels: You can drop the `as e_info`.
What kind of exception? ValueError, this is not specific enough.

tests/values_test.py
```diff
@@ -0,0 +1,23 @@
...
+    np.testing.assert_array_equal(ds['name'].values, np.array(ds.evaluate('name')))
+
+    np.testing.assert_array_equal(ds['obj'].values.mask, ds.evaluate('obj').mask)
+    ind = ds['obj'].values.data == ds.evaluate('obj').data
+    assert np.concatenate((ind[:5], ind[6:])).all()
+    assert np.isnan(ds['obj'].values.data[5])
+
+    np.testing.assert_array_equal(ds[['x', 'y']].values, np.array([ds.evaluate('x'), ds.evaluate('y')]).T)
+    # The missing values are included. This may not be the correct behaviour
+    np.testing.assert_array_equal(ds[['x', 'y', 'nm']].values, np.array([ds.evaluate('x'), ds.evaluate('y'), ds.evaluate('nm')]).T)
```
maartenbreddels: `assert ds[['x', 'y', 'nm']].values.T.tolist() == [ds.x.tolist(), ds.y.tolist()]`
is much easier on the eyes don't you think?
I have a hard time reading it as it is now, feel gree to disagree.

packages/vaex-core/vaex/dataset.py
```diff
@@ -5754,3 +5754,13 @@ def add_column(self, name, data):
...
         self._length_unfiltered = int(round(self._length_original * self._active_fraction))
         # self.set_active_fraction(self._active_fraction)
+
+    @property
+    def values(self):
+        """
+        Return a numpy representation of the Dataset.
+        Only the values in the Dataset are returned, the column labels are removed.
+
+        Does not preserve the mask if any of the converted columns are masked arrays.
```
maartenbreddels: This docstring should be the same as `__array__`, or refer to it.  Also, always start directly after """ with see https://www.python.org/dev/peps/pep-0257/. Now you start on the second, so it will not show up with a simple ds.values?
In general I think sth like this should be good:
```
"""Gives a full memory copy of the dataset into a 2d numpy array of shape (#rows, #columns).
Note that the memory order is fortran, so all values of 1 column are contiguous in memory, for performance reasons

Note this returns the same result as:

>>> np.array(ds)

Except this can optionally take a dtype argument

Masked values are replaced by NaN's when the dtype if float, for other types the behaviour is undefined.
"""

And update the `__array__` docstring accordingly

File: tests/values_test.py (hunk 3 above)
Comment (maartenbreddels) [reply tests/values_test.py]: Again, what does this test?

File: tests/values_test.py [hunk 9]
```diff
@@ -0,0 +1,23 @@
...
+
+    np.testing.assert_array_equal(ds['x'].values, np.array(ds.evaluate('x')))
+    np.testing.assert_array_equal(ds['name'].values, np.array(ds.evaluate('name')))
+
+    np.testing.assert_array_equal(ds['obj'].values.mask, ds.evaluate('obj').mask)
+    ind = ds['obj'].values.data == ds.evaluate('obj').data
+    assert np.concatenate((ind[:5], ind[6:])).all()
+    assert np.isnan(ds['obj'].values.data[5])
+
+    np.testing.assert_array_equal(ds[['x', 'y']].values, np.array([ds.evaluate('x'), ds.evaluate('y')]).T)
```
maartenbreddels: Maybe explicitly test the shape?
```python
assert ds[['x', 'y']].values.shape == (len(ds), 2)
```

packages/vaex-core/vaex/dataset.py
```diff
@@ -5053,10 +5053,16 @@ def __call__(self, *expressions, **kwargs):
...
+        """Gives a full memory copy of the dataset into a 2d numpy array of shape (n_rows, n_columns).
+        Note that the memory order is fortran, so all values of 1 column are contiguous in memory for performance reasons.
 
-        Example:
-        >>> ar = np.array(ds)
+        Note this returns the same result as:
+
+        >>> np.array(ds)
+
+        except this can optionally take a dtype argument
```
maartenbreddels: Remove this line here, since this one does take dtype=

packages/vaex-core/vaex/dataset.py
```diff
@@ -5754,3 +5760,18 @@ def add_column(self, name, data):
...
+        """Gives a full memory copy of the dataset into a 2d numpy array of shape (n_rows, n_columns).
+        Note that the memory order is fortran, so all values of 1 column are contiguous in memory for performance reasons.
+
+        Note this returns the same result as:
+
+        >>> np.array(ds)
+
+        except this can optionally take a dtype argument
+
+        Masked values are replaced by NaN's when the dtype if float, for other types the behaviour is undefined.
```
maartenbreddels: Is this true, if so, we need to test it, if not, we should update this.

packages/vaex-core/vaex/dataset.py
```diff
@@ -5053,10 +5053,14 @@ def __call__(self, *expressions, **kwargs):
...
+        """Gives a full memory copy of the dataset into a 2d numpy array of shape (n_rows, n_columns).
+        Note that the memory order is fortran, so all values of 1 column are contiguous in memory for performance reasons.
 
-        Example:
-        >>> ar = np.array(ds)
+        Note this returns the same result as:
+
+        >>> np.array(ds)
+
+        If any of the columns contain masked arrays, the masks are ignored (i.e. the masked elements are returned as well).
```
maartenbreddels: My point was, whatever we say here, should be reflected in a unittests, now we are free to change this without hitting a failure at the unittests.

tests/values_test.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+from common import *
+
+
+def test_values(ds_local):
+    ds = ds_local
+
+    assert ds['x'].values.tolist() == np.array(ds.evaluate('x')).tolist()
```
maartenbreddels: Why the cast to np.array on the rhs, it should return an array already right?

PR #124, opened 2018-11-17
Adding unit-tests for the .dtype() method and a proposal for dtypes().

JovanVeljanoski:


tests/dtypes_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def test_dtypes(ds_local):
+  assert ds.dtype(ds.x) == np.float64
+  assert ds.dtype(ds.y) == np.float64
+  assert ds.dtype(ds.m) == np.float64
+  assert ds.dtype(ds.nm) == np.float64
+  assert ds.dtype(ds.mi) == np.int64
+  assert ds.dtype('ints') == np.int64
+  assert ds.dtype('name') == 'S25'
```
maartenbreddels: This seems inconsistent with the rest right

JovanVeljanoski: Which one? the 'S25' type? Maybe I am wrong, but manual checks indicated that this is the dtype of the underlying array (or 'S|25' )

maartenbreddels:
Also would be nice to split it in 2 tests, since you now have two tests in 1 function, better 2 functions or more in 1 file

tests/dtypes_test.py
```diff
@@ -0,0 +1,20 @@
...
+  assert ds.dtype('obj') == np.object
+
+def test_dtypes(ds_local):
+  ds = ds_local
+  all_dtypes = np.array([np.float64, np.float64, np.float64, np.float64, np.int64, np.int64, 'S25', np.object])
+  np.testing.assert_array_equal(ds.dtypes(columns=None), all_dtypes)
```
maartenbreddels: On writing tests, don't use np.array for non-numerical stuff, it's not idiomatic Python, better/simpler would be:
```
all_dtypes = [np.float64, ....]
ds.dtypes(None) == all_dtypes
```
Less code, easier to read, and the use of numpy only confuses.

tests/dtypes_test.py
maartenbreddels: Theses tests will never pass, Pandas will do the same, see https://stackoverflow.com/questions/26921836/correct-way-to-test-for-numpy-dtype

tests/dtypes_test.py
maartenbreddels: This has nothing to do with numpy, so I recommend just use assert (otherwise it implies some numerical equality, which is not tested here)
assert ds.dtypes(columns=None).values == all_dtypes

maartenbreddels:
Thanks, awesome work.

PR #136, opened 2018-11-29
WIP: lazy in polygon(s) test, for e.g. massive geojson regions

maartenbreddels:
This needs a bit of refactoring and/or benchmarking, the last function (inside_which_polygons) is using some precompute functionality, which may need some polishing. It also needs benchmarking to see if it is significantly faster without precompute.
If good, the other *_polygon functions should behave similar.


maartenbreddels:
Replaced by https://github.com/vaexio/vaex/pull/685

PR #139, opened 2018-11-29
fix appveyor by getting more deps from conda-forge

maartenbreddels:


maartenbreddels:
Thanks for the help, but the import should not have been top level. The py27 failure seems because arrow is not supported for python27 on windows it seems.

DougRzz:
Apologies, my comment wasn't directly related to this issue. It was more of a general comment for Vaex in Windows with the fcntl module.

maartenbreddels:
Ah ok, then I misunderstood. Hope to get windows on board again!

maartenbreddels:
windows support back in by #117

PR #144, opened 2018-12-05
a cleaner shortcut to read a dictionary 

xdssio:
I recommend to use it in docs as well as it easier to consume

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -243,6 +243,20 @@ def from_astropy_table(table):
     return vaex.file.other.DatasetAstropyTable(table=table)
 
 
+def from_dict(input):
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/__init__.py
```diff
@@ -243,6 +243,20 @@ def from_astropy_table(table):
     return vaex.file.other.DatasetAstropyTable(table=table)
 
 
+def from_dict(data):
+    """Create an in memory dataset from numpy arrays, in contrast to from_arrays this keeps the order of columns intact
```
JovanVeljanoski: Please update the Docstring to reflect what the function does. 
Since the latest release, we have a established way of writing the Docstrings. Please rebase your branch to master, and look at the other examples. 

Otherwise it looks good

PR #158, opened 2018-12-17
Adding docstrings to DataFrame.split and DataFrame.split_random. Also…

JovanVeljanoski:
… adding a default value to the frac keyword.

The the fraction default (`frac=0.8`) is as it is commonly seen in for train/test split in popular libraries.

### DIFF
--- a/packages/vaex-core/vaex/utils.py
+++ b/packages/vaex-core/vaex/utils.py
@@ -829,7 +829,7 @@ def _split_and_combine_mask(arrays):
 			mask |= other
 	return arrays, mask
 
-def listify(fn=None, wrapper=list):
+def gen_to_list(fn=None, wrapper=list):
     '''A decorator which wraps a function's return value in ``list(...)``.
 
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/utils.py
```diff
@@ -829,3 +829,38 @@ def _split_and_combine_mask(arrays):
...
+    :rtype: list or tuple
+    '''
+    def listify_return(fn):
+        @functools.wraps(fn)
+        def listify_helper(*args, **kw):
+            return wrapper(fn(*args, **kw))
+        return listify_helper
+    if fn is None:
+        return listify_return
+    return listify_return(fn)
```
maartenbreddels: missing newline

JovanVeljanoski: Fixed

JovanVeljanoski:
Modified the new function to `gen_to_list` to avoid namespace conflicts with the existing `listify` function.

PR #160, opened 2018-12-21
Updated the limits function to use the latest is_category instead of …

JovanVeljanoski:
…the deprecated iscategory method.


maartenbreddels:
thanks!

PR #161, opened 2018-12-23
Added isnan, isfinite

DougRzz:
Added support for gradient, isnan, isfinite and roll from NumPy.
Also added rolling sum and rolling mean using Pandas


maartenbreddels:
Nice, although the rolling cannot work. Vaex does this in chunks, and the kernel therefore misses the info from the next chunk. We would need overlapping chunks to really support this.
I also wonder why one would use this with large datasets (or do you mean to use this for small datasets?). Maybe a better way to do is is to do it on the binned data?

DougRzz:
I did wonder whether there might be a problem with the rolling calcs at the edges of each bin.  I guess the gradient calc won't be good either at the bin edges.  However, visually they all  looked quite good. Thanks for the warning, I will check the output again.  If its not 100% correct, then it shouldn't go into Vaex-core

Here is an example of performing a combination of binning and rollingMean on time series data  (btw you might also notice the use of the really cool ipyVolume debounce function).  I hope the 10MB animated gif is not a problem.  The red line is the rolling mean and the blue line in the original channel.  I have binned each channel with 600 chunks and calculated the mean, min and max for each bin.    For the rolling mean channel, i used this expression  `rollingMean(where(isfinite(channelName),channelName,0),30)` . The rolling mean is performed over a window length of 30.   I know i shouldn't replace nans with zeros for a correct rolling mean, its just an example.

zoomin_rollingmean_orig

Does this example demonstrate what I am trying to do?  Other types of aggregation might be better than rolling mean in this case. But rolling window and gradients in Vaex would be a potentially really powerful tool for time series analysis.

maartenbreddels:
Don't merge master in this branch please, you should rebase it on master.
Or we just merge it as is right?

(from mobile phone)

JovanVeljanoski:
Yeah my bad, please rebase it so that the tests can run again.

DougRzz:
Hi @JovanVeljanoski,  I am not too familiar with Git.  Do I have to do the rebasing on my repo or is that for a vaex.io admin to do on this repo?  If its my job,  I will figure out how to do that this weekend.

maartenbreddels:
Thanks for you patience @DougRzz ! :)

PR #164, opened 2018-12-27
apply on an expression

xdssio:
When run apply on an Expression, it run it through the dataframe

### DIFF
--- a/tests/apply_test.py
+++ b/tests/apply_test.py
@@ -3,4 +3,4 @@ import vaex
 
 def test_apply():
     ds = vaex.from_arrays(**{'A': [1, 2, 3]})
-    assert ds['A'].apply(lambda x: x + 1) == [2, 3, 4]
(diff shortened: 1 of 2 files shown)


tests/apply_test.py
```diff
@@ -0,0 +1,6 @@
+import vaex
+
+
+def test_apply():
+    ds = vaex.from_arrays(**{'A': [1, 2, 3]})
+    assert ds['A'].apply(lambda x: x + 1) == [2, 3, 4]
```
maartenbreddels: the left hand side is an expression, comparing that with a list, gives a new expression, You probably want to test:
```python
 assert ds['A'].apply(lambda x: x + 1).values.tolist() == [2, 3, 4]
```
I'm ok with skipping the docstring for now (since it does not have any it does not show up with sphinx). I think df.apply should be made more automatic, where you don't need to explicitly give the arguments.

maartenbreddels:
great, thanks all!

PR #125, opened 2018-11-17
value_counts support

maartenbreddels:
This is a start for value_counts, only works/tested for 'integer' data. 
TODO:
 - [x] test with strings
 - [x] test with float values 0.1, etc
 - [x] consider refactoring/renaming label_encoding/catergorise
 - [ ] test the changes in catergorise

PR #165, opened 2018-12-27
Value counts

xdssio:
Using map reduce for value_counts.
* issue with strings
* inefficient move from dict to vaex


maartenbreddels:
dup of #125, which should be more efficient, I'll close this PR and see if #125 covers everything in this PR once I work on it again.

PR #175, opened 2019-01-30
New: experimental expression graph access methods

maartenbreddels:
This can be used to gain insight into the virtual expression, but its main usecase now is to be able to run refit for vaex-ml
Example:
screenshot 2019-01-28 at 09 18 25


maartenbreddels:
Considering the leading underscore, we can merge this, and consider it private/internal first.

PR #151, opened 2018-12-14
from_dict

xdssio:
Read a dict to a DataFrame

PR #177, opened 2019-02-04
change ui import due to dataframe refactor, fixes #173

### LINKED ISSUE #173: ImportError: cannot import name 'Task'
https://github.com/vaexio/vaex/blob/f654e895e48a1c5080ad9f84df9a2c31c7add64a/packages/vaex-ui/vaex/ui/layers.py#L2740

Sorry but this version does not find Task in `vaex.dataset`. Should this be imported from tasks instead?

maartenbreddels:


maartenbreddels:
Test fail unrelated, introduced in #151 
I guess the test fails because dict order is not guaranteed, would be great if you could fix the test @xdssio

PR #180, opened 2019-02-16
Adding a map method to the vaex expressions, which maps new values ov…

JovanVeljanoski:
…er the existing ones according to a corresponding dict or a mapping function.

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -3,6 +3,7 @@ import cloudpickle as pickle
 import functools
 import operator
 import six
+import collections.abc
 
 from future.utils import with_metaclass
 import numpy as np
@@ -510,8 +511,8 @@ def f({0}):
     def apply(self, f):
         return self.ds.apply(f, [self.expression])
 
-    def map(self, mapper):
-        '''Map values of an expression or in memory column accoring to an input
+    def map(self, mapper, missing_values=None):
+        """Map values of an expression or in memory column accoring to an input
         dictionary or a custom callable function.
 
         :param mapper: (dict or function) used to map the values on top of the existing ones
@@ -526,20 +527,13 @@ def f({0}):
         >>> df['color_mapped'] = df.color.map(mapper)
         >>> df.color_mapped.values
         array([1, 1, 2, 1, 3])
-        '''
-        if isinstance(mapper, dict):
+        """
+        if isinstance(mapper, collections.abc.Mapping):
             def lookup_func(x):
-                if isinstance(x, np.bytes_):
-                    x = x.decode('utf-8')
-                    if x in mapper.keys():
-                        return mapper[x]
-                    else:
-                        return np.nan
+                if x in mapper:
+                    return mapper[x]
                 else:
-                    if x in mapper.keys():
-                        return mapper[x]
-                    else:
-                        return np.nan
+                    return missing_values
             return self.ds.apply(lookup_func, [self.expression])
         elif callable(mapper):
             return self.ds.apply(mapper, [self.expression])
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -503,12 +503,48 @@ def f({0}):
                 return new
         expr = expresso.translate(self.expression, translate)
         return Expression(self.ds, expr)
- 
+
     def astype(self, dtype):
         return self.ds.func.astype(self, str(dtype))
 
     def apply(self, f):
         return self.ds.apply(f, [self.expression])
+
+    def map(self, mapper):
+        '''Map values of an expression or in memory column accoring to an input
+        dictionary or a custom callable function.
+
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+                    x = x.decode('utf-8')
+                    if x in mapper.keys():
+                        return mapper[x]
+                    else:
+                        return np.nan
+                else:
+                    if x in mapper.keys():
+                        return mapper[x]
+                    else:
+                        return np.nan
```
maartenbreddels: we should not return nan, nan is only for numerical values. return None instead if we have to, even better would be to missing_value argument is map.

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+        >>> mapper = {'red': 1, 'blue': 2, 'green': 3}
+        >>> df['color_mapped'] = df.color.map(mapper)
+        >>> df.color_mapped.values
+        array([1, 1, 2, 1, 3])
+        '''
+        if isinstance(mapper, dict):
+            def lookup_func(x):
+                if isinstance(x, np.bytes_):
+                    x = x.decode('utf-8')
+                    if x in mapper.keys():
```
maartenbreddels: see below

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+                        return mapper[x]
+                    else:
+                        return np.nan
```
maartenbreddels: see below

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+                else:
+                    if x in mapper.keys():
```
maartenbreddels: if x in mapper: it will call mapper.__contains__ which can be (and will be) more efficient and flexible.

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
- 
+
     def astype(self, dtype):
         return self.ds.func.astype(self, str(dtype))
 
     def apply(self, f):
         return self.ds.apply(f, [self.expression])
 
+    def map(self, mapper):
+        '''Map values of an expression or in memory column accoring to an input
```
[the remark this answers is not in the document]
JovanVeljanoski: I find the  ''' more easy on the eye :). But ok, changed!

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+        >>> import vaex
+        >>> df = vaex.from_arrays(color=['red', 'red', 'blue', 'red', 'green'])
+        >>> mapper = {'red': 1, 'blue': 2, 'green': 3}
+        >>> df['color_mapped'] = df.color.map(mapper)
+        >>> df.color_mapped.values
+        array([1, 1, 2, 1, 3])
+        '''
+        if isinstance(mapper, dict):
+            def lookup_func(x):
+                if isinstance(x, np.bytes_):
```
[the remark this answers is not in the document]
JovanVeljanoski: Indeed! Very happy to remove it.

packages/vaex-core/vaex/expression.py
JovanVeljanoski: Ok added `missing_values=None` argument. I've updated the tests, so in principle one could put `missing_values=np.nan` is that is required.

packages/vaex-core/vaex/expression.py
```diff
@@ -460,13 +460,49 @@ def translate(id):
...
+        Example:
+
+        >>> import vaex
+        >>> df = vaex.from_arrays(color=['red', 'red', 'blue', 'red', 'green'])
+        >>> mapper = {'red': 1, 'blue': 2, 'green': 3}
+        >>> df['color_mapped'] = df.color.map(mapper)
+        >>> df.color_mapped.values
+        array([1, 1, 2, 1, 3])
+        '''
+        if isinstance(mapper, dict):
```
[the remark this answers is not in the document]
JovanVeljanoski: Done

packages/vaex-core/vaex/expression.py
JovanVeljanoski: Done

JovanVeljanoski:
I find this potential issue:

```
import vaex
import numpy as np

x = np.array(['a', 'a', 'a', 'b', 'b', 'c'], dtype=np.object)
ds = vaex.from_arrays(x=x)

ds.x.value_counts()
```

that returns
```
Series([], dtype: float64)
```

For big datasets this also sometimes kills the kernel, although I always reproduce it.

maartenbreddels:
This is replaced by #197

PR #186, opened 2019-03-02
Improved string support based on arrow

maartenbreddels:
This is the start of much better string support, which is based/compatible with arrow (which gives some issues).

Vaex already supported a Column (besides plain numpy arrays). The objects need to mimic a numpy array (i.e. have a .dtype/.shape/len(column)/__getitem__(slice/indices)). The ColumnStringArrow is a subclass of Column and own a byte buffer and an index buffer that the arrow format specifies.
The column object can convert it to an list of str objects on the fly (actually passed as an ndarray with dtype=object now). 

Exporting these columns is fully supported (the arrow path may not be optimal yet arrow->array->arrow), but the hdf5 export also supports it. This means we could pass the columns between vaex and arrow with 0 memory copies.

Changes / issues
 * BIG issue: arrow uses int32 for the indices, meaning a single column can only have 2 billion chars! This is unacceptably low for any of the large datasets we use. We should also see how arrow handles this. A solution for vaex could be:
   * Break compatibility with arrow, and upcast to int64.
   * Split into multiple arrays. We have concatenated columns already, but it will lead to a performance penalty. (on a related note, we should support chunking)
 * If name is a string column, df.name.dtype == str (meaning that dtype is not always a numpy dtype, which makes code complex, but we can think about other objects now as well, images?)
 * an ndarray with the first element having type str will be reported of having a str dtype, whatever the rest of the values are, maybe we should rethink this. It does means that of you pass an ndarray of dtype=object with strings in it and export it to hdf5, it will be saved in the (efficient) arrow format.
 * arrays with strings are not converted to bytes anymore (avoiding the b'name' printing, which looks ugly, and is annoying)
 * a lot of code is a pure Python loop, this is a good reason to start with a new string extension module using pybind11 (and maybe xtensor) to do the hard work in C++, and also implement things like strlen and maybe regular expression functionality (and do it in C++ to get real multithreading).

To use
```python
# this forces vaex to ask arrow to convert it to a real arrow structure
x = vaex.string_column(['aap', 'noot', 'mies'])
df = vaex.from_arrays(x=x)
df['xr'] = df.x.apply(lambda x: x[::-1] + " reversed")
df
```


xhochy:
> BIG issue: arrow uses int32 for the indices, meaning a single column can only have 2 billion chars! This is unacceptably low for any of the large datasets we use. We should also see how arrow handles this.

We also have some code in Arrow when we construct string arrays that automatically splits them into mutliple ones (mostly in the Parquet handling code). So our path there is to use `ChunkedArray` instances.

I will have a look later through the PR but as always, feel free to ping me and also it is very helpful that when you encounter an issue to report an issue at https://issues.apache.org/jira/projects/ARROW. We may not be able to fix all things but it gives us at least some awareness.

maartenbreddels:
Thanks @xhochy, a look at the code would be great, and I will report issues for sure.

maartenbreddels:
Initial benchmark, pandas contains vs vaex' container (because of the c++ usage can actually use multithreading, quadcore) shows we can do 50x faster on testing if a string is in an expression.
image

Regex is much slower, where in general it seems that the regular expression library in Python is >2x as fast, so the almost 2x speedup (quadcore) seems reasonable:
image

packages/vaex-core/vaex/column.py
```diff
@@ -0,0 +1,252 @@
...
+        self.dtype = str_type
+        self.length = length if length is not None else len(indices) - 1
+        self.shape = (self.__len__(),)
+        self.nbytes = self.bytes.nbytes + self.indices.nbytes
+        self.string_sequence = vaex.strings.StringList(self.bytes, self.indices, self.length, self.offset)
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, slice):
```
xhochy: This looks familiar to https://github.com/xhochy/fletcher/blob/cbc35c1b168ad140bcaabdd8efd22a2425b52a03/fletcher/base.py#L331 Not the exact same code but I guess a bit of this shoudl go upstream

maartenbreddels: Ah, thanks for the pointer. Yes, that would be great. With upstream you mean in arrow?
I have the issue of having to attack a bit of a broader problem with strings than arrow (other formats, such as numpy fixed strings), so it goes a bit beyond arrow, but I hope in the end I can contribute something back!

xhochy: Yes, upstream is Arrow. `fletcher` is my toy project to better understand what users like Pandas or Vaex could need from Arrow.

maartenbreddels:
> We also have some code in Arrow when we construct string arrays that automatically splits them into mutliple ones (mostly in the Parquet handling code). So our path there is to use `ChunkedArray` instances.

Do you have pointers to the code for that? And I see fletcher also needs to deal with the chunks.

maartenbreddels:
The idea behind this PR seems solid, we have `df['x'].str.contains` and `df['x'].str.upper`, to proof the concepts. The open isues are:
 * [ ] More optimizations, removing python loops (in exporting for instance)
 * [ ] Dealing/testing long chunks` >= 2^31 ` string buffers.
 * [ ] A fairy decent list of string operations
   * replace
   * slice / slicing operator
   * find/rfind
   * match/contains/startswith/endswidth
   * lower/upper/title/capitalize/swapcase
   * strip/lstrip/lstrip
   * cat and + operator
   * count
   * split/join (these will be less efficient if we don't have a list-list-string implementation)

maartenbreddels:
We have a fair list of string operations for now, and performance seems great
image
(I see a performance regression in 'contains')

In the future we, when LargeStringArray lands in arrow, we may start depending on the c++ library of arrow, also the c++ code needs a refactor/clean/modernization.

maartenbreddels:
This will be merged first, completing the string operations will be a different PR.

PR #187, opened 2019-03-15
Updating the fillna test for expressions.

JovanVeljanoski:
Updating a unit-test for `fillna`.

There is a problem when `fillna` is used on an expression, and the output of that operation is overwriting an already existing expression in the DataFrame.


maartenbreddels:
The problem was that when all expressions got rewritten, the function calls AST -> string conversion didn't support keyword arguments, now fixed.

PR #190, opened 2019-03-18
Adding a generic test for the value_counts method.

JovanVeljanoski:
This adds a generic  test of the `value_counts` method, and it tests the application of the method to the various data types that vaex now supports.


maartenbreddels:
The objects type we cannot handle at the moment since numpy.unique simply cannot handle that. We'll have to attack that once we start adding groupby, and proper hashtables.

maartenbreddels:
If you are happy with this state, feel free to merge. Some tests are skipped, because of the object type, or because of masked values, we can later attack those. Would be good to open an issue for that and refer to this PR.

JovanVeljanoski:
I am happy to merge this.  Indeed for the xfailing tests, that work is planned for the near future, and as you suggest, we will open a PR to address these things again specifically. The main point of this PR was to test the compatibility with the new string support, and that is now aligned.

PR #192, opened 2019-03-20
Fix: trim long strings when printing dataframe

maartenbreddels:
This will truncate/trim very long strings/bytes.  `vaex.formatting. MAX_LENGTH_STRING` could be configured to change this.


JovanVeljanoski:
I am very happy with this PR. Some changes:
- the `MAX_LENGTH_STRING` was increased to 50 chars;
- some changes to `repr` tests to account for older `pandas` versions;
- small changes in how `np.datetime64` and `np.timedelta64` are being printed. 

One small drawback at the moment, for `datetime` like entries, if there are decimal places after the second, total of 6 decimal places will be printed in all cases. If there are no decimal places, none will be printed. We may want to update this behaviour in the future, but for now it is fine.

PR #194, opened 2019-03-29
Adding additional datetime operations, and improving test coverage.

JovanVeljanoski:
Adding additional `datetime` operations. Each operation is covered by a unit-test.

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -149,73 +149,61 @@ def fillna(ar, value, fill_nan=True, fill_masked=True):
 @register_function(scope='dt', as_property=True)
 def dt_dayofweek(x):
     import pandas as pd
-    # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.dayofweek.values
 
 @register_function(scope='dt', as_property=True)
 def dt_dayofyear(x):
     import pandas as pd
-    # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.dayofyear.values
 
 @register_function(scope='dt', as_property=True)
 def dt_is_leap_year(x):
     import pandas as pd
-    # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.is_leap_year.values
 
 @register_function(scope='dt', as_property=True)
 def dt_year(x):
     import pandas as pd
-    # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.year.values
 
 @register_function(scope='dt', as_property=True)
 def dt_month(x):
     import pandas as pd
-    # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.month.values
 
 @register_function(scope='dt', as_property=True)
 def dt_month_name(x):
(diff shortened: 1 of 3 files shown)


packages/vaex-core/vaex/functions.py
```diff
@@ -158,12 +158,42 @@ def dt_dayofyear(x):
     # x = x.astype("<M8[ns]")
     return pd.Series(x).dt.dayofyear.values
 
+@register_function(scope='dt', as_property=True)
+def dt_is_leap_year(x):
+    import pandas as pd
+    # x = x.astype("<M8[ns]")
```
maartenbreddels: copy paste artifacts.

tests/datetime_test.py
```diff
@@ -1,10 +1,27 @@
...
+
+    assert df.date.dt.hour.values.tolist() == pandas_df.date.dt.hour.values.tolist()
+    assert df.date.dt.minute.values.tolist() == pandas_df.date.dt.minute.values.tolist()
+    assert df.date.dt.second.values.tolist() == pandas_df.date.dt.second.values.tolist()
+    assert df.date.dt.day.values.tolist() == pandas_df.date.dt.day.values.tolist()
+    assert df.date.dt.day_name.values.tolist() == pandas_df.date.dt.day_name().values.tolist()
+    assert df.date.dt.month.values.tolist() == pandas_df.date.dt.month.values.tolist()
+    assert df.date.dt.month_name.values.tolist() == pandas_df.date.dt.month_name().values.tolist()
+    assert df.date.dt.year.values.tolist() == pandas_df.date.dt.year.values.tolist()
+    assert df.date.dt.is_leap_year.values.tolist() == pandas_df.date.dt.is_leap_year.values.tolist()
```
maartenbreddels: does one of them contain a leapyear?
you could do:
assert any(df.date.dt.is_leap_year.values.tolist())

packages/vaex-core/vaex/functions.py
JovanVeljanoski: Ok, I've cleaned this up now.

tests/datetime_test.py
JovanVeljanoski: Good point. I made one of them to contain a leap year. I still find it better to just do a direct comparison across all elements.

maartenbreddels: I mean both. 
```python
assert df.date.dt.is_leap_year.values.tolist() == pandas_df.date.dt.is_leap_year.values.tolist()
assert any(df.date.dt.is_leap_year.values.tolist())
```

JovanVeljanoski: Ok, added both.

maartenbreddels:
Actually, you can remove all these!
```
    @property
    def day_name(self):
        return self.expression.ds.func.dt_day_name(self.expression)
```

`@register_function(scope='dt', as_property=True)` will add these automatically (look at the stringoperations class!)

I almost forgot.

JovanVeljanoski:
I removed the redundant code. Tests pass.

PR #197, opened 2019-04-02
Superhash: binby, groupby, unique, value_counts and xarray support

maartenbreddels:
This is the start of using hashmaps in vaex for 
 * [x] groupby support
 * [x] binby support
 * [x] unique
 * [x] value_counts

Value count is 235x faster than the old implementation, 4.7x faster compared to pandas (benchmarked on 1e8 random integers between 0 and 99).


maartenbreddels:
cc @rabernat

I've chosen to split 'groupby' into two:
 * A 'traditional' groupby, that simply returns a new DataFrame,
 * A new binby method, that returns an [xarray](http://xarray.pydata.org) DataArray, which is a *much* more friendly data structure to work with if you do higher dimensional 'groupbys'. Also, it seems like a perfect match for vaex, since it produces regular N-d arrays. It also makes visualization *very* easy to do:

image

JovanVeljanoski:
I find this potential issue:

```
import vaex
import numpy as np

x = np.array(['a', 'a', 'a', 'b', 'b', 'c'], dtype=np.object)
ds = vaex.from_arrays(x=x)

ds.x.value_counts()
```

that returns
```
Series([], dtype: float64)
```
For big datasets this also sometimes kills the kernel, although I always reproduce it.

maartenbreddels:
Due to fundamental issues with the aggregation code, this led to a complete refactor, which also gives a 20-100% speedup in certain cases for 1 or 2 binning for non-float (and even float in some cases).
This leaves a lot of legacy code in, but there are a few aggregations still using the old code path, once we fully migrated, we can cut out a lot of legacy code.
We'll do this in a separate PR.

maartenbreddels:
I've squashed some commits (mainly CI debugging), but since it's a big change, I'm gonna merge this as is to have a fine grained history.

PR #200, opened 2019-04-03
Adding unit-tests to improve exporting of data to local storage.

JovanVeljanoski:
The added unit-tests exposes an issue when saving a `dataframe` which has a boolean column to the `arrow` format.


maartenbreddels:
Even though this is fixed, similar to bitmasks, we may want to have bitmap support for booleans as well. Maybe if we fully rely on arrow we get this for 'free'

PR #215, opened 2019-04-15
Fix small bug in dataframe.info()

JovanVeljanoski:


maartenbreddels:
I think fixes should always come with a test to expose the bug :)

maartenbreddels:
thanks!

PR #217, opened 2019-04-16
remove pandas dependency in formatting

### LINKED ISSUE #209: Pandas dependency
We now depends on Pandas:
https://github.com/vaexio/vaex/blob/255ccbc192d54c619a273de21a05f919da8ffadf/packages/vaex-core/vaex/formatting.py

Introduced in https://github.com/vaexio/vaex/pull/192

We should not depend on pandas, it is not a dependency of vaex-core and should not become, we might also grow to large to run on AWS Lambda.

JovanVeljanoski:
The printing of `datetime` and `timedelta` data types is no longer dependent on `pandas`.
Thus, `pandas` is no longer a hard dependency for `vaex`, which fixes #209 

This is how the printing looks like after the update:
image

### DIFF
--- a/packages/vaex-core/vaex/formatting.py
+++ b/packages/vaex-core/vaex/formatting.py
@@ -1,7 +1,7 @@
 import numpy as np
 import numbers
 import six
-import pandas as pd
+import datetime
 
 
 MAX_LENGTH = 50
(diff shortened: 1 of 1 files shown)


maartenbreddels:
see https://stackoverflow.com/questions/8337004/how-to-print-1-in-python-as-1-with-plus-sign-instead-of-1 to get rid of the code duplication

JovanVeljanoski:
> see https://stackoverflow.com/questions/8337004/how-to-print-1-in-python-as-1-with-plus-sign-instead-of-1 to get rid of the code duplication

Done. However, now we have behaviour like `2 days +1:40:00` where the time gets positive even though the days are also positive (before the time `+` sign was showing only for negative `timedelta`). 
Since we are talking about `timedelta`s here, I assume this is ok.

PR #224, opened 2019-04-23
fix deprecation issues

### LINKED ISSUE #223: Removal of (old?) deprecation warnings
It would be nice if we can remove some of the older deprecation warnings, and move to the new system wherever possible.  It would make following the pytest results much easier.

The most common is:

```
vaex/packages/vaex-core/vaex/hdf5/dataset.py:178: DeprecationWarning: Call to deprecated function selections_favorite_load: Will most likely disappear or move
  self.selections_favorite_load()
```

maartenbreddels:
Closes #223


maartenbreddels:
Failure is unrelated

PR #225, opened 2019-04-23
Fix autocomplete

### LINKED ISSUE #222: Deletion of a (virtual) column shows up in autocomplete (jupyter, ipython)
I am not 100% sure if this is the correct way we want to "delete" in-memory columns, in this example:
```
import vaex
ds = vaex.example()
del ds['x']
```
in Jupyter and IPython one still gets `ds.x` to show up via the auto-complete, while of course, the "x" column is no longer accessible.

(Note to anyone reading this: `del ds['x']` does not modify the original hdf5/arrow/csv file on disk. Well not unless one exports the `ds` and overwrites the original file. )

maartenbreddels:
Fixes #222


maartenbreddels:
CI failture unrelated, dependency issues with py35.

PR #226, opened 2019-04-23
fix: pandas has issues with read only buffers, this is a workaround

maartenbreddels:
Bug is fixed in pandas, we have to live with it for a while
https://github.com/pandas-dev/pandas/pull/21688

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -146,6 +146,16 @@ def fillna(ar, value, fill_nan=True, fill_masked=True):
 
 ########## datetime operations ##########
 
+
+def _pandas_dt_fix(x):
+    # see https://github.com/pandas-dev/pandas/issues/23276
+    import pandas as pd
+    # not sure which version this is fixed in
+    if not x.flags['WRITEABLE']:
+        x = x.copy()
+    return x
+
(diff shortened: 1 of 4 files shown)


packages/vaex-core/vaex/functions.py
```diff
@@ -146,6 +146,16 @@ def fillna(ar, value, fill_nan=True, fill_masked=True):
 
 ########## datetime operations ##########
 
+
+def _pandas_dt_fix(x):
+    # see https://github.com/pandas-dev/pandas/issues/23276
+    import pandas as pd
+    # not sure which version this is fixed in
```
xhochy: The fix came in 0.24 ;)

maartenbreddels: I thought so! I had a test on this, so it didn't copy it for 0.24, but the tests failed. Then I saw I had version 0.24.1 and @JovanVeljanoski (who reported the issue) had 0.24.2.
In case you want to reproduce, comment out line 154-155 and run
```
py.test tests/datetime_test.py
```
Thanks for keeping an eye on this!

maartenbreddels:
Failure is unrelated, merging this. If we find out how to not copy for pandas, we can always fix that later.

PR #233, opened 2019-04-27
groupby rows consistency test

JovanVeljanoski:
Adding a `groupby` which tests for the consistency of results when the group is made on a feature extracted from a datetime column. There is sometimes a problem when the `df.executor.buffer_size` is much smaller than the number of rows in that `df`.


maartenbreddels:
Ignore appveyor again, is you're happy feel free to merge. Thanks!

JovanVeljanoski:
The updates you made are excellent. There also seems to be a nice performance improvement as well. So I am very happy indeed!. Thanks!

PR #234, opened 2019-04-27
Fixes the info panel where the string types were not being displayed

JovanVeljanoski:
Fixes the `df.info()` where in the column description table the type of the string columns was not displayed.


maartenbreddels:
Looks good, error seems unrelated, will pick that up later, feel free to push the green button! Thanks

PR #235, opened 2019-04-27
Timedelta operations

JovanVeljanoski:
This PR adds a unit-test checking the basic arithmetic operations between a `vaex.Expression` of type `timedelta64` and `numpy.timedelta64`.

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -92,6 +92,9 @@ class Meta(type):
                     if isinstance(b, Expression):
                         assert b.ds == a.ds
                         b = b.expression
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I think this is a temporary workaround, basically we need to have support for object in expressions besides int, float, strings and expression. For instiance, this may not work for ds.state_get().
Maybe you want to open an issue for that. Or leave that for me, and remind me of it.

JovanVeljanoski:
Thanks for this. It works for now, and we can do other enhancements as needed. I may open an issue if I can get a nice unit-test, otherwise I will remind you about this soon.

Thanks!

PR #239, opened 2019-04-29
Remove deprecation warning due to collection.abc import

JovanVeljanoski:
Small fix to remove the deprecation warning from vaex-core/vaex/settings.py to remove the deprecation warning coming from importing the `collections.abc` module. 

The same fix was already applied in vaex-core/vaex/expression.py

### DIFF
--- a/packages/vaex-core/vaex/settings.py
+++ b/packages/vaex-core/vaex/settings.py
@@ -1,8 +1,6 @@
-import logging
-import yaml
-from yaml import Loader, Dumper
-import vaex.utils
 import os
+import logging
+import vaex.utils
 import collections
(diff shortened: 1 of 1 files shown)


packages/vaex-core/vaex/settings.py
```diff
@@ -1,10 +1,13 @@
+import os
 import logging
-import yaml
-from yaml import Loader, Dumper
 import vaex.utils
-import os
 import collections
 
+try:
+    collectionsAbc = collections.abc
```
maartenbreddels: Please use snake_case instead of camelCasing, which seems unnatural to me.

PR #244, opened 2019-05-07
Unit-test for datetime/timedelta arithmetic and filtering

JovanVeljanoski:
Adding a unit-test for testing the behaviour of a filtered DataFrame which contains arithmetic with datetime and timedelta inputs. 

Note that the test as it is exposes some weird bug, and this test itself is not registered py `pytest`, however the logfiles register that something has gone wrong.


maartenbreddels:
This is replaced by #245

Issue #240, opened 2019-05-05
Get columns based on type


xdssio:
Examples:
```
data[int] -> data[colums of type int]
data[float, int] -> data[colums of types float and int]
```

JovanVeljanoski:
Closed via #2160

PR #251, opened 2019-05-11
Get columns based on types

JovanVeljanoski:
This is regarding #240.
This is a proposal PR that enables one to select columns based in types. An example use case is shown below.

```
s = ['a', 'ab', 'abc']  
b = [True, False, True]
f = [1.2, 2.3, 5.5]
display(df)
df = vaex.from_arrays(s=s, b=b, f=f)
display(df[str])
display(df[float, bool])
```
image

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4416,8 +4416,8 @@ class DataFrame(object):
             return df
         elif isinstance(item, (tuple, list)):
             # check if list or tuple contains types
-            if all([True if type(i)==type else False for i in item ]):
-                columns = [column for dt in item for column in self if self.dtype(column) == dt]
+            if all([type(i) == type for i in item ]):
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/dataframe.py
```diff
@@ -4415,8 +4415,20 @@ def __getitem__(self, item):
             df._selection_masks[FILTER_SELECTION_NAME] = vaex.superutils.Mask(df._length_unfiltered)
             return df
         elif isinstance(item, (tuple, list)):
-            df = self.copy(column_names=item)
-            return df
+            # check if list or tuple contains types
+            if all([True if type(i)==type else False for i in item ]):
```
maartenbreddels: `True if expression else False` can be replaced `expression`, also check pylint.

maartenbreddels:
Note: take a look at https://docs.scipy.org/doc/numpy-1.10.1/reference/arrays.scalars.html

JovanVeljanoski:
closed via #2160

PR #253, opened 2019-05-12
Superjoin: join now uses the new hash tables

maartenbreddels:
Joining is now using the same hash table structures. Benchmarks should follow soon.


maartenbreddels:
failures are unrelated.

PR #254, opened 2019-05-12
vaex-ml: package centred around machine learning related tasks

JovanVeljanoski:
This is a big PR in which we are introducing `vaex-ml`, a `vaex` package centred around machine learning related tasks and applications. The following describes the contents:

- `vaex.ml.transformations`: methods related to preprocessing: scalers, categorical encoders, PCA
- `vaex.ml.cluster`: provides an efficient KMeans clustering algorithm
- `vaex.ml.ui`: provides means to construct an `ipywidget` for nearly any transformer in this package
- `vaex.ml.xgboost`: a binding to the [`xgboost`](https://github.com/dmlc/xgboost) library
- `vaex.ml.lightgbm`: a binding to the [`lightgbm`](https://github.com/microsoft/LightGBM) library
- `vaex.ml.catboost`: a binding to the [`catboost`](https://github.com/catboost/catboost) library
- `vaex.ml.sklearn`: a binding to the [`scikit-learn`](https://github.com/scikit-learn/scikit-learn) library. At the moment, only the estimators are supported.
- `vaex.ml.incubator`: a module housing various machine learning models. The bindings in the incubator are considered experimental and are under testing. The API, implementation or support may change without notice. 
- `vaex.ml.datasets`: contains datasets for experimentation and training. Currently contains the titanic and the iris classical datasets. It also contains methods for replicating the iris dataset such that it contains a total of 10^9 samples, creating a "big data" example.
- `vaex.ml.generate`: module that auto-generates an alternative API for the transformers and ML models.
- `vaex.ml.pipeline`: provides a pipeline object for the `vaex-ml` transformers and estimators
- `vaex.ml.state`: methods for serialisation of `vaex` objects
 - `vaex.ml.linear_model`: provides an implementation of linear models that operate on a grid (binned data) instead of individual samples.

Everything comes with a full suite of tests handled by `pytest`. 

NB: The contents above and this description itself may change slightly as `vaex-ml` is integrated with `vaex`.


packages/vaex-meta/setup.py
```diff
@@ -21,7 +21,8 @@
       'vaex-server==0.2',
       'vaex-hdf5==0.4',
       'vaex-astro==0.4',
-      'vaex-arrow==0.3'
+      'vaex-arrow==0.3',
+      'vaex-ml==0.4'
```
maartenbreddels: I don't think vaex-ml should be installed by default when you install vaex, like vaex-ui, what do you think?

maartenbreddels: Especially since vaex-ml now still depends on numba.

packages/vaex-ml/vaex/ml/.coverage
```diff
@@ -0,0 +1 @@
+!coverage.py: This is a private format, don't read it directly!{"lines":{}}
```
maartenbreddels: I don't think this file should go into the repo.

packages/vaex-meta/setup.py
JovanVeljanoski: Why not?  I thought it is good to come as a default. I am afraid it will not get noticed or add extra complexity otherwise. 

 For production environments, people can of course choose what to install.

packages/vaex-ml/vaex/ml/.coverage
JovanVeljanoski: Agreed! I tried to exclude it.. i will check the gitignore file again

JovanVeljanoski: Same for the pycache/ like things.

maartenbreddels:
🎉 yeah!

PR #255, opened 2019-05-14
Unit-tests for additional groupby aggregators

JovanVeljanoski:
Added unit-tests for

- standard deviation aggregator for numeric types;
- mode (most common value) for string/object type.

### DIFF
--- a/packages/vaex-core/src/superagg.cpp
+++ b/packages/vaex-core/src/superagg.cpp
@@ -654,6 +654,51 @@ public:
 };
 
 template<class StorageType=double, class IndexType=default_index_type, bool FlipEndian=false>
+class AggSumMoment : public AggBase<StorageType, StorageType, IndexType> {
+public:
+    using Base = AggBase<StorageType, StorageType, IndexType>;
+    using Type = AggSumMoment<StorageType, IndexType, FlipEndian>;
+    using Base::Base;
+    AggSumMoment(Grid<IndexType>* grid, uint32_t moment) : Base(grid), moment(moment) {
+    }
+    virtual void reduce(std::vector<Type*> others) {
+        for(auto other: others) {
+            for(size_t i = 0; i < this->grid->length1d; i++) {
+                this->grid_data[i] = this->grid_data[i] + other->grid_data[i];
+            }
+        }
+    }
+    virtual void aggregate(default_index_type* indices1d, size_t length, uint64_t offset) {
+        if(this->data_ptr == nullptr) {
+            throw std::runtime_error("data not set");
+        }
+
+        if(this->data_mask_ptr) {
+            for(size_t j = 0; j < length; j++) {
+                // if not masked
+                if(this->data_mask_ptr[j+offset] == 1) {
+                    StorageType value = this->data_ptr[j+offset];
+                    if(FlipEndian)
+                        value = _to_native(value);
+                    if(value != value) // nan
+                        continue;
+                    this->grid_data[indices1d[j]] += pow(value, moment);
+                }
+            }
+        } else {
+            for(size_t j = 0; j < length; j++) {
+                StorageType value = this->data_ptr[offset + j];
+                if(FlipEndian)
+                    value = _to_native(value);
+                if(value == value) // nan check
+                    this->grid_data[indices1d[j]] += pow(value, moment);
+            }
+        }
+    }
+    size_t moment;
+};
+
+template<class StorageType=double, class IndexType=default_index_type, bool FlipEndian=false>
 class AggFirst : public AggBase<StorageType, StorageType, IndexType> {
 public:
     using Base = AggBase<StorageType, StorageType, IndexType>;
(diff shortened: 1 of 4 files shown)


tests/groupby_test.py
```diff
@@ -140,14 +140,34 @@ def test_groupby_count():
...
+    assert dfg.g.tolist() == [3.112474899497183, 1.118033988749895, 1.5]
+
+
 def test_groupby_count_string():
     g = np.array([0, 0, 0, 0, 1, 1, 1, 1, 2, 2])
     s = np.array(list(map(str, [0, 0, 0, 0, 1, 1, 1, 1, 2, 2])))
     df = vaex.from_arrays(g=g, s=s)
     groupby = df.groupby('s')
-    dfg = groupby.agg({'c': vaex.agg.count('s')})
+    dfg = groupby.agg({'m': vaex.agg.mode('s')})
```
maartenbreddels: I guess this should still be count? (copy paste bug?)

tests/groupby_test.py
```diff
@@ -140,14 +140,34 @@ def test_groupby_count():
...
+
+def test_groupby_mode_string():
+    g = np.array([0, 0, 0, 0, 1, 1, 1, 1, 2, 2])
+    s = np.array(list(map(str, [0, 0, 0, 0, 1, 1, 1, 1, 2, 2])))
+    df = vaex.from_arrays(g=g, s=s)
+    groupby = df.groupby('s')
+    dfg = groupby.agg({'m': vaex.agg.mode('s')})
     assert dfg.s.tolist() == ['0', '1', '2']
-    assert dfg.c.tolist() == [4, 4, 2]
+    assert dfg.m.tolist() == ['0', '1', '2']
```
maartenbreddels: I'm not sure the test as it is is very solid, it does not have within 1 group a mix of different 's', a more thorough test would test a mix, where in 1 group the order is different from the other etc, this test is too easy to pass.

tests/groupby_test.py
JovanVeljanoski: Ah yes, indeed. Fixed in the upcoming commit.

tests/groupby_test.py
JovanVeljanoski: Indeed, the test can be made more proper. I guess my idea was just to present a use-case. Better test is coming in commit soon.

maartenbreddels:
Note that mode is not implemented yet.

packages/vaex-core/src/superagg.cpp
```diff
@@ -653,6 +653,51 @@ class AggSum : public AggBase<StorageType, StorageType, IndexType> {
...
+                        continue;
+                    this->grid_data[indices1d[j]] += pow(value, moment);
+                }
+            }
+        } else {
+            for(size_t j = 0; j < length; j++) {
+                StorageType value = this->data_ptr[offset + j];
+                if(FlipEndian)
+                    value = _to_native(value);
+                if(value == value) // nan check
```
manodeep: Thank you again for creating and maintaining `vaex`. Hope this is helpful.

This `value == value` condition will be optimised away by most modern compilers when compiling with any optimisation flags, and therefore will not work as expected. One way to check for NANs (at least in `C`) is to add a `volatile` modifier to variable "value". 

I also notice that you are using `pow` for computing the moment, where `moment` is a `size_t`. Might be faster to compute the chained products for `moment = [1, 2, 3] -> [value, value*value, value*value*value]` and then default to `pow` for the general case. 

Thanks!

maartenbreddels: Thank you!

value == value will indeed be optimized away with `--ffast-math`, but I've never seen it been optimized away without it (the unittest will trigger it). But I've never really dug into what is the fasted nan test.

I really like the idea for the moments, I think we should have a moments aggregator, that will do this (up to a particular moment order)! Would you mind opening an issue for this?

cheers,

Maarten

manodeep: Interesting point about the NAN check. All of my reading implies that the use of `volatile` is necessary to ensure correct behaviour. May be that relates to `icc` being over-eager and enabling some degree of `ffast-math` by default. My strategy is always to call `isfinite(float)` and not worry about the run-time performance of the resulting call. (On my OSX laptop, `math.h` does contain two implementations for `isfinite` -> one for the "normal" case, and another for when `ffast-math` is enabled.)

I will open the issue regarding the moments. This is the `C` code I would write for that implementation (likely to be faster than an equivalent `switch` statement):

```C
const StorageType moment_values[] = {1, value, value*value, value*value*value};
const size_t max_moment = sizeof(moment_values)/sizeof(StorageType);
this->grid_data[indices1d[j]] += (moment < max_moment) ? moment_values[moment]: pow(value, moment);
```

manodeep: I will note that my implementation above does not check for overflow/underflow. A better implementation would use an inter-mediate (larger-width) variable to hold the sum, and then cast down to the appropriate type at the end. Plus, depending on how the sums are evaluated, there might need to be a compensated algorithm (e.g, Kahan sum etc)

PR #261, opened 2019-05-16
Unit-test for the string formatter (e.g. df.x.format() )

JovanVeljanoski:
A unit-test exposing a potential problem with the string formatter when integers are in question.


JovanVeljanoski:
In addition to the cases covered in this PR, a use case such as 
```
df.x.format('%s')
```

kills the kernel of the Jupyter notebook. While usage as above is not supported per http://www.cplusplus.com/reference/string/to_string/ probably this need to be handled better at some point.

maartenbreddels:
The merge commit passes, so this is ok.

PR #263, opened 2019-05-17
support for np.isnan and np.isinf

JovanVeljanoski:
This PR adds and tests the support for `np.isnan` and `np.isinf`. This was initially added by #161 , but since the significant changes to `vaex` that can no longer me merged. Credit and thanks goes to @DougRzz for initiating this.

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -101,6 +101,8 @@ minimum
 maximum
 clip
 searchsorted
+isnan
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Note: This requires some changes in Expression and a more modern numpy.

maartenbreddels:
np.isinf() does not seem to like to take expressions, should we rewrite this first, to use df.x.isinf? An open an issue for future support?

PR #264, opened 2019-05-18
fix(core,Expression.derivative): derivative passed Expression instead of DataFrame to Expression constructor

maartenbreddels:


maartenbreddels:
failure unrelated.

PR #266, opened 2019-05-19
Unit-test for fillna() when the expression is virtual.

JovanVeljanoski:
A unit-test exposing a potential problem when the `fillna()` method is used on a virtual expression, where the virtual expression contains `nan` values as a result of the calculationnot tri.

Note that the unit-test does trigger a failure: due to the nature of the problem `pytest` does not register the test at all, but looking at the logs one can easily spot the problem.


maartenbreddels:
Thinking about it I don't think this has anything to do with fillna, it is simply a bug for reassigning to a virtual column. Can you confirm that is the issue Jovan?

PR #285, opened 2019-06-05
Adding a from_json method

JovanVeljanoski:
Adding a `from_json` method, allowing to read JSON files directly into a `vaex` DataFrame.


xdssio:
this is super useful, in almost every scenario but files, when data goes through the network, it passes as JSON, also for queries, samples, and databases many times.

maartenbreddels:
Thanks, looks good!

maartenbreddels:
I think this is asking for to_json :) I know @xdssio wants it badly right?

PR #289, opened 2019-06-07
nunique() method for expressions

JovanVeljanoski:
A method which returns the number of unique elements in an `expression`.

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -462,6 +462,13 @@ class Expression(with_metaclass(Meta)):
     def unique(self):
         return self.ds.unique(self.expression)
 
+    def nunique(self, dropna=True):
+        """Return the number of unique elements in the expression."""
+        n = len(self.unique())
+        if dropna:
+            return n - 1
+        return n
+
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/expression.py
```diff
@@ -462,6 +462,13 @@ def reduce(a, b):
     def unique(self):
         return self.ds.unique(self.expression)
 
+    def nunique(self, dropna=True):
+        """Return the number of unique elements in the expression."""
+        n = len(self.unique())
+        if dropna:
+            return n - 1
```
maartenbreddels: this won't always work, what if there is no missing values or nans? Also, I think we should have a consistent argument set each time: `dropna=True, dropnan=True, dropmissing=True` maybe? @xdssio: what do you think?

JovanVeljanoski: Ah yes... also the presence of `None` values also messes the counts up.. but I just noticed that `None` values are not dropped by `dropna`.

If we want to be similar to `pandas`, we should adopt `dropna` as Not Available (n/a)

maartenbreddels:
Already implemented with #412

PR #290, opened 2019-06-07
isna(), notna(), countna() methods for `vaex.expression`

JovanVeljanoski:
[Edited]

This PR adds `veax.expression` methods:

- `isna()` which returns a boolean expression in which an element is `True` if it is not available (N/A) (either `None` or `np.nan`);
- `notna()` which returns a boolean expression which is opposite of `notna()`;
- `countna()` which returns the number of not available (N/A) values (either `None` or `np.nan`).

Each of the methods comes with a unit-test.

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -482,6 +482,13 @@ class Expression(with_metaclass(Meta)):
         """
         return (self == self)
 
+    def countna(self):
+        """Returns the number of NaN values in the expression. Note that valus
+        such as None, and empty strings do not get counted, i.e. they are not
+        treated as NaN values.
+        """
+        return (self != self).astype('int').sum() + 0  # so the output is int, not array
+
     def evaluate(self, i1=None, i2=None, out=None, selection=None):
         return self.ds.evaluate(self, i1, i2, out=out, selection=selection)
 
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -469,25 +469,22 @@ class Expression(with_metaclass(Meta)):
         return self.ds.unique(self.expression)
 
     def isna(self):
-        """Returns a boolean expression indicating if the values are NaN.
-        Note that values such as None, and an empty string get mapped to False,
-        i.e. they are not treated as NaN values.
+        """Returns a boolean expression indicating if the values are Not Availiable
+        (N/A), i.e. if they are None or numpy.nan.
         """
-        return ~(self == self)
+        return ~(self == self) | (self==None)
 
     def notna(self):
-        """Returns a boolean expression indicating if the values are not NaN.
-        Note that values such as None, and an empty string get mapped to True,
-        i.e. they are not treated as NaN values.
+        """Returns a boolean expression indicating if the values are not Not Availiable
+        (N/A), i.e. if they are not None and numpy.nan.
         """
-        return (self == self)
+        return (self == self) & ~(self==None)
 
     def countna(self):
-        """Returns the number of NaN values in the expression. Note that valus
-        such as None, and empty strings do not get counted, i.e. they are not
-        treated as NaN values.
+        """Returns the number of Not Availiable (N/A) values in the expression.
+        This includes both None and np.nan values.
         """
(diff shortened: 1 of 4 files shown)


maartenbreddels:
I think we should discuss what we mean by 'na'. I think in general we took the following approach
 * nan means really NaN (so only for floats)
 * missing means missing as in a masked array
 * na means no available, so it's nan or missing.
So I think we should rename it to isnan, or make isna reflect the decision we made in other parts of the code and write this down in the docs, or reconsider.

xdssio:
Although the logic is valid, for practitioners, NA and NaN mean the same, in the sense that you need to clean them in the same way, therefore it's adding complexity which in my opinion unnecessary. 
I think that we only need "na" which includes np.nan and None (not empty string) and if you want to test for specific case, which is an edge case, you can do something like `df[df.x == None]` or `df[df.x is None]`  or `ds[ds.x == np.nan]` and such.

maartenbreddels:
The reason you need isnan is that `nan ~= nan`.

I think `.na()` should intuitively be what I described, nan or missing (masked), not just NaN or just missing. I think we agree here right? Note that `None` doesn't really exist, except for dtype=object, but I think what you mean by None is missing, which would be a masked array.

xdssio:
The important thing is that `countna` will return all invalid values, Nones, NaN, anything which will be needed to drop or fill, this is what we expect when counting it. 
`dropna()` should remove rows with them, and `fillna(<value>)` to fill them up.    
having `isnan`, `fillnan` and `dropnan` could be nice, but I think people will always use the first three I wrote.

maartenbreddels:
I think we're agreeing, but it does not agree with what Jovan said
> isna() which returns a boolean expression in which an element is True if it is a np.nan;

xdssio:
I would suggest we should have Nones on str columns, which is different than an empty string.

maartenbreddels:
we're not talking about empty strings at all, those are not None/missing/NaN whatever.

Basically, I don't agree that isna is implemented as isnan, we can have isnan, but if we add a isna, is should also test for missing values.

JovanVeljanoski:
Hi all, 

Thanks for the comments. Indeed, my idea was to treat the 'na' as N/A, i.e. not available which would include `numpy.None` and `None` (I am not sure whether the empty strings should be considered as N/A here). 

The reason why I only implemented the `np.nan` part is.. simply because I do not know how to detect `None`s yet with `vaex`. So if you can give me any pointers.. I would be happy to do it, i agree it is much more useful this way. 

So this was  just so I can get your attention, and now that I have it, I would appreciate any pointers on how to do this in a more complete way. 
Thank you.

maartenbreddels:
> Thanks for the comments. Indeed, my idea was to treat the 'na' as N/A, i.e. not available which would include `numpy.None` and `None` (I am not sure whether the empty strings should be considered as N/A here).

I don't think empty strings should be seen an NA. But what is missing here is support for masked arrays. Cmd-f for `np.ma` and you'll see many cases in vaex where we check that.

> The reason why I only implemented the `np.nan` part is.. simply because I do not know how to detect `None`s yet with `vaex`. So if you can give me any pointers.. I would be happy to do it, i agree it is much more useful this way.

None's only have to be detected for dtype=object right? Actually, if the column is seen as string it automatically will be seen as missing.

Yes, great work and I think we agree now, let me recap:
NA means:
 * None for dtype=object (and automatically for strings which are dtype=object)
 * NaN for floats or dtype=object
 * a masked array with value set to 1

NaN means:
 * NaN :)

JovanVeljanoski:
Ok, I made an update to the PR. 

Now both `None` and `np.nan` values are counted as N/A (not available). 

I also improved the unit-tests, added cases with masked arrays, and assert against `pandas` methods.

However I am not able to account for the masked array version (which is why the unit-tests fail). I saw examples of how they are handled in other parts of the source code, and it seems to me that in those cases, the pure underlying `numpy` arrays are being used, and not expressions.  I run into a weird edge case (see unit-tests), where you have a masked array but also a `None`, `np.nan` values that **are not masked!**  I know this is strange and it should not happen in practice, but hey, there are lots of use cases out there, and one can manually create such arrays for whatever reason.  

@maartenbreddels can you give me a hand with this, how can we handle it?

PR #294, opened 2019-06-10
fix(core): file column should support nbytes property

### LINKED ISSUE #284: #267 s3 file reading
I installed vaex in develop mode and I'm trying to read a file in S3, but it doesn't work. I run tests and **s3_test** failed with  

`OSError: Unknown error opening: s3://vaex/testing/xy.hdf5?cache=false&anon=true`

among few others (7 in total)

This looks like before having the feature, my guess is that I didn't install vaex well. Though it seems okay. I tried reinstalling, I dunno...

maartenbreddels:
This is the start for the support of strings for s3 files needed to fix #284. It requires more changes however. We assume the data is always locally available as contiguous array, we should only do that when we want to access the data.

### DIFF
--- a/packages/vaex-core/vaex/file/column.py
+++ b/packages/vaex-core/vaex/file/column.py
@@ -51,6 +51,10 @@ class ColumnFile(vaex.column.Column):
         self.shape = (length,)
         self.write = write
 
+    @property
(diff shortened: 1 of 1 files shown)


xdssio:
+1

PR #297, opened 2019-06-13
fix(hdf5): support for string lists with zero length, fixes #273

### LINKED ISSUE #273: Error when exporting df with a column containing only empty strings.
There is an error appearing when one tries to export a DataFrame that contains a column which has only empty strings.

`df.export_arrow` does not seem affected by this, but opening such a file causes an error.

maartenbreddels:


maartenbreddels:
Seems a dependency for windows is broken, ignoring appveyor for now

PR #299, opened 2019-06-14
fix(core): give a better error msg on copying a dataframe with non-existing column

### LINKED ISSUE #221: Better handling when a non-existend column is used.
We need a better way to catch an error when a user makes a typo and queries a column that does not exist in a `DataFrame`. For example:
```
import vaex
ds = vaex.example()
ds[['x', 'xy']]   # gives a not so friendly error
```

On the other hand:
```
import vaex
ds = vaex.example()
ds['xy']   # gives a helpful error message
```

maartenbreddels:
Closes #221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
       * Executor does not block after failed/interrupted tasks. [#571](https://github.com/vaexio/vaex/pull/557)
       * Default limits (e.g. for plots) is minmax, so we don't miss outliers [#581](https://github.com/vaexio/vaex/pull/581)
       * Do no fail printing out dataframe with 0 rows [#582](https://github.com/vaexio/vaex/pull/582)
+      * Give proper NameError when using non-existing column names [#299](https://github.com/vaexio/vaex/pull/299)
    * Features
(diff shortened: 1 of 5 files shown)


maartenbreddels:
Instead of all names, we only hint at similar names:
image

maartenbreddels:
Not all paths have this extra check, this is the normal error. Try with
[['sexx']], if you agree with what that looks like, we can then check at
other places as well.

(from mobile phone)

JovanVeljanoski:
Ah yes you are right! So selecting a subset of a dataframe works
df[['pclas]] for example. This also works df['pclas', 'sexx'] (which is not supported by pandas for example. 

Is it possible for the same message to be shown when an expression is queried?
df['sexx'], i.e. to give the "Column or variable 'sexx' does not exist. ". Does not have to give the suggestions. Just for consistency.

maartenbreddels:
> This also works df['pclas', 'sexx'] (which is not supported by pandas for example.

that is accidental, I don't think we should do/support that, do you?

maartenbreddels:
like it? 
image

JovanVeljanoski:
Stellar!

Issue #238, opened 2019-04-27
Better row interface


JovanVeljanoski:
It would be very helpful if `df.expression` supported

- [ ] slicing,  (e.g. `df.x[:10]` and `df.x[:10].values`)
- [ ] `argmin` and `argmax`, (e.g. `df.x.argmin()` and `df.x.argmax()`)

Perhaps it would be also nice to have support for `argsort` as a whole, although that would be quite intensive for large datasets. The use-case I see coming up is the need to find the _n_ smallest or largest indices of a  particular expressions, so those rows in the DataFrame can be queries or inspected in more detail.

PR #300, opened 2019-06-14
feat(core,expression): allow slicing of expressions

maartenbreddels:
See https://github.com/vaexio/vaex/issues/238


acubed3:
Dear Maarten, does it work or not?

PR #309, opened 2019-06-25
binary operations with numpy.datetime64 instance

JovanVeljanoski:
Adding the possibility to perform binary operations with `numpy.datetime64` instance.

e.g.:
```
df.startDate > np.datetime64('2019-05-05')
```
will not result in an expression.


maartenbreddels:
Awesome, thanks!

PR #313, opened 2019-06-26
add object_type is str test

xdssio:
No more Object type.
For starters, make it str and ket the user deal with the consequences if it's not.

Not really at the same importance level, but can we change `<class 'str'>` to `str`? on the way
```
>>> df.dtype:
...
PassengerId         int64
Survived               int64
Pclass                   int64
Name           <class 'str'> -> str
```

### DIFF
--- a/tests/obj_type_test.py
+++ b/tests/obj_type_test.py
@@ -1,0 +1,16 @@
+import vaex
+
+
+def object_type_test():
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Yes, I don't want dtype=object, but that's what pandas gives us. I think you test passes, which means that we already do this.
Regarding the nice printing, we cannot change the way that Python `repr(str)` works, so this is not straight forward (I see no way we can do this, and we discussed this multiple times already)

PR #314, opened 2019-06-26
test for lists

xdssio:
using a column of lists.
obviously, it will be "nice to have" to have a list of multiple types, but mostly it's important to have a homogenise column for pixels, text tokens, and packed columns (the one-hot-encoder could result with a single column instead of 5000 columns for 5000 values). 

* In-text case, lists are not the same length, which was a challenge for TF, so they made a special Tensor that can have multiple lengths, but paddings can be the first solution.
* We might need a pack_column, unpack methods to turn pack columns to a column of lists and back to columns.

### DIFF
--- a/tests/list_column_test.py
+++ b/tests/list_column_test.py
@@ -1,0 +1,12 @@
+import vaex
+import os
+
+
+def test_list_column():
+    df = vaex.from_dict(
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Agree, images are N-d arrays, but other structures will have to wait for when we will fully depend on arrow.

PR #315, opened 2019-06-26
filters

xdssio:
a standard function for text tokenisers


maartenbreddels:
I don't think we should come up with new names, translate would be best for this:
https://docs.python.org/2/library/stdtypes.html#str.translate

PR #316, opened 2019-06-27
Better str representation in dataframe.dtypes

JovanVeljanoski:
Modify the `dataframe.dtypes` to make the printing of the `str` `dtype` more pretty.


maartenbreddels:
We cannot do this, you want to access really the dtype. E.g
```
assert df.text.dtype == df.dtypes['text']
```
Now `df.dtypes['text']` is "str", not the str object.

JovanVeljanoski:
Ah ok I see. 

Since we have no control over the `repr` for `str`, I will close this for now.

PR #324, opened 2019-07-01
Improving the LabelEncoder

JovanVeljanoski:
This PR makes improvements on the implementation of `vaex.ml.LabelEncoder()`. It leverages the super fast `map` method. and runs fully out of core. There is now support for values that were not present during the training (fitting) stage. 

The past implementation was not fully out of core as far as I can tell.

Also, there is a small adjustment to the unit-test. The encoder is longer tested against the `sklearn`s implementation, for the simple reason that the order of the unique elements (categories) in a feature is not guaranteed to be the same in `vaex` and `sklearn`. However, I am comfortable with this, since the `map` method is properly tested.

### DIFF
--- a/packages/vaex-ml/vaex/ml/__init__.py
+++ b/packages/vaex-ml/vaex/ml/__init__.py
@@ -23,15 +23,16 @@ def pca(self, n_components=2, features=None, prefix='PCA_', progress=False):
     return pca
 
 
-def label_encoder(self, features=None, prefix='label_encoded_'):
+def label_encoder(self, features=None, prefix='label_encoded_', allow_unseen=False):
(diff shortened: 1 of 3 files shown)


JovanVeljanoski:
Yeah, the test was just making sure the core runs. It is tricky, since now `LabelEncoder` uses the internal `unique` method, which gives the results in a somewhat random order, so not easy to test against an external library like before.

I have tried to circumvent this and to "solve" the problem just with `numpy` (since the data generated for this test is tiny) and assert the results against that. This sort of verifies the implementation a bit better and provides a better test, if not a perfect one. 

I am satisfied with this as it is (given the latest commit), so if you are also happy I am willing to merge this.

PR #325, opened 2019-07-01
Unit-test for the unique method when the column created from bytes

JovanVeljanoski:
Adding a unit-test which tests the `unique` method, in the case when the column is created from bytes. At the moment this exposes a potential issue.


maartenbreddels:
I'm not super happy with this fix, since it basically converts bytes to strings.

Voloda357:
Володя Пылыпенко Ивановыч мне 48 лет проживаю в Сели Андреевка Машивского ра-ну Полтавская об Украине 🇺🇦

PR #334, opened 2019-07-04
Improved test for LabelEncoder (vaex-ml)

JovanVeljanoski:
The `LabelEncoder` (`vaex-ml`) test is improved to remove the inherent flakiness due to the randomness with which the data was generated. 

The data now is fixed, and thus we have much better control over the test results. Everything is tested much in the same way as before.


maartenbreddels:
LGTM

PR #335, opened 2019-07-04
Unit-test for unique in case of dtype object and n/a values

JovanVeljanoski:
Unit-test for `unique`, which tests the behaviour in case of `dtype` `object` and the presence of n/a data (`np.nan`, `None`).

### DIFF
--- a/tests/unique_test.py
+++ b/tests/unique_test.py
@@ -37,3 +37,10 @@ def test_unique_nan():
         mask = np.isnan(values)
         assert values[~mask].tolist() == df.x.values[~mask].tolist()
         # assert indices.tolist() == [0, 1, 2, 0, 3, 0]
+
+def test_unique_missing():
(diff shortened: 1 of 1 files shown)


maartenbreddels:
vaex does not see this as dtype=object, because the first element is a string.

maartenbreddels:
Also, I don't agree with what you think the answer should be, None should be in it I think.

JovanVeljanoski:
> vaex does not see this as dtype=object, because the first element is a string.

Sure but even if all elems are assumed as strings, not all unique ones are found.. 

> Also, I don't agree with what you think the answer should be, None should be in it I think.

Yea we can do that.. one might argue that perhaps we can also display `np.nan` even though `np.nan != np.nan` .... I am find with your suggestion. More important is to find all the uniques :)

PR #345, opened 2019-07-07
feat(core,groupby,binby): category support

maartenbreddels:
When groupby is used with categories, it will show the labels in the aggregated dataframe.
```
g = np.array([0, 0, 0, 0, 1, 1, 1, 1, 2, 2])
df = vaex.from_arrays(g=g)
df.categorize('g', ['cat', 'dog', 'snake'])
df.groupby(by=df.g, agg='count')
```

### DIFF
--- a/packages/vaex-core/vaex/agg.py
+++ b/packages/vaex-core/vaex/agg.py
@@ -81,7 +81,7 @@ class AggregatorDescriptorBasic(AggregatorDescriptor):
         return spec
 
     def pretty_name(self, id=None):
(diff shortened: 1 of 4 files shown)


maartenbreddels:
@JovanVeljanoski 
```
dfg = df.groupby(by=df.g, agg='count')
```
Now dfg.g contains the labels of the category, instead of the numbers. Also, it is much faster, since we do not need a hashmap now.

maartenbreddels:
before:
image
after:
image

PR #353, opened 2019-07-11
Fix min and max

maartenbreddels:


maartenbreddels:
windows failures are unrelated

PR #364, opened 2019-07-24
Improve catboost unit test

JovanVeljanoski:
This makes a `catboost` unit-test be more robust to changes its performance.

### DIFF
--- a/tests/ml/catboost_test.py
+++ b/tests/ml/catboost_test.py
@@ -102,11 +102,11 @@ def test_catboost_validation_set():
     booster = vaex.ml.catboost.CatBoostModel(features=features, num_boost_round=10, params=params_reg)
     # fit the booster - including saving the history of the validation sets
     booster.fit(train, 'E', evals=[train, test])
+    assert hasattr(booster, 'booster')
     assert len(booster.booster.evals_result_['learn']['MAE']) == 10
     assert len(booster.booster.evals_result_['learn']['R2']) == 10
     assert len(booster.booster.evals_result_['validation_0']['MAE']) == 10
     assert len(booster.booster.evals_result_['validation_0']['R2']) == 10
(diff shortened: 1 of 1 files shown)


tests/ml/catboost_test.py
```diff
@@ -101,12 +101,14 @@ def test_catboost_validation_set():
...
-    assert len(booster.booster.evals_result_['learn']['R2']) == 5
-    assert len(booster.booster.evals_result_['validation_0']['MAE']) == 5
-    assert len(booster.booster.evals_result_['validation_0']['R2']) == 5
-    assert booster.booster.best_iteration_ == 2
+    booster.fit(train, 'E', evals=[train, test])
+    assert len(booster.booster.evals_result_['learn']['MAE']) == 10
+    assert len(booster.booster.evals_result_['learn']['R2']) == 10
+    assert len(booster.booster.evals_result_['validation_0']['MAE']) == 10
+    assert len(booster.booster.evals_result_['validation_0']['R2']) == 10
+    assert hasattr(booster, 'booster')
```
maartenbreddels: Does this happen? If so, it should go before the `len(...)`, but I guess it never happens right?

JovanVeljanoski: Ah you are right.. it should be before the `len(...)`. It should never happen, bit it is added just in case `catboost` changes.

maartenbreddels: I don't wanna be a huge pain, but booster.booster is something we add right?

JovanVeljanoski: `.booster` is a property that is added to the `booster` object after it is fit. It does not exist when the object is instantiated, and it is nothing to do with `veaex-ml`. Checking for it is just another way of seeing if the fit was triggered correctly.

Issue #369, opened 2019-07-30
Problems with vaex to support Python3


fprada:
Once we installed vaex using conda the original Python3 was replaced by Python2 when call ipython. In principle vaex supports Python3, how to avoid this?

maartenbreddels:
What I suspect happening is:
 * you have a Python2 anaconda distribution for the root
 * you installed vaex in an environment (lets assume you call it vaex)
 * you've activated the vaex environment (I assume with Python3).

Now when you execute `$ ipython`, you may end up in the root Python environment, which is Python 2.
You can check this by executing `$ which ipython` (which will not e.g. ~/anaconda/envs/vaex/bin/ipython).
Is this correct?

fprada:
Hi again, here more details,

We have installed last anaconda version with python3 as root. 

When we follow your instructions:

conda install -c maartenbreddels vaex

Vaex is installed sucesfully but it downgrades our conda python version to 2.7.6

Then I can invoke python from conda that is 2.7.6 version, and I can import vaex with “import vaex”, but everything works in python 2.7.6

Can we use vaex in python3?

Regards.

P.S. Here is the attached message when we try to install vaex with conda:

----
[root@skun6 ~]# conda install -c maartenbreddels vaex
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: done
Solving environment: done

## Package Plan ##

  environment location: /usr/local/anaconda3

  added / updated specs:
    - vaex


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    _ipyw_jlab_nb_ext_conf-0.1.0|           py27_0           4 KB
    _libgcc_mutex-0.1          |             main           3 KB
    alabaster-0.7.12           |           py27_0          17 KB
    anaconda-client-1.7.2      |           py27_0         140 KB
    anaconda-navigator-1.9.7   |           py27_0         4.8 MB
    anaconda-project-0.8.3     |             py_0         212 KB
    aplus-0.11.0               |           py27_0           9 KB  maartenbreddels
    asn1crypto-0.24.0          |           py27_0         155 KB
    astroid-1.6.5              |           py27_0         402 KB
    astropy-2.0.9              |   py27hdd07704_0         6.8 MB
    atomicwrites-1.3.0         |           py27_1          13 KB
    attrdict-2.0.0             |           py27_0          18 KB  maartenbreddels
    attrs-19.1.0               |           py27_1          56 KB
[...]

PR #370, opened 2019-07-31
feat(core, hdf5): support python keywords and invalid variable names

maartenbreddels:
Implemented using a df._column_aliases dict that will translate those names
when possible.
Note that it means that you cannot do:
```
df['class * 2']  # this expression cannot be parsed
```
but instead, you can do
```
df['class'] * 2  # this can use the aliasing
```
If for some reason you need to write an expression using strings, use:
```
expression_string = str(df['class']) + ' * 2'
```
Closes #253,  should also help with #369

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -2762,7 +2762,7 @@ class DataFrame(object):
                         raise ValueError("Array is of length %s, while the length of the DataFrame is %s due to the filtering, the (unfiltered) length is %s." % (len(ar), len(self), self.length_unfiltered()))
                 raise ValueError("array is of length %s, while the length of the DataFrame is %s" % (len(ar), self.length_original()))
             # assert self.length_unfiltered() == len(data), "columns should be of equal length, length should be %d, while it is %d" % ( self.length_unfiltered(), len(data))
-            valid_name = vaex.utils.find_valid_name(name, used=self.get_column_names(hidden=True))
+            valid_name = vaex.utils.find_valid_name(name)
             if name != valid_name:
                 self._column_aliases[name] = valid_name
             self.columns[valid_name] = f_or_array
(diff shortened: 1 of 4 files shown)


maartenbreddels:
This is how it looks like in the notebook:
image

aanilpala:
there seems to be issues with column names containing "+" as well. One hot encoder usually gives rises to such situations.

aanilpala:
The workaround I've found and posted above won't work for some other non-python keywords than "+". Therefore, it is a very brittle solution as such characters are very typical in the text data that is passed to OHE. So, I had to come up with a better solution.

Turns out that all I needed to do was to get the column names with alias=False before passing them to the sklearn wrapper like in the following:

```
train_cols = [col for col in train.get_column_names(alias=False) if col != self.target_variable] 
target_col = self.target_variable
booster = train_lgbm_booster(hyper_params, train, valid=test, train_cols=train_cols, target_col=target_col)

vaex_lgbm_model = vaex.ml.sklearn.Predictor(
            features=train_cols,
            target=target_col,
            model=booster,
            prediction_name="pred")
```

maartenbreddels:
Hi,

did you try with the latest alpha, i've fixed a bunch more corner cases. If that still doesn't work, could you open an issue with an example to reproduce it?

cheers,

Maarten

aanilpala:
Hi, I haven't yet tried the alpha. I was waiting for the 4.0 release. Now that my problem is solved with get_column_names(alias=False), it is not an urgency anymore but I will try to see soon how the latest alpha behaves in a similar scenario.

aanilpala:
I've tried this with 4.0.0a16. Unfortunately, it was not possible to run my pipeline smoothly without adding some additional logic that takes care of the non-python-expression column names. With the alias flag removed from get_column_names, the following code was necessary to be able to do state_set for OHE encoding (without using the transform from OHE but rather loading the state from another dataframe - .transform still worked but I am interested in applying vaex's implicit pipelines using state transform)

```
cols_to_rename = \
            [col for col in df.get_column_names() if col not in self.cat_feats and not valid_identifier(col)]

        for col in cols_to_rename:
            suggested_colname = re.sub('\W|^(?=\d)','_', col)
            df.rename(col, suggested_colname, unique=True)
```

aanilpala:
I will try to add an example to reproduce it soon.

aanilpala:
here: https://github.com/vaexio/vaex/issues/1200

PR #371, opened 2019-07-31
Remove semicolon from python example code

cirosantilli:
Unacceptable! :-)


maartenbreddels:
There is a reason for it! :)
since it returns the return value of matplotlib's scatter, it will print it out, which looks ugly.

PR #381, opened 2019-08-22
Adding scikit-learn with specific version to make testing more robust.

JovanVeljanoski:
This is in order to help with the testing of the `vaex-ml` package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
   - conda update -q conda
   - conda info -a
   # future: kapteyn should only be needed for astro package testing
-  - conda create -q -n test-environment -c conda-forge python=$PYTHON_VERSION pip numpy scipy pyqt matplotlib pyopengl h5py numexpr astropy tornado cython runipy pandas pytest numba pyarrow graphviz python-graphviz pcre catboost $OPTIONAL_PACKAGES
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Why is this needed? Was the test running with a too new version or too old?

JovanVeljanoski:
The assertion test checking the consistency between the results in `vaex` and `scikit-learn` for`RobustScaler` in `vaex-ml` was failing for Linux (python-3.6) only.  This way the `scikit-learn` version will be fixed to the latest one, and should remove flakiness. 

Previously, `scikit-learn` was not explicitly installed, but came as a dependency of other packages, which can vary as those packages change. 

In fact, I would be in favour of fixing the versions of some of the dependencies explicitly, especially the optional ones. I should make the tests more robust and things easier to debug.

maartenbreddels:
> In fact, I would be in favour of fixing the versions of some of the dependencies explicitly, especially the optional ones. I should make the tests more robust and things easier to debug.

I'm against that, you could do a minimal version, but we do want to test against new versions as well.

JovanVeljanoski:
The problem does not seem to lie with the `scikit-learn` versioning. I will investigate this further.

JovanVeljanoski:
The problem of why `RobustScaler` fails for python3.6 on Linux is due to the `percentile_approx` returning `nan`. 

This was also reported by #385. Thus I will close this PR, and open a new one regarding that particular issue.

PR #386, opened 2019-08-25
percentile_approx issue fix for python3.6 on Linux 

JovanVeljanoski:
This PR is for the handling of the issue of `percentile_approx` method that pops up only on Linux and for python3.6.

### DIFF
--- a/tests/percentile_approx_test.py
+++ b/tests/percentile_approx_test.py
@@ -1,0 +1,11 @@
+import vaex.ml
+import numpy as np
+
+def test_percentile_approx():
+    df = vaex.ml.datasets.load_iris()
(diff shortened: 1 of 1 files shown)


JovanVeljanoski:
I introduced yet another test via pytest for faster iterations.

JovanVeljanoski:
This is a most bizarre issue. While debugging, I noticed that putting print statements of certain variables alleviates the issue for python3.6 on Linux, but now Travis tells us that doing that creates a problem for Windows, again python3.6. 

My guess this is either something to do with the `@delay` decorator, or with `vaexfast.grid_find_edges`. 

I think I have narrowed down the issue there.. @maartenbreddels can you please take a quick look?

PR #390, opened 2019-08-28
Test for join when both datasets are filtered.

JovanVeljanoski:
This PR adds a unit-test for `join` which exposes an issue which happens when joining two datasets, both of which are filtered. 

Apart from the formal assertions that is done in the unit-test, a more complete error message might be seen if one simply does:
```print(df_joined)```


maartenbreddels:
The fundamental issue here is that we want to merge two dataframes with different filters. We avoid the issue when the left is filtered by actually joining as if the left was not filtered at all.

We can do the same for the right, but there it is important that we make sure we don't actually accidentally join to a row that was filtered out.

maartenbreddels:
False positive, this was already fixed in https://github.com/vaexio/vaex/pull/389

PR #400, opened 2019-09-02
Unit-test: add pd.NaT towards the total count of NA in describe()

JovanVeljanoski:
At the moment the `describe` method does not detect `pandas.NaT` (not a time) values as N/A. This unit-test exposes this.


JovanVeljanoski:
Note that the `.countna()` method correctly counts the number of not-available values (i.e. it detects and includes the `pd.NaT` values in the total sum.

PR #430, opened 2019-09-26
Support for selections in vaex.agg functions

JovanVeljanoski:
With this PR we add `selection` arguments to the aggregators, i.e. `vaex.agg` methods. 
This will enable very flexible way of aggregating columns, for instance

```
df.groupby(df.x).agg({'y_mean_sel': vaex.agg.mean(df.y, selection='z > 10')})
```

In the above example, the `y_mean_sel` column will have the mean only of those values that pass the selection condition. This can be done per aggregation operation, offering great flexibility.

### DIFF
--- a/packages/vaex-core/vaex/agg.py
+++ b/packages/vaex-core/vaex/agg.py
@@ -28,11 +28,12 @@ class AggregatorDescriptor(object):
 
 
 class AggregatorDescriptorBasic(AggregatorDescriptor):
-    def __init__(self, name, expression, short_name, multi_args=False, agg_args=[]):
+    def __init__(self, name, expression, short_name, multi_args=False, agg_args=[], selection=None):
         self.name = name
         self.short_name = short_name
         self.expression = str(expression)
         self.agg_args = agg_args
+        self.selection = selection
         if not multi_args:
             if self.expression == '*':
                 self.expressions = []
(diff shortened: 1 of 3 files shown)


ManuelZ:
Hi, what kind of syntax does selection supports?

JovanVeljanoski:
Just the same as any other method that supports the `selection` kwarg. 

For example:

```
import vaex.ml
df = vaex.ml.datasets.load_titanic()
df.groupby('pclass').agg({'age_f': vaex.agg.mean('age', selection='sex == "female"'),
                          'age_m': vaex.agg.mean('age', selection=df.sex == "male"),
                          'age_f_surv': vaex.agg.mean('age', selection='(sex=="female") & (survived==True)'),
                          'age_m_surv': vaex.agg.mean('age', selection=((df.sex == "male") & (df.survived==True))),
                         })

```

And the output is 
image


Hope this makes sense!

ManuelZ:
Oh nice thank you! Your last example helps a lot, I though that I only could write a string there.

JovanVeljanoski:
You are welcome. Same goes for any other method that supports the `selection` argument (like plotting, `df.count`, `df.mean` etc)

ManuelZ:
Another question. How can I create my own agg function? I need the mode, but I can't find it in [agg.html](https://vaex.readthedocs.io/en/latest/_modules/vaex/agg.html)

PR #435, opened 2019-10-02
fix: upcast dtype of sum aggregators to avoid overflow

### LINKED ISSUE #428: dtype automatic conversion when needed
There is some inconsistent behaviour between `numpy` and `vaex` when it comes to operations on floats and ints that have lower bits than the result. Consider the following example:

```
import vaex
import numpy as np

x = np.ones(500, dtype=np.int8)
df = vaex.from_arrays(x=x)

print('Numpy:', x.sum())
print('Vaex:', df.x.sum())
print('Vaex (manual conversion):', df.x.astype('int').sum())
```
For the operation `x.sum()`, `numpy` uses the default `int64` type even though `x` is of type `int8`, and will return a `int64`. On the other hand, `vaex` tries to do the same, however the output of operations such as `sum` or `mean` return an `array`  with the same `dtype` as the column/expression in which this method was applied. In the above example, the result of the sum (500) will be converted to an array of type `int8`, hence the value -12.

The question is how do we want to handle this case. Should all output arrays be cast to the python default `int64` or `float64`, or do we want to keep this behaviour?

maartenbreddels:
Fixes #428

### DIFF
--- a/packages/vaex-core/src/superagg.cpp
+++ b/packages/vaex-core/src/superagg.cpp
@@ -456,10 +456,70 @@ public:
     }
 };
 
+template<class T>
+struct upcast {
+};
+
+template<>
+struct upcast<float> {
+    typedef double type;
+};
+
+template<>
+struct upcast<double> {
+    typedef double type;
+};
+
+template<>
+struct upcast<bool> {
+    typedef int64_t type;
+};
+
+template<>
+struct upcast<int8_t> {
+    typedef int64_t type;
+};
+
+template<>
+struct upcast<int16_t> {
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Given a dataframe:
```python
df = vaex.from_arrays(b=[False, True, True])
```
Calculating a sum would not upcast it, and you had to manually do
```
df.b.astype('int').sum()
```
Now we can simply do:
```
df.b.sum()
```
casting rules are simple:
bool->int64, intX->int64, uintX->int64, floatX->float64

Issue #375, opened 2019-08-02
How to make a selection like .isin() in pandas?


jackyckma:
It seems like a typical use case. Shall I use vaex.register_function dectorator to add this functionality? or it is indeed possible with current API?

jackyckma:
just settled with this mapping and select approach:

```
selected_items = {'itemA': 1, 'itemB': 1}
df['select'] = df.item_key.map(selected_items, default_value=0, allow_missing=True)
df.y_value.max(selection=[None, df.select==0, df.select==1])
```

maartenbreddels:
`isin()` is not supported yet, but a must-have, I'll keep this open till then.

JovanVeljanoski:
Closed via #441

PR #441, opened 2019-10-11
Feat: add isin() method

JovanVeljanoski:
This PR implements the `isin()` method. 

The test is done following the expected results form `pandas`.  Important to note the `np.nan`, and the `None` behaviour. 

Related to #375

### DIFF
--- a/packages/vaex-core/src/strings.cpp
+++ b/packages/vaex-core/src/strings.cpp
@@ -367,6 +367,43 @@ class StringSequenceBase : public StringSequence {
         }
         return std::move(matches);
     }
+    py::object isin(const StringSequence* others) {
+        py::array_t<bool> matches(length);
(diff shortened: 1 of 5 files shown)


JovanVeljanoski:
This now works, however when `isin` is used on numerical columns there are some artifacts printed which should probably not be there. Here is an example:

image

Otherwise it is all good!

stokhos:
Following code returns an incorrect result

```python
df=vx.from_arrays(x=np.array([1.01,1.02,1.03]))

df.x.isin([1])
```

Expression = isin_set(x, var_isin_ordered_set)
Length: 3 dtype: bool (expression)
----------------------------------
0  True
1  True
2  True

should be all false

maartenbreddels:
Hi,

thanks for reporting this! Would you mind opening a new issue, otherwise this might get lost.

Regards,

Maarten

PR #454, opened 2019-10-25
Test github action

maartenbreddels:


maartenbreddels:
Failure was unrelated, we still need osx and windows though, that's why we keep travis and appveyor

Issue #448, opened 2019-10-20
Timedelta mean/std consistency


JovanVeljanoski:
Something we need to consider regarding the output of `mean` and `std` methods when applied to `timedelta64` type columns. Let's consider this example:
<img width="709" alt="image" src="https://user-images.githubusercontent.com/18574951/67162863-c5ab5880-f368-11e9-82bc-0182c8cb1d9c.png">

Let's calculate the mean and stdev of the timedelta column, also as a part of a `groupby` operation:
<img width="644" alt="image" src="https://user-images.githubusercontent.com/18574951/67162869-eecbe900-f368-11e9-964b-40d60f31b97e.png">

See the output of the `std` operation is not nicely formatted compared to the output of the `mean` method. This should be consistent. Also, simply doing 
```
groupby_df.delta_std.astype('timedelta64[s]')
```
<img width="464" alt="image" src="https://user-images.githubusercontent.com/18574951/67162949-c09ad900-f369-11e9-8703-66ee43896e70.png">
gives incorrect results. This can be checked against pandas:
<img width="469" alt="Screen Shot 2019-10-20 at 18 43 53" src="https://user-images.githubusercontent.com/18574951/67162928-9ea15680-f369-11e9-93a7-af8c781bb9c9.png">

Not that also Vaex is ignoring the decimal places of the `delta_mean` column in the groupby result. This is however mostly likely just a formatting style difference. 

This should not be a difficult issue to fix. @maartenbreddels let me know how we want to proceed here, and I will make the appropriate unit-test.

PR #458, opened 2019-11-01
feat: restore dtype for std aggregation

maartenbreddels:
@JovanVeljanoski this should improve #448 a lot, I'll leave this for you to test/review.


JovanVeljanoski:
Formatting looks great, but I am afraid the numerical correctness is a bit off. 
There is a `groupby` test failing. I also added an additional using timedelta columns which are being aggregated.

PR #462, opened 2019-11-19
Fillna df.mean() compatibility test

xdssio:
Desired functionality for fillna in a standard use case.
```
# Current implementation
df['x'] = df.x.fillna(float(df.x.mean()))
```
vs
```
# Desired functionality
df['x'] = df.x.fillna(df.x.mean())
```

- [ ] implement new functionality 
- [ ] add unit-test 
- [ ] review


maartenbreddels:
This was not a specific problem with fillna/mean, but with and function added to a dataframe. I've made the test and code more general.

PR #463, opened 2019-11-19
Enable multi-key sorting of a DataFrame

JovanVeljanoski:
- [x] Enable multi-key sorting of a DataFrame
- [x] Add unit-test
- [x] Update docstring to account for the changes.

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -3927,6 +3927,8 @@ class DataFrame(object):
     def sort(self, by, ascending=True, kind='quicksort'):
         '''Return a sorted DataFrame, sorted by the expression 'by'
 
+        The kind keyword is ignored if doing multi-key sorting.
+
         {note_copy}
 
         {note_filter}
@@ -3954,8 +3956,14 @@ class DataFrame(object):
         :param str kind: kind of algorithm to use (passed to numpy.argsort)
         '''
         self = self.trim()
-        values = self.evaluate(by)
-        indices = np.argsort(values, kind=kind)
+        if not isinstance(by, list):
+            values = self.evaluate(by)
+            indices = np.argsort(values, kind=kind)
+        if isinstance(by, list):
+            by = _ensure_strings_from_expressions(by)[::-1]
+            by = ', '.join(by)
+            values = self.evaluate(by)
+            indices = np.lexsort(values)
         if not ascending:
             indices = indices[::-1].copy()  # this may be used a lot, so copy for performance
         return self.take(indices)
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -3959,8 +3959,8 @@ class DataFrame(object):
         if not isinstance(by, list):
             values = self.evaluate(by)
             indices = np.argsort(values, kind=kind)
-        if isinstance(by, list):
-            by = _ensure_strings_from_expressions(by)[::-1]
+        if isinstance(by, (list, tuple)):
+            by = _ensure_strings_from_expressions(by)[::-1] # TODO: this is using an unsupported features, when evaluate takes a list of expressions, we should change this.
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/dataframe.py
```diff
@@ -3954,8 +3956,14 @@ def sort(self, by, ascending=True, kind='quicksort'):
         :param str kind: kind of algorithm to use (passed to numpy.argsort)
         '''
         self = self.trim()
-        values = self.evaluate(by)
-        indices = np.argsort(values, kind=kind)
+        if not isinstance(by, list):
+            values = self.evaluate(by)
+            indices = np.argsort(values, kind=kind)
+        if isinstance(by, list):
```
maartenbreddels:
```suggestion
        if isinstance(by, (list, tuple)):
```
Is more standard

packages/vaex-core/vaex/dataframe.py
```diff
@@ -3954,8 +3956,14 @@ def sort(self, by, ascending=True, kind='quicksort'):
...
+            by = _ensure_strings_from_expressions(by)[::-1]
+            by = ', '.join(by)
```
maartenbreddels: This is actually not supported, I'm working on a parallel evalulate that takes a list of expressions. I'm ok with merging this, but we need to put a reminder
```suggestion
            by = ', '.join(by)  # TODO: this is using an unsupported features, when evaluate takes a list of expressions, we should change this.
```

tests/sort_test.py
```diff
@@ -26,3 +26,19 @@ def test_sort_filtered():
...
+
+def test_sort_multikey():
+    x = np.array([5, 3, 1, 1, 5])
+    y = np.array([0, 3, 4, 2, 1])
+    z = np.array(['dog', 'cat', 'cat', 'dog', 'mouse'])
+    df = vaex.from_arrays(x=x, y=y, z=z)
+
+    # Case 1: numeric keys
+    df_sorted_1 = df.sort(by=['x', 'y'])
+    assert df_sorted_1.values.tolist() == [[1, 2], [1, 4], [3, 3], [5, 0], [5, 1]]
```
maartenbreddels: You are assuming that the order of the columns is x, then y. This is not supported before Python3.6. Use from_items to ensure order. Of simply use df_sorted_1.x.tolist() and y separate

packages/vaex-core/vaex/dataframe.py
maartenbreddels: Note that the unittest will catch it, but it's good to have this in the code.

tests/sort_test.py
```diff
@@ -26,3 +26,19 @@ def test_sort_filtered():
...
+    assert df_sorted_1.z.tolist() == ['dog', 'cat', 'cat', 'dog', 'mouse']
+
+    # Case 2: str key followed by numeric key
+    df_sorted_2 = df.sort(by=['z', 'x'])
+    assert df_sorted_2.values.tolist() == [[1, 4], [3, 3], [1, 2], [5, 0], [5, 1]]
+    assert df_sorted_2.z.tolist() == ['cat', 'cat', 'dog', 'dog', 'mouse']
```
maartenbreddels: cool!

JovanVeljanoski:
Thanks for the comments @maartenbreddels, useful stuff. I will get it sorted out soon.

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: Good idea - implemented.

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: Fair point - comment added.

tests/sort_test.py
JovanVeljanoski: Ah..i keep on forgetting that. Now there is an assertion for each column.

PR #467, opened 2019-11-21
feat: Arrow string integration

maartenbreddels:
Since arrow >= 0.15 has support for large_string types, we can now start relying on pyarrow. The ultimate goal is that vaex-core will depend on pyarrow, and use as much as it's datastructures. Here we start by using the StringArray. After this we should try to move the string algorithms in vaex upstream to arrow.


packages/vaex-core/vaex/column.py
```diff
@@ -107,6 +113,14 @@ def __len__(self):
     def trim(self, i1, i2):
         return ColumnIndexed(self.df, self.indices[i1:i2], self.name)
 
+    def __arrow_array__(self, type=None):
+        # TODO: without a copy we get a buserror
+        values = self[:]
+        if hasattr(values, "to_arrow"):
+            return values.to_arrow()
+        else:
+            return pa.array(values)
```
xhochy: The bus error is probably a stackoverflow as `pa.array` would otherwise call the same `__arrow__array__()` again

maartenbreddels: Yes, I think so too. I tried to reproduce how I got it, but failed :(
Thanks for the review!

maartenbreddels:
The idea is explored more in #517, but now with any Apache Arrow data.

PR #469, opened 2019-11-22
chore(ci): use GitHub actions only

maartenbreddels:


maartenbreddels:
This will have to do for now, lets keep travis and appveyor running for now, but merge when the GitHub actions are green.

Issue #259, opened 2019-05-15
Materialize and/or to_pandas_df in parallel.


melovv:
Hello, guys.

I just found Vaex and am trying to use it in a project. I'm more interested in the expression system. 

Right now, I have a table with tens of virtual columns and would like to materialize them in parallel and/or export them to a pandas dataframe in parallel. Is that possible? When I run df.to_pandas_df(virtual=True) it is using a single core.

Thanks a lot!

Vinícius.

maartenbreddels:
Hi Vinicius,

that is a good observations, at the core of the problem is `DataFrame.evaluate` which is single threaded. In the early days this was the core entry point which all threads used to evaluate, hence it didn't multithread again. This is a nuisance for me as well, and high on my wishlist.
A few workarounds:
 * export to hdf5, only useful if too large to fit into memory and you plan to reuse it
 * use df.map_reduce, sth like this:
```python
import vaex
import numpy as np
df = vaex.open('/Users/maartenbreddels/datasets/nyc_taxi2015.arrow')
df['r'] = df.dropoff_latitude + 1
df['q'] = df.dropoff_longitude + 1
virtual_columns = ['r', 'q']

outputs = [np.zeros(len(df), dtype=df.dtype(k)) for k in virtual_columns]
                    
def map(thread_index, i1, i2, *chunks):
    for output, chunk in zip(outputs, chunks):
        output[i1:i2] = chunk
def reduce(a, b):
    pass
df.map_reduce(map, reduce, virtual_columns, info=True)
```
This ensures all chunks are done in parallel, with 1 pass over the data. The last point is important when the expressions depend a lot on the same columns.

Does this help?

I'll keep this issue open until we have a nice solution + documentation on how to do this properly.

Cheers,

Maarten

melovv:
Hi, Maarten.

Thanks for the quick reply and the code. It runs super fast!
I'm gonna open another ticket regarding another topic.

Cheers,

Vinícius.

maartenbreddels:
This should be solved in #474

PR #474, opened 2019-11-25
feat(core): parallel evaluate

maartenbreddels:
Evaluate has become quite a beast, but there need to be 2 modes, parallel (default for users), and non-parallel (interally mostly, but also if someone else already does the multithreading).

For vaex internally, if it's in a task (say map reduce), it can call `df.dtype(..)`, which may call `df.evaluate`, which will start another task, which is not supported.

These are the two reasons why evaluate has a parallel and non-parallel mode.


maartenbreddels:
This should close #259

PR #479, opened 2019-11-27
Adding a GroupByTransformer in vaex.ml

JovanVeljanoski:
Adding a `vaex.ml.GroupByTransformer`. This transformer creates aggregation of features just like a `groupby` operation which are then added to a DataFrame. With the `.transform` method, the same aggregations can be added to different DataFrames (e.g. validation, test sets).
This is useful for creating aggregate features useful for certain ML tasks.

- [x] Implement `GroupbyTransformer`
- [x] Implement alternative API
- [x] Unit-test
- [ ] Code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-# vaex 3.0.x (unreleased)
+# vaex 3.1.0 (unreleased)
+
+## vaex-ml 0.10.0
+   * Features
+      * Implementation of `GroupbyTransformer` [#479](https://github.com/vaexio/vaex/pull/479)
 
 # vaex-hdf5 0.6.1 (2020-6-2)
    * Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 # vaex-core 2.0.2 (2020-6-2)
    * Fixes
       * Masked arrays supported in hdf5 files on s3
+      * Expression.map always uses masked arrays to be state transferrable (a new dataset might have missing values) [#479](https://github.com/vaexio/vaex/pull/479)
 
 # vaex-core 2.0.1 (2020-5-28)
    * Fixes
--- a/packages/vaex-ml/vaex/ml/__init__.py
+++ b/packages/vaex-ml/vaex/ml/__init__.py
@@ -102,6 +102,19 @@ class DataFrameAccessorML(object):
         minmax_scaler.fit(self.df)
         return minmax_scaler
 
+    def groupby_transformer(self, by, agg, rprefix='', rsuffix=''):
+        '''Requires vaex.ml: Create :class:`vaex.ml.transformations.GroupByTransformer` and fit it.
+
+        :param by: The feature on which to do the grouping.
+        :param agg: Dict where the keys are feature names and the values are vaex.agg objects.
+        :param rprefix: Prefix for the names of the aggregate features in case of a collision.
+        :param rsuffix: Suffix for the names of the aggregate features in case of a collision.
+        :return vaex.ml.transformations.GroupByTransformer: Fitted GroupByTransformer.
+        '''
+
+        group_trans = GroupByTransformer(by=by, agg=agg, rprefix=rprefix, rsuffix=rsuffix)
+        group_trans.fit(self.df)
+        return group_trans
 
     def xgboost_model(self, target, features=None, num_boost_round=100, params={}, prediction_name='xgboost_prediction'):
         '''Requires vaex.ml: create a XGBoost model and train/fit it.
@@ -230,3 +243,4 @@ from .transformations import LabelEncoder, OneHotEncoder, FrequencyEncoder
 from .transformations import CycleTransformer
 from .transformations import BayesianTargetEncoder
 from .transformations import WeightOfEvidenceEncoder
+from .transformations import GroupByTransformer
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
+# vaex (unreleased)
```
maartenbreddels:
```suggestion
# vaex 2.5.0-dev (unreleased)
```

packages/vaex-ml/vaex/ml/_version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = '0.6.2'
-__version_tuple__ = (0, 6, 2)
+__version__ = '0.7.0-dev'
+__version_tuple__ = (0, 7, 0)
```
maartenbreddels:
```suggestion
__version_tuple__ = (0, 7, 0, 'dev')
```

packages/vaex-ml/vaex/ml/__init__.py
```diff
@@ -103,6 +103,20 @@ def minmax_scaler(self, features=None, feature_range=[0, 1], prefix='minmax_scal
         return minmax_scaler
 
 
+    def groupby_transformer(self, by, agg, rprefix='', rsuffix=''):
```
maartenbreddels: we dont autogen these?

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -573,3 +573,65 @@ def transform(self, df):
                 expr = expr / self.scale_[i]
             copy[name] = expr
         return copy
+
+
+@register
+@generate.register
+class GroupByTransformer(Transformer):
+    '''The GroupByTransformer creates aggregations via the groupby operation, which are
```
maartenbreddels: """ is default for docstrings

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -573,3 +573,65 @@ def transform(self, df):
...
+      1  cat      5  cat      15.0
+      2  dog      5  dog      3.0
+      3  mouse    5  --       --
+    '''
+
+    by = traitlets.Unicode(allow_none=False, help='The feature on which to do the grouping.')
+    agg = traitlets.Dict(help='Dict where the keys are feature names and the values are vaex.agg objects.')
+    rprefix = traitlets.Unicode(default_value='', help='Prefix for the names of the aggregate features in case of a collision.')
+    rsuffix = traitlets.Unicode(default_value='', help='Suffix for the names of the aggregate features in case of a collision.')
+    df_group_ = traitlets.Instance(klass=vaex.dataframe.DataFrame, allow_none=True)
```
maartenbreddels: Does this mean we cannot serialize it?

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -573,3 +573,65 @@ def transform(self, df):
...
+        Transform a DataFrame with a fitted GroupByTransformer.
+
+        :param df: A vaex DataFrame.
+
+        :returns copy: a shallow copy of the DataFrame that includes the aggregated features.
+        :rtype: DataFrame
+        '''
+
+        copy = df.copy()
+        copy = copy.join(other=self.df_group_, on=self.by, how='left', rprefix=self.rprefix, rsuffix=self.rsuffix)
```
maartenbreddels: my  guess is join had inplace=False as default, so no copy needed?

packages/vaex-ml/vaex/ml/__init__.py
JovanVeljanoski: Not in this API. We do in the sklearn-like API, where we use the traitlets..

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: None of the others have it tho.. i was just following what we had so far.. 

Should I change everything?

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Ah indeed that is correct

JovanVeljanoski:
@maartenbreddels you are right, at the moment there is a problem with the serialisation. Do you know what is causing this? If it is the storing of the grouped dataframe.. in principle I can save it as dict, and then during transform, convert to a DataFrame object and merge it.

Or do you have another proposal?

In any case there is a test exposing this

JovanVeljanoski:
NB: Tests also fail due to a `catboost` issue, and that is tracked in #481

maartenbreddels:
I think this is a general problem, for which I'm not sure how we want to attack it. Out first solution can be specific to this transformer, by doing to_json/from_json. That will not serialise dates for instance. Long term solution would be to serialise it using Arrow. If we want to make this backwards compatible, we want to store how we store it (JSON/arrow) inside the serialisation JSON blob. 
Note that we are growing 2 serialisation frameworks now, one in vaex-core and one in vaex-ml. We may want to bring those together.

packages/vaex-ml/vaex/ml/transformations.py
maartenbreddels: lets keep it then

JovanVeljanoski:
Looks great!

PR #480, opened 2019-11-28
Duplicated columns when joining on the same column name

JovanVeljanoski:
Remove duplications of columns when joining DataFrames on the same column name:

- [x] Unit-test
- [ ] Modify join
- [ ] Review


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
+# vaex 2.5.0-dev (unreleased)
```
maartenbreddels: I'll fix this, but it would be 2.4.1, since it's just a fix

PR #487, opened 2019-12-09
Improvements to the ml sklearn wrapper.

JovanVeljanoski:
Small improvements to the sklearn wrapper:

- [x] implemented `.fit_transform` method for convenience
- [x] implemented `.fit_predict` method for convenience
- [x] while doing transform, the lazy function is set to be unique. This allows for multiple sklearn-like models to be used.
- [x] small updates to the tests to improve code coverage
- [ ] review

### DIFF
--- a/packages/vaex-ml/vaex/ml/sklearn.py
+++ b/packages/vaex-ml/vaex/ml/sklearn.py
@@ -1,7 +1,3 @@
-import pickle
-import base64
-import tempfile
-
 import traitlets
 import numpy as np
 
@@ -10,11 +6,12 @@ import vaex.serialize
 from vaex.ml import state
 from vaex.ml import generate
 from vaex.ml.state import serialize_pickle
+from vaex.ml.transformations import Transformer
 
 
 @vaex.serialize.register
 @generate.register
-class SKLearnPredictor(state.HasState):
+class SKLearnPredictor(Transformer):
     '''This class wraps any scikit-learn estimator (a.k.a predictor) making it a vaex pipeline object.
 
     By wrapping any scikit-learn estimators with this class, it becoes a vaex
@@ -81,7 +78,7 @@ class SKLearnPredictor(state.HasState):
         :rtype: DataFrame
         '''
         copy = df.copy()
-        lazy_function = copy.add_function('sklearn_prediction_function', self)
+        lazy_function = copy.add_function('sklearn_prediction_function', self, unique=True)
         expression = lazy_function(*self.features)
         copy.add_virtual_column(self.prediction_name, expression, unique=False)
         return copy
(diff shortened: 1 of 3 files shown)


packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +91,24 @@ def fit(self, df, target, **kwargs):
         X = df[self.features].values
         y = df.evaluate(target)
         self.model.fit(X=X, y=y, **kwargs)
+
+    def fit_transform(self, df, target, **kwargs):
```
maartenbreddels: I don't think it makes sense to have this in this class, I'd say some base class in vaex.ml should do this so all transformers have it? (in fact we do, ml/transformers.py::Transformer.

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +91,24 @@ def fit(self, df, target, **kwargs):
...
+
+    def fit_predict(self, df, target):
+        '''Fit the SKLearnPredictor and immediately obtain the predictions on the supplied DataFrame.
+        :param df: A vaex DataFrame containing the features on which to train the model.
+        :param target: The name of the column containing the target variable.
+        :return copy: A shallow copy of the DataFrame that includes the SKLearnPredictor prediction as a virtual column.
+        :rtype: DataFrame
+        '''
+        self.fit(df=df, target=target)
+        return self.predict(df=df)
```
maartenbreddels: same applies here

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -31,14 +31,15 @@ class Transformer(HasState):
...
+    def fit_transform(self, df, target=None):
         '''Fit and apply the transformer to the supplied DataFrame.
 
         :param df: A vaex DataFrame.
+        :param target: A vaex expression containing the target variable.
 
         :returns copy: A shallow copy of the DataFrame that includes the transformations.
         '''
-        self.fit(df=df)
+        self.fit(df=df, target=target)
```
maartenbreddels:
```suggestion
        if target is None:
            self.fit()
        else:
            self.fit(df=df, target=target)
```

JovanVeljanoski: I was following the example set py `scikit-learn`. I prefer that to your suggestion. Since in veax all these objects contain the `.transform` method, it makes sense that thay subclass form the same base. 

Alternatively, one could implement a different base for predictors/estimators, but they would also have the same methods as a transformer would + `.predict`.  So I don't know what is the best choice here. For me following the `sklearn` style made sense.

JovanVeljanoski:
I will close this PR since at this time we can't agree on the best way to proceed.

PR #497, opened 2019-12-10
feat(ml): IncrementalPredictor

JovanVeljanoski:
This PR adds a `veax.ml.sklearn.IncrementalPredictor` class that wraps any `scikit-learn` estimator (a.k.a predictor) that has the `.partial_fit` method. 

This means that we can send batches of data from `vaex` to such a predictor, while controlling the memory needed. 

Each batch can be "seen" by the predictor multiple times (epochs), and optionally it can be shuffled.

- [x] Implementation
- [x] Unit-tests
- [ ] Documentation
- [ ] Code review

### DIFF
--- a/packages/vaex-ml/vaex/ml/sklearn.py
+++ b/packages/vaex-ml/vaex/ml/sklearn.py
@@ -95,3 +95,146 @@ class SKLearnPredictor(state.HasState):
         X = df[self.features].values
         y = df.evaluate(target)
         self.model.fit(X=X, y=y, **kwargs)
+
+
+@vaex.serialize.register
+@generate.register
+class IncrementalPredictor(state.HasState):
+    '''This class wraps any scikit-learn estimator (a.k.a predictions) that has
+    a `.partial_fit` method, and makes it a vaex pipeline object.
+
+    By wrapping "on-line" scikit-learn estimators with this class, they become a vaex
+    pipeline object. Thus, they can take full advantage of the serialization and
+    pipeline system of vaex. While the underlying estimator need to call the
+    `.partial_fit` method, this class contains the standard `.fit` method, and
+    the rest happens behind the scenes. One can also iterate over the data
+    multiple times (epochs), and optionally shuffle each batch before it is sent
+    to the estimator. The `predict` method returns a numpy array, while the `transform`
+    method adds the prediction as a virtual column to a vaex DataFrame.
+
+    Note: the `.fit` method will use as much memory as needed to copy one
+    batch of data, while the `.predict` method will require as much memory as
+    needed to output the predictions as a numpy array. The `transform` method is
+    evaluated lazily, and no memory copies are made.
+
+    Note: we are using normal sklearn without modifications here.
+
+    Example:
+
+    >>> import vaex
+    >>> import vaex.ml
+    >>> from vaex.ml.sklearn import IncrementalPredictor
+    >>> from sklearn.linear_model import SGDRegressor
+    >>>
+    >>> df = vaex.example()
+    >>>
+    >>> features = df.column_names[:6]
+    >>> target = 'FeH'
+    >>>
+    >>> standard_scaler = vaex.ml.StandardScaler(features=features)
+    >>> df = standard_scaler.fit_transform(df)
+    >>>
+    >>> features = df.get_column_names(regex='^standard')
+    >>> model = SGDRegressor(learning_rate='constant', eta0=0.01, random_state=42)
+    >>>
+    >>> incremental = IncrementalPredictor(model=model,
+    ...                                    features=features,
+    ...                                    batch_size=10_000,
+    ...                                    num_epochs=3,
+    ...                                    shuffle=True,
+    ...                                    prediction_name='pred_FeH')
+    >>> incremental.fit(df=df, target=target)
+    >>> df = incremental.transform(df)
+    >>> df.head(5)[['FeH', 'pred_FeH']]
+      #        FeH    pred_FeH
+      0  -2.30923     -1.66226
+      1  -1.78874     -1.68218
+      2  -0.761811    -1.59562
+      3  -1.52088     -1.62225
+      4  -2.65534     -1.61991
+    '''
+
+    model = traitlets.Any(default_value=None, allow_none=True, help='A scikit-learn estimator with `.fit_predict` method.').tag(**serialize_pickle)
+    features = traitlets.List(traitlets.Unicode(), help='List of features to use.')
+    batch_size = traitlets.Int(default_value=1_000_000, allow_none=False, help='Number of samples to be sent to the model in each batch.')
+    num_epochs = traitlets.Int(default_value=1, allow_none=False, help='Number of times each batch is sent to the model.')
+    shuffle = traitlets.Bool(default_value=False, allow_none=False, help='If True, shuffle the samples before sending them to the model.')
+    prediction_name = traitlets.Unicode(default_value='prediction', help='The name of the virtual column housing the predictions.')
+
+    def __call__(self, *args):
+        X = np.vstack([arg.astype(np.float64) for arg in args]).T.copy()
+        return self.model.predict(X)
+
+    def predict(self, df):
+        '''Get an in-memory numpy array with the predictions of the SKLearnPredictor.self
+
+        :param df: A vaex DataFrame, containing the input features.
+        :returns: A in-memory numpy array containing the SKLearnPredictor predictions.
+        :rtype: numpy.array
+        '''
+
+        return self.transform(df)[self.prediction_name].values
+
+    def transform(self, df):
+        '''Transform a DataFrame such that it contains the predictions of the IncrementalPredictor.
+        in form of a virtual column.
+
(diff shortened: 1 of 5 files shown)


packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
         X = df[self.features].values
         y = df.evaluate(target)
         self.model.fit(X=X, y=y, **kwargs)
+
+
+@vaex.serialize.register
+@generate.register
+class IncrementalPredictor(state.HasState):
+    '''This class wraps any scikit-learn estimator (a.k.a predictions) that has
+    a `.partial_fit` method, and makes it a vaex pipeline object.pipeline
```
maartenbreddels:
```suggestion
    a `.partial_fit` method, and makes it a vaex pipeline object.
```

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+
+    By wrapping "on-line" scikit-learn estimators with this class, they becoes a vaex
```
maartenbreddels:
```suggestion
    By wrapping "on-line" scikit-learn estimators with this class, they become a vaex
```

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+    pipeline object. Thus, they can take full advantage of the serialization and
+    pipeline system of vaex. While the underlying estimator need to call the
+    `.partial_fit` method, this class contains the standard `.fit` method, and
+    the rest happens behind the scenes. One can also iterate over the data
+    multiple times (epochs), and optionally shuffle each batch before it is sent
+    to the estimator. The `predict` method returns a numpy, while the `transform`
```
maartenbreddels:
```suggestion
    to the estimator. The `predict` method returns a numpy array, while the `transform`
```

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+    method adds the prediction as a virtual column to a vaex DataFrame.
```
maartenbreddels: this is general docs, should not be put here right?

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+
+    Note that the `.fit` method will use as much memory as needed to copy one
+    batch of data, while the `.predict` method makes a full memory copy.
+    The `transform` method is evaluated lazily, and no memory copies are made.
+
+    The scikit-learn estimators themselves are not modified at all, they are
+    taken from your local installation of scikit-learn.
+
```
maartenbreddels: I think the wording is a bit odd, maybe we can completely remove it, or rephrase it simpler. "Note, we are using normal sklearn without modifications here."

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+        # For reference: tqdm version was - would be nice to have an informative progress bar
+        # for i in tqdm(range(num_batches), leave=True, desc='Training in batches...'):
+
+        # Main loop: pass on batches to the model
+        for batch in range(num_batches):
+            # Get data for this batch
+            X = df[index_min:index_max][self.features].values
+            y = df[index_min:index_max][target].values
+            shuffle_index = np.arange(len(X))
```
maartenbreddels:
```suggestion
```

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+        '''Fit the IncrementalPredictor to the DataFrame.
+
+        :param df: A vaex DataFrame containing the features on which to train the model.
+        :param target: The name of the column containing the target variable.
+        '''
+        # Check whether the model is appropriate
+        assert hasattr(self.model, 'partial_fit'), 'The model must have a `.partial_fit` method.'
+
+        # Number of batches to send (or training rounds)
+        num_batches = np.int(np.ceil(len(df) / self.batch_size))
```
maartenbreddels:
```suggestion
        N_total = len(df)
        num_batches = (N_total + self.batch_size -1) // self.batch_size  # round up int
```

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+                    np.random.shuffle(shuffle_index)
+                    X = X[shuffle_index]
+                    y = y[shuffle_index]
+
+                # train the model
+                self.model.partial_fit(X, y)
+
+            # update the slicing indices to be ready for the next batch
+            index_min += self.batch_size
+            index_max += self.batch_size
```
maartenbreddels:
```suggestion
            index_max += self.batch_size
            index_max = min(N_total, index_max)  # clip upper value
```
Is this not triggered by the test?

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -95,3 +95,141 @@ def fit(self, df, target, **kwargs):
...
+
+    def predict(self, df):
+        '''Get an in-memory numpy array with the predictions of the SKLearnPredictor.self
+
+        :param df: A vaex DataFrame, containing the input features.
+        :returns: A in-memory numpy array containing the SKLearnPredictor predictions.
+        :rtype: numpy.array
+        '''
+        data = df[self.features].values
+        return self.model.predict(data)
```
maartenbreddels: why don't do here
```
return self.transform(df)[self.prediction_name].values
```
should be much more efficient right? should also update the docs for that

tests/ml/sklearn_test.py
```diff
@@ -160,3 +160,56 @@ def test_sklearn_estimator_classification_validation():
...
+def test_sklearn_incremental_predictor_serialize(tmpdir):
+    df = vaex.example()
+    df_train, df_test = df.ml.train_test_split(test_size=0.1, verbose=False)
+
+    features = df_train.column_names[:6]
+    target = 'FeH'
+
+    incremental = IncrementalPredictor(model=SGDRegressor(),
+                                       features=features,
+                                       batch_size=10_000,
```
maartenbreddels: Make batchsize an odd number and see if you can check from the skmodel how many samples it has seen.

packages/vaex-ml/vaex/ml/sklearn.py
JovanVeljanoski: I just copy-pasted it from the regular SKLearnPredictor, but we can change the wording

packages/vaex-ml/vaex/ml/sklearn.py
JovanVeljanoski: we could indeed!

tests/ml/sklearn_test.py
JovanVeljanoski: Unfortunately the sklearn models do not keep track on the number of samples they have seen.. however I can test this manually..

packages/vaex-ml/vaex/ml/sklearn.py
JovanVeljanoski: Yeah, but I don't find it overly long, and it might be helpful to new users.

maartenbreddels: But my point is that general info should not be put everywhere, but in one place to avoid maintenance burden, just an advice.

tests/ml/sklearn_test.py
maartenbreddels: You could use mock to wrap the method, e.g.:
https://stackoverflow.com/questions/18869141/using-python-mock-to-spy-on-calls-to-an-existing-object

JovanVeljanoski: Added an additional test uses `mock` to count the number of calls to the underlying `partial_fit` method to the model.

tests/ml/sklearn_test.py
```diff
@@ -213,3 +214,25 @@ def test_sklearn_incremental_predictor_serialize(tmpdir):
...
+    num_batches = (N_total + batch_size - 1) // batch_size
+
+    incremental = IncrementalPredictor(model=mock.Mock(),
+                                       features=features,
+                                       batch_size=batch_size,
+                                       num_epochs=num_epochs,
+                                       shuffle=False,
+                                       prediction_name='pred')
+    incremental.fit(df=df_train, target=target)
+    assert incremental.model.partial_fit.call_count == num_batches * num_epochs
```
maartenbreddels: This is a pretty trivial assertion. I am more worried about the # of rows being passed being equal to len(df_train).

Issue #499, opened 2019-12-11
fillna() got an unexpected keyword argument 'fill_nan'


laozhuang727:
Hi dear，

Follow the document to process the fillna, but got below error, would you please help to take a look?

x = np.array([3, 1, np.nan, 10, np.nan])
        df = vaex.from_arrays(x=x)
        df_filled = df.fillna(value=-1, column_names=['x'])
        print(df_filled)

ERROR:MainThread:vaex:error evaluating: x at rows 0-5
Traceback (most recent call last):
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/scopes.py", line 94, in evaluate
    result = self[expression]
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/scopes.py", line 141, in __getitem__
    raise KeyError("Unknown variables or column: %r" % (variable,))
KeyError: "Unknown variables or column: 'fillna(__original_x, -1, fill_nan=True, fill_masked=True)'"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/dataframe.py", line 3467, in table_part
    values[name] = df.evaluate(name)
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/dataframe.py", line 5038, in evaluate
    dtype = dtypes[expression] = self.dtype(expression, internal=False)
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/dataframe.py", line 2005, in dtype
    data = self.evaluate(expression, 0, 1, filtered=False, internal=True, parallel=False)
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/dataframe.py", line 5143, in evaluate
    value = scope.evaluate(expression)
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/scopes.py", line 94, in evaluate
    result = self[expression]
  File "/Users/ryan/opt/miniconda3/envs/tensorflow/lib/python3.6/site-packages/vaex/scopes.py", line 136, in __getitem__
    self.values[variable] = self.evaluate(expression)  # , out=self.buffers[variable])
[...]

PR #500, opened 2019-12-11
Fix fillna when applied to the whole dataframe

JovanVeljanoski:
This is a fix for #499 .

- [x] Write a test that exposes the problem
- [x] Fix the issue
- [ ] Review


maartenbreddels:
Could your test masked values? Since that is NA

JovanVeljanoski:
> Could your test masked values? Since that is NA

Good idea! I've updated the test.

PR #512, opened 2019-12-16
feat: autogenerate fast api

maartenbreddels:
This is a way we can autogenerate the API.

There was an issue with this before, that in order the autogenerate the API, we had to import all packages, which is slow.

Instead, we now generate a spec:
```
$ python -m vaex.ml.spec packages/vaex-ml/vaex/ml/spec.json
```

Which allows us to build the API at runtime.

The spec also will allow us to track spec changes in the models, with CI helping us detect changes.

TODO:
 * [x] make the copy argument a trait
 * [x] make sure the spec file is packaged
 * [x] include CI checks to make sure the spec matches our classes.

### DIFF
--- a/packages/vaex-ml/vaex/ml/__init__.py
+++ b/packages/vaex-ml/vaex/ml/__init__.py
@@ -1,207 +1,16 @@
 import warnings
+import json
+import os
 
 import vaex
-import vaex.dataframe
 from . import datasets
 from .pipeline import Pipeline
-
-from vaex.utils import InnerNamespace
 from vaex.utils import _ensure_strings_from_expressions
 
 
 class DataFrameAccessorML(object):
     def __init__(self, df):
         self.df = df
-
-    def pca(self, n_components=2, features=None, prefix='PCA_', progress=False):
-        '''Requires vaex.ml: Create :class:`vaex.ml.transformations.PCA` and fit it.
-
-        :param n_components: Number of components to retain. If None, all the components will be retained.
-        :param features: List of features to transform.
-        :param prefix: Prefix for the names of the transformed features.
-        :param progress: If True, display a progressbar of the PCA fitting process.
-        '''
-        features = features or self.df.get_column_names()
-        features = _ensure_strings_from_expressions(features)
-        pca = PCA(n_components=n_components, features=features, progress=progress)
-        pca.fit(self.df)
-        return pca
-
-
-    def label_encoder(self, features=None, prefix='label_encoded_', allow_unseen=False):
-        '''Requires vaex.ml: Create :class:`vaex.ml.transformations.LabelEncoder` and fit it.
-
-        :param features: List of features to encode.
-        :param prefix: Prefix for the names of the encoded features.
-        :param allow_unseen: If True, encode unseen value as -1, otherwise an error is raised.
-        '''
-        features = features or self.df.get_column_names()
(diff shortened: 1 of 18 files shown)


packages/vaex-ml/vaex/ml/__init__.py
```diff
@@ -224,6 +46,41 @@ def train_test_split(self, test_size=0.2, strings=True, virtual=True, verbose=Tr
...
+    # add DataFrameAccessorML.<snake_name> wrapper methods
+    with open(filename_spec) as f:
+        try:
+            spec = json.load(f)
+        except json.decoder.JSONDecodeError:
+            pass  # we are generating the file probably
+        else:
+            for class_spec in spec:
+                def closure(class_spec=class_spec):
+                    def wrapper(self, features=None, transform=False, **kwargs):
```
maartenbreddels, line 61:
```suggestion
                    def wrapper(self, features=None, transform=True, **kwargs):
```
should we do this by default? if yes, I guess we need to change the tests.

packages/vaex-ml/vaex/ml/__init__.py
```diff
@@ -75,6 +75,17 @@ def wrapper(self, features=None, transform=False, **kwargs):
                             return dft
                         else:
                             return object
+
+                    # Add docstrings
+                    doc = '\n'
+                    for trait in class_spec['traits']:
+                        doc += f':param {trait["name"]}: {trait["help"]} \n'
+                    doc+= ':param transform: If True, return a shallow copy of the transformed DataFrame, otherwise the return fitted transformer. \n'
```
maartenbreddels, line 83:
```suggestion
                    doc += ':param transform: If True, return a shallow copy of the transformed DataFrame, otherwise the return fitted transformer. \n'
```

packages/vaex-ml/vaex/ml/__init__.py
```diff
@@ -75,6 +75,17 @@ def wrapper(self, features=None, transform=False, **kwargs):
                             return dft
                         else:
                             return object
+
+                    # Add docstrings
```
maartenbreddels, line 79:
```suggestion
                    # Append trait help strings to the docstring
```

packages/vaex-ml/vaex/ml/__init__.py
```diff
@@ -75,6 +75,17 @@ def wrapper(self, features=None, transform=False, **kwargs):
...
+                    doc = '\n'
+                    for trait in class_spec['traits']:
+                        doc += f':param {trait["name"]}: {trait["help"]} \n'
+                    doc+= ':param transform: If True, return a shallow copy of the transformed DataFrame, otherwise the return fitted transformer. \n'
+                    try:
+                        wrapper.__doc__= class_spec['doc'] + doc
+                    except TypeError:
+                        wrapper.__doc__= class_spec['doc']
+
```
maartenbreddels, line 88: Which TypeError? Only thing I can think of is that class_spec['doc'] is None, but then it should assign doc right?

JovanVeljanoski: Ah you are right! Thanks!

maartenbreddels:
If you can rebase this, that would be great!

JovanVeljanoski:
Rebased!

maartenbreddels:
The OSX failure might have to do with the GHA cache, unrelated.

PR #517, opened 2019-12-20
Feat: Apache Arrow integration

maartenbreddels:
Replaces #467

The idea is that a vaex dataframe can take numpy and/or Arrow arrays, and tries to leave them as that type. When needed it will do on the fly conversions to convert to numpy or vice versa.

More in 2020

TODO
 - [x] cleanup
 - [x] tests: df fixture with arrow data


maartenbreddels:
Given that vaex dataframes have mixed arrow/numpy support, and people want to convert between the two, I propose we have the following behavior:

# Opening/reading arrow data

For v3, we by default (lazily) cast all arrow numerical types (double, int) to numpy types. This makes it easy to do math, and does not break opening arrow files compared to v2. Since arrow (v0.16) arrays do not support multiplication etc, it will render those columns useless, unless we add support for this.
```
df = vaex.open('test.arrow')  # all numerical columns will be available as numpy data
df = vaex.open('test.arrow', as_numpy=False)  # leave columns alone, it's the arrow data as it is on disk
```
# Conversion

## To numpy
### Complete dataframe
`df.as_numpy()` will convert all numerical columns to numpy columns, which means that string colums are not touched. We could add `df.as_numpy(strict=True)` that will convert everything it can, and throw an error otherwise.

### Single column
`df.<col>.as_numpy()` will lazily convert a column to numpy array always, even if of string type, it will throw an error if not possible.

`df.<col>.to_numpy()` is equivalent to `df.<col>.as_numpy().values`

## To arrow
### Complete dataframe
`df.as_arrow()` will convert whatever it can to an arrow array (all except the annoying dtype=object).

### Single column
`df.<col>.as_arrow()` will convert to arrow

`df.<col>.to_arrow()` is equivalent to `df.<col>.as_arrow().values`

# LTDR

For vaex v3 we do not expose numerical arrow arrays by default, unless explicitly opted in. String columns will stay in arrow format, and will/should replace our own arrow-like structure.

xdssio:
How would that effect any of the standard code writing? 
lazy columns and expression will work normally? what are the use cases?

maartenbreddels:
Nothing will change for v3 for the user, unless they explicitly say they don't want this auto conversion.

One thing that will change is checking types, for strings you have to check if the data type is `pyarrow.string` or `pyarrow.large_string`, which is slightly annoying.

xdssio:
are we going to wrap it so dtype will stay str? for both cases?

maartenbreddels:
I don't think so, it will be impossible to maintain to have 3 cases, np.dtype, arrow types and 'str' special case.

we might introduce helper functions `df.x.is_string()`

xhochy:
@maartenbreddels Have a look at https://pandas.pydata.org/docs/development/extending.html#compatibility-with-apache-arrow for supporting `pa.array(<vaex-column)`. Might be handy if this worked out of the box.

maartenbreddels:
> Might be handy if this worked out of the box.

👍

maartenbreddels:
The PR as it is now, will be non-breaking with vaex 3.x (as far as i consider it). We could merge this earlier then expected, so before 4.0.

To get it non-breaking, `vaex.open('file.arrow')` will convert all apache arrow arrays to NumPy by default (of course lazily), to opt-out use `vaex.open('file.arrow', as_numpy=False)`.

For vaex v4, we can change the default value from `as_numpy=True` to `as_numpy=False` (in https://github.com/vaexio/vaex/blob/ad77d9ac987075e41bbc1b3dd3fa4e05df01995f/packages/vaex-core/vaex/arrow/dataset.py#L13), meaning that arrow files will then by default expose the data as Apache Arrow arrays. 

To write code that is future compatible with v4, you have to explicitly set the value. E.g.
```
df = vaex.open('file.arrow')  # this might behave different with v4
df = vaex.open('file.arrow', as_numpy=True)   # always defined behaviour
df = vaex.open('file.arrow', as_numpy=False)  # always defined behaviour
```

Issue #528, opened 2019-12-25
Left (or Right) Outer joins on string fields yields an error on missing values (on left or right dataframes)


nageshdanturti:
Hi, 

I get a **non-fatal** error when I join two dataframes where the joining keys on both dataframes are string fields. The error happens when I perform an outer join (left or right) when I know one of the 2 dataframes contains fewer rows than the other. I am looking for a SQL-like outer join output where the resulting dataframe should show empty values for non-existing rows (instead it shows **error** in the output).  

Example:

df1 = 800K rows (4 fields all strings datatypes, joining (string) key = part_number)
df2= 1Million rows (5 fields all string datatypes, joining (string) key = prd_part_number)

syntax:

`
df1 = vaex.open(write_directory + 'ERPPartNumbers.parquet')
print("ERP Part Numbers = ", df1.count())

df2 = vaex.open(write_directory + 'ProductItems.parquet')
print("Product Items = ", df2.count())

df3=df1.join(df2, left_on='part_number', right_on='part_number', rprefix='cedi_', how='right', inplace=False)

print("Join between ERP and Product Items = ", df3.count())
df3
`

The error is as follows:

_ERROR:MainThread:vaex:error evaluating: part_number at rows 0-5
Traceback (most recent call last):
  File "/Applications/Anaconda/anaconda3/envs/eriks/lib/python3.7/site-packages/vaex/dataframe.py", line 3467, in table_part
    values[name] = df.evaluate(name)
  File "/Applications/Anaconda/anaconda3/envs/eriks/lib/python3.7/site-packages/vaex/dataframe.py", line 5038, in evaluate
    dtype = dtypes[expression] = self.dtype(expression, internal=False)
  File "/Applications/Anaconda/anaconda3/envs/eriks/lib/python3.7/site-packages/vaex/dataframe.py", line 2002, in dtype
    data = column[0:1]
  File "/Applications/Anaconda/anaconda3/envs/eriks/lib/python3.7/site-packages/vaex/column.py", line 124, in __getitem__
    ar = ar_unfiltered[indices]
  File "/Applications/Anaconda/anaconda3/envs/eriks/lib/python3.7/site-packages/vaex/column.py", line 398, in __getitem__
    return type(self).from_string_sequence(self.string_sequence.index(slice))
TypeError: index(): incompatible function arguments. The following argument types are supported:
[...]

PR #530, opened 2019-12-26
Fix: join issue when data is memory-mapped.

JovanVeljanoski:
This PR is tackling #528.

This issue occurs when the data is memory-mapped. It has to do with the string columns of the resulting joined dataframe having missing values, since the two dataframes being joined to each other do not correspond 1:1. 

- [x] Reproduce and implement a test
- [x] Fix


maartenbreddels:
It appeared to be a problem with slicing arrow string array, which your test exposed. I've removed it and reduced it to the fundamental problem/test. I did confirm your test passed before removing it.

maartenbreddels:
Fixed #530

Issue #529, opened 2019-12-25
vaex.concat issue


hanwsf:
df=vaex.open(DIR+'train_0.hdf5')
df1=vaex.open(DIR+'train_1.hdf5')
df3=vaex.concat([df,df1])
df3.head()

or：
from glob import glob
DIR ='./train/'
files = glob(DIR+'train_*.hdf5')
df = vaex.open(DIR+'train_*.hdf5')

These 2 cases can cause following issue from 2.1.0 to latest version (2.0.2 is trim issue):
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/.local/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/.local/lib/python3.6/site-packages/vaex/dataframe.py in _repr_mimebundle_(self, include, exclude, **kwargs)
   3615         # TODO: optimize, since we use the same data in both versions
   3616         # TODO: include latex version
-> 3617         return {'text/html':self._head_and_tail_table(format='html'), 'text/plain': self._head_and_tail_table(format='plain')}
   3618 
   3619     def _repr_html_(self):

~/.local/lib/python3.6/site-packages/vaex/dataframe.py in _head_and_tail_table(self, n, format)
   3408         N = _len(self)
   3409         if N <= n * 2:
-> 3410             return self._as_table(0, N, format=format)
   3411         else:
   3412             return self._as_table(0, n, N - n, N, format=format)

~/.local/lib/python3.6/site-packages/vaex/dataframe.py in _as_table(self, i1, i2, j1, j2, format)
   3540                 # parts += [""]
   3541             # return values_list
-> 3542         parts = table_part(i1, i2, parts)
   3543         if j1 is not None and j2 is not None:
   3544             values_list[0][1].append('...')

~/.local/lib/python3.6/site-packages/vaex/dataframe.py in table_part(k1, k2, parts)
   3518             # slicing will invoke .extract which will make the evaluation
   3519             # much quicker
-> 3520             df = self[k1:k2]
   3521             for i, name in enumerate(column_names):
   3522                 try:

~/.local/lib/python3.6/site-packages/vaex/dataframe.py in __getitem__(self, item)
   4552                 assert stop != -1
   4553                 stop = stop+1  # +1 to make it inclusive
-> 4554             df = self.trim()
[...]

PR #531, opened 2019-12-26
Fix head/tail for concatenated DataFrames with missing values

JovanVeljanoski:
This addresses #529 

The issue happens when trying to display portion of a concat dataframe with `.head` or `.tail` methods of a dataframe that contains missing values. 

- [x] Reproduce issue and create test
- [ ] Fix

### DIFF
--- a/tests/concat_test.py
+++ b/tests/concat_test.py
@@ -4,16 +4,16 @@ import numpy as np
 
 
 def test_concat():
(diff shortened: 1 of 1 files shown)


maartenbreddels:
my guess is that these are fixed by #590 , but still good to have the tests in.

PR #532, opened 2019-12-27
Cyclical features transformer

JovanVeljanoski:
Implementation of `CycleTransformer`. 

This is useful for encoding cyclical features (angles, time etc.) for machine learning tasks. The idea is to think of such circular features as the angle of a unit circle in polar coordinates, and then make projection on the x and y axes. This preserves the continuity (January is as close to December as it it is to February for example). For each encoded feature, two new ones are created. 

- [x] Implementation
- [x] Tests
- [x] Review

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -629,8 +629,8 @@ class CycleTransformer(Transformer):
         copy = df.copy()
         for feature in self.features:
             name_x = self.prefix_x + feature + self.suffix_x
-            copy[name_x] = (np.cos(2 * np.pi * copy[feature] / self.n))
+            copy[name_x] = np.cos(2 * np.pi * copy[feature] / self.n)
             name_y = self.prefix_y + feature + self.suffix_y
-            copy[name_y] = (np.sin(2 * np.pi * copy[feature] / self.n))
+            copy[name_y] = np.sin(2 * np.pi * copy[feature] / self.n)
 
         return copy
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -573,3 +573,64 @@ class RobustScaler(Transformer):
                 expr = expr / self.scale_[i]
             copy[name] = expr
         return copy
+
+
+@register
+@generate.register
+class CycleTransformer(Transformer):
+    '''A strategy for transforming cyclical features (e.g. angles, time).
+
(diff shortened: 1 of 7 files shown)


packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -573,3 +573,64 @@ def transform(self, df):
...
+    def transform(self, df):
+        '''
+        Transform a DataFrame with a CycleTransformer.
+
+        :param df: A vaex DataFrame.
+        '''
+        copy = df.copy()
+        for feature in self.features:
+            name_x = self.prefix_x + feature + self.suffix_x
+            copy[name_x] = (np.cos(2 * np.pi * copy[feature] / self.n))
```
maartenbreddels: Extra unnecessary parenthesis (we should use a linter!)

xdssio:
Isn't *n* should be a list, so it can handle multiple columns?

xdssio:
When it's implemented, it can open the door to a **DatetimeCyclicalTransformer** which can break the datetime feature into the year, month, day, hour,  minutes (depending on the params) and run the CyclicalTransformer on them. so one line kick ass on DateTime columns. maybe even read the str→DateTime too. future magic.

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Yup, linted :)

JovanVeljanoski:
@xdssio I agree that `n` can be auto-determined if not passed by the user. 

As for `n` being a list, I actually thought about it and started implementing it, but I then I thought it is a bit too involved to catch all corner cases and exceptions due to the combinations of params the user may input. 

I propose to keep it as it is now: One can still pass multiple input features, but they need to have the same "resolution" e.g. hour of check-in hour and check-out; day sent and day received etc. If one needs to encode different features that have different resolutions (days, months, angle etc..), one needs to instantiate a different transformer per resolution element. It will make maintenance easier. 

Any thoughts @maartenbreddels ? If you both think it is a good idea to be able to pass a list of features, and each having its own `n` so everything to be done in one go, I can work on it to make it possible.

maartenbreddels:
No, I don't think we should pass a list of features. Maybe we should do it in the future, and make everything consistent. I like n being automatically determined, but I'm not sure what the behaviour should be of the future refit. Should, if n was not set, it be set on the initial fit, and then reused? Or determined each case (that can lead to unexpected issues if the data is not complete).

So, difficult to answer, I think we merge this first, and we can improve later on, removing features will break, so this is a safer approach.

PR #536, opened 2019-12-30
Improve API to correlation and mutual_information DataFrame methods

JovanVeljanoski:
In this PR i propose certain changes to the API of `dataframe.correlation` and `dataframe.mutual_information`. The changes are related to how the data (columns) are passed on to the method. In summary:
- A single column or expression given for `x` and `y`: returns and array with a single number/item;
- A list of columns or expressions for `x`, `y=None`: returns a correlation matrix between the columns listed, which is an array of shape `(len(x), len(x))`;
- A list of columns or expressions given for `x`, and a single column or expression for `y`: returns an array of the correlation results between the elements of x and y, an array of shape `len(x)`;
- A list of columns or expressions for `x` and a list of columns or expressions for `y`: returns a 2d array of shape `(len(y), len(x)`, that contains the results of the correlation, each element of `x` to each element of `y`;
- No arguments are passed, i. e. `x=None`, `y=None`: returns a correlation matrix (2d array) of all numerical features in the DataFrame.

- [x] Implement tests to guide the API re-design
- [ ] Implementation of changes
- [ ] Review and test

### DIFF
--- a/tests/correlation_test.py
+++ b/tests/correlation_test.py
@@ -1,0 +1,35 @@
+import numpy as np
+
+import vaex
+
+
+def correlation_test():
+    df = vaex.example()
+
+    # A single column pair
+    desired = df.correlation('x', 'y')
+    expected = np.array(-0.066913)
+    np.testing.assert_array_almost_equal(desired, expected)
+
+    # A list of columns
+    desired = df.correlation(x=['x', 'y', 'z'])
+    expected = np.array([[ 1.00000000, -0.06691309, -0.02656313],
+                         [-0.06691309,  1.00000000,  0.03083857],
+                         [-0.02656313,  0.03083857,  1.00000000]])
+    np.testing.assert_array_almost_equal(desired, expected)
+
+    # A list of columns and a single target
+    desired = df.correlation(x=['x', 'y', 'z'], y='vx')
+    expected = np.array([-0.00779179,  0.01804911, -0.02175331])
+    np.testing.assert_array_almost_equal(desired, expected)
+
+    # A list of columns and a list of targets
+    desired = df.correlation(x=['x', 'y', 'z'], y=['vx', 'vy'])
+    expected = np.array(([-0.00779179,  0.01804911, -0.02175331],
+                         [0.00014019, -0.00411498, 0.02988355]))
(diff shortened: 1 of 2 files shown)


JovanVeljanoski:
For now, the proposal is that the result of `dataframe.correlation` and `dataframe.mutual_information` is a numpy array, but it can also be xarray, or a DataFrame in the future.  The tests can be easily modified to accommodate for this.

JovanVeljanoski:
For some reason the tests pass.. but they should not.

maartenbreddels:
A test function should between with `def test....`, i reverse it for the filename, to make it tab completion friendly in the shell.

maartenbreddels:
(i'll rename it btw)

tests/correlation_test.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+
+import vaex
+
+
+def correlation_test():
```
maartenbreddels, line 6:
```suggestion
def test_correlation():
```

tests/mutual_information_test.py
```diff
@@ -0,0 +1,36 @@
+import numpy as np
+
+import vaex
+
+
+def mutual_information_test():
```
maartenbreddels, line 6:
```suggestion
def test_mutual_information():
```

JovanVeljanoski:
This is all excellent! This is exactly the API i had in mind.
Few things that would be nice to have (and I believe we have the groundwork for them):
- enable progressbars :)
- perhaps `array_type` - like keyword to specify the output type (numpy, xarray, python lists). The xarray would especially be helpful for quick plotting as it "remembers" the names etc.. I guess keep numpy as default?
- enable the binby arguments. Don't know how often they are used, but this is a nice place where vaex really excels so I'd keep them if possible

Great great work!

maartenbreddels:
> * perhaps `array_type` - like keyword to specify the output type (numpy, xarray, python lists). The xarray would especially be helpful for quick plotting as it "remembers" the names etc.. I guess keep numpy as default?

is it ok to skip this for now?

maartenbreddels:
whow, old PR :)  Thanks!

PR #539, opened 2020-01-02
Fix IncrementalPredictor to work with classifiers

JovanVeljanoski:
A small fix in `vaex.ml.sklearn.IncrementalPredictor` so it can be used with `sklearn.linear_model.SGDClassifier`.  This requires that the classes are passed to the `.partial_fit` method of the scikit-learn model.

- [x] Add test
- [x] Apply fix (propagation of `**kwargs`)
- [ ] Review

### DIFF
--- a/packages/vaex-ml/vaex/ml/sklearn.py
+++ b/packages/vaex-ml/vaex/ml/sklearn.py
@@ -231,9 +231,4 @@ class IncrementalPredictor(state.HasState):
 
                 # train the model
                 self.model.partial_fit(X, y, **self.partial_fit_kwargs)
-
-            # update the slicing indices to be ready for the next batch
-            index_min += self.batch_size
-            index_max += self.batch_size
-            index_max = min(N_total, index_max)  # clip upper value
         progressbar(1.0)
--- a/packages/vaex-ml/vaex/ml/sklearn.py
+++ b/packages/vaex-ml/vaex/ml/sklearn.py
@@ -228,6 +228,11 @@ class IncrementalPredictor(state.HasState):
                     X = X[shuffle_index]
                     y = y[shuffle_index]
 
-                # Train the model
-                self.model.partial_fit(X, y)
+                # train the model
+                self.model.partial_fit(X, y, **kwargs)
+
+            # update the slicing indices to be ready for the next batch
+            index_min += self.batch_size
+            index_max += self.batch_size
+            index_max = min(N_total, index_max)  # clip upper value
         progressbar(1.0)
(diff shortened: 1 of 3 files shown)


packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -230,7 +230,7 @@ def fit(self, df, target, progress=None, **kwargs):
                     y = y[shuffle_index]
 
                 # train the model
-                self.model.partial_fit(X, y)
+                self.model.partial_fit(X, y, **kwargs)
```
maartenbreddels: Note that if we want to support refit, we cannot add extra kwargs to fit, all fit methods should have the same arguments. I think it is weird anyway to have extra arguments, for sklearn as well, why not have that as extra propery/traits?

tests/ml/sklearn_test.py
```diff
@@ -188,6 +189,34 @@ def test_sklearn_incremental_predictor():
...
+
+    incremental.fit(df=df_train, target=target, classes=[0, 1, 2])
+    df_train = incremental.transform(df_train)
+
+    # State transfer
+    state = df_train.state_get()
+    df_test.state_set(state)
+
+    assert df_test.column_count() == 6
+    assert df_test.pred.values.shape == (10050,)
```
maartenbreddels: I don't like magic values :)
```suggestion
    assert df_test.pred.values.shape == (len(df),)
```

packages/vaex-ml/vaex/ml/sklearn.py
maartenbreddels: I vote for having a 'partial_fit_kwargs' trait.
I think we should work towards a fixed fit method signature:
`def fit(self, df, progress=False):`
Any other arguments should be put as traits, so they become part of the state, and we can implement a refit, do online learning in the future etc.

packages/vaex-ml/vaex/ml/sklearn.py
```diff
@@ -228,6 +229,11 @@ def fit(self, df, target, progress=None, **kwargs):
...
 
-                # Train the model
-                self.model.partial_fit(X, y)
+                # train the model
+                self.model.partial_fit(X, y, **self.partial_fit_kwargs)
+
+            # update the slicing indices to be ready for the next batch
+            index_min += self.batch_size
+            index_max += self.batch_size
+            index_max = min(N_total, index_max)  # clip upper value
```
maartenbreddels: We don't need these anymore right?

JovanVeljanoski: Ah yeah, that is indeed correct. Fixed

PR #545, opened 2020-01-07
feat(core): progress update

maartenbreddels:
Refactored our own (old) progress bar a bit so the widget version can use much of the same logic.
The widget version also has cpu usage, which is the cpu usage between start end end:
progress-bar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # vaex-core 1.5.0-dev
    * Features
       * df.evalute_iterator for efficient parallel chunked evaluation [#515](https://github.com/vaexio/vaex/pull/515)
+      * Widget progress bar has time estimation [#545](https://github.com/vaexio/vaex/pull/545)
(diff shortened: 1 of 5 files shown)


maartenbreddels:
Also, the progress bar only updates at max 10x per second, instead of 100x.

JovanVeljanoski:
The widget progressbar is great! I like the addition of the CPU usage :).

However, when doing a `pca` transformation, the default (progressbar2) progressbar is used even when `progress=widget` is specified. Here is a copy-pastable example (intended for a Jupyter notebook):

```
import vaex
import vaex.ml

df = vaex.example()
pca = vaex.ml.PCA(features=['x', 'y', 'z'], n_components=3, progress='widget')
new_df = new_df = pca.fit_transform(df)
```

maartenbreddels:
I don't think progress should be a property of PCA, but a parameter of fit. In any case it is a separate issue.

PR #552, opened 2020-01-14
Improve the API of vaex.ml model transformers

JovanVeljanoski:
This PR aims to standardize the API of vaex.ml. The main change is that it moves the 'target' kwarg from the `.fit` method to the model constructor. The idea is that all params, properties etc.. of a model should be in the constructor. The `.fit` method should only accept the incoming data, and perhaps other options during the training that should not be stored in the state of the model (e.g. progressbars).

Apart from this, the PR brings some minor clean-ups and  consistency improvements in the docstrings and in the API. 

- [x] Move the `target` kwarg from the `.fit` fit method to the constructors for the sklearn, xgboost, lightgbm and catboost model wrappers. 
- [x] Update the fast(er) API for the models, making it more consistent;
- [x] Update the docstrings throughout, making them more consistent;
- [x] Update tests to account for this change;
- [x] Allow the target to be an expression, and the features to be a list of expressions for the sklearn, xgboost, lightgbm and catboost model wrappers (fast API only);
- [x] Update the tutorial in the documentation;
- [x] Update the examples in the documentation;
- [x] Update the changelog.


packages/vaex-ml/vaex/ml/catboost.py
```diff
@@ -97,17 +96,14 @@ def fit(self, df, evals=None, early_stopping_rounds=None, verbose_eval=None, plo
         :param bool plot: if True, display an interactive widget in the Jupyter
             notebook of how the train and validation sets score on each boosting iteration.
         '''
-        # Ensure strings
-        target = _ensure_strings_from_expressions(self.target)
-        features = _ensure_strings_from_expressions(self.features)
 
-        data = df[features].values
-        target_data = df[target].values
+        data = df[self.features].values
```
maartenbreddels: If/when we move to arrow, `.values` might return an arrow array. I think they automatically cast to a numpy array, and I think it's fine that we attach the problem when it appears, just a heads up.

JovanVeljanoski:
The CI fails because of setup_tools & python2.7 installation.
Is this safe to ignore @maartenbreddels, or should I find a way to make it pass?

maartenbreddels:
hmm, that's odd. I wanted to make vaex 2.x the last py2 release actually. So would be nice if we can test it. We could also test it locally and accept the failure, whatever takes the least amount of time.

JovanVeljanoski:
This is the reason why it fails..

image

I'm trying it by forcing an older version of setuptools

maartenbreddels:
This overlaps a bit with #512
But I'm more comfortable first merging this, doing a release, and see if we can use #512 instead for vaex3?

JovanVeljanoski:
I guess this is the template for #512. I agree, we can hold of on this until the next major release of vaex. 

However, please do not make a release right away, there is another PR coming right after this is merged. 

Happy for this to be merged tho.

maartenbreddels:
Are the checkbox that are unchecked done? I think so right?

JovanVeljanoski:
Ah yes they are, my bad

maartenbreddels:
And could you squash all but the last commit?

JovanVeljanoski:
Like this?  I hope I didn't mess up big time..

maartenbreddels:
Thanks!

PR #559, opened 2020-01-22
Feat: support python38

maartenbreddels:


maartenbreddels:
CI issues :-(

@JovanVeljanoski: do you want to take a look, maybe try installing it locally?

maartenbreddels:
The deadlock of #725 seems to be happening on windows a lot

PR #572, opened 2020-01-28
Update of the vaex tutorial in docs

JovanVeljanoski:
This PR presents an update of the main tutorial on the docs page. 

It features various small typo corrections, wording improvement, addition of a Groupby sections, and little changes that were improved/changed since the last time the tutorial was updated. 

- [ ] Review


JovanVeljanoski:
The failing test is unrelated to the PR (numpy issue most likely)

maartenbreddels:
Thanks! Found a few issues, but I could not annotate in github (the diff was too large):

> 
If you do need to use the underlying Numpy array (e.g. to pass it to another library) is to use evaluate.

Remove 'is to' ?

> This is similar as when using Pandas, expect that Vaex does not copy the data.

except

> Using the .get on the promise will also return the resut

Missing l

>Note that when multiple aggregations as passed to a

as -> are

PR #573, opened 2020-01-29
Refactor variables

maartenbreddels:
Various fixes (and breaking changes) coming out of #415


maartenbreddels:
Failure is unrelated, we should extend when to skip that test, you want to do that @JovanVeljanoski ?

https://ci.appveyor.com/project/maartenbreddels/vaex-4gh4b/builds/30429064

PR #581, opened 2020-02-07
fix!: default limits is 'minmax' so we do not miss outliers

maartenbreddels:


maartenbreddels:
This means that if you plot something, it will show everything (up to the edges) by default:
image

PR #584, opened 2020-02-10
feat: support jitting using OSX Metal

maartenbreddels:
Proof of concept for jitting using Metal. Unfortunately, this does not work that great. We can't schedule multiple compute kernels multithreaded, it seems to share some memory/state, because the results are garbled up without the lock in `metal.py`. Disabling ths lock gives speedups of `>2x`, on a 2018 MacBook 15" with Radeon Pro 560 4GB. With the lock the performance is the same as on my CPU.

Requires https://github.com/wtnb75/runmetal/
`pip install git+https://github.com/wtnb75/runmetal/` should do it.

A way out could be to do an MetalExecutor, meaning we take control of the whole execution part, single threaded, and execute all the code on the GPU, but that requires a lot more work.

Putting this out there in case someone finds this interesting, or has ideas how to avoid that lock.

Example usage:
```python
import vaex
import numpy as np
# From http://pythonhosted.org/pythran/MANUAL.html
def arc_distance(theta_1, phi_1, theta_2, phi_2):
    """
    Calculates the pairwise arc distance
    between all points in vector a and b.
    """
    temp = (np.sin((theta_2-theta_1)/2)**2
           + np.cos(theta_1)*np.cos(theta_2) * np.sin((phi_2-phi_1)/2)**2)
    distance_matrix = 2 * np.arctan2(np.sqrt(temp), np.sqrt(1-temp))
    return distance_matrix

df = vaex.open('s3://vaex/taxi/yellow_taxi_2009_2015_f32.hdf5?anon=true')[:200_000_000]
expr = arc_distance(df.pickup_longitude, df.pickup_latitude, df.dropoff_longitude, df.dropoff_latitude)
df['arc_distance'] = expr
#assert df.arc_distance.expression == expr.expression
df['arc_distance_jit'] = df['arc_distance'].jit_metal()

# cell2
%%time
df.arc_distance.sum()

# cell 3
%%time
df.arc_distance_jit.sum()
```

### DIFF
--- a/packages/vaex-core/vaex/array_types.py
+++ b/packages/vaex-core/vaex/array_types.py
@@ -6,7 +6,8 @@ import vaex.utils
 supported_arrow_array_types = (pa.Array, pa.ChunkedArray)
 supported_array_types = (np.ndarray, ) + supported_arrow_array_types
 string_types = [pa.string(), pa.large_string()]
-_type_names = ["float64", "float32", "int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64"]
+_type_names_int = ["int8", "int16", "int32", "int64", "uint8", "uint16", "uint32", "uint64"]
+_type_names = ["float64", "float32"] + _type_names_int
 map_arrow_to_numpy = {getattr(pa, name)(): np.dtype(name) for name in _type_names}
 map_arrow_to_numpy[pa.bool_()] = np.dtype("?")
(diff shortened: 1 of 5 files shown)


maartenbreddels:
Note that don't use runmetal anymore, using the bare Metal package (https://pypi.org/project/pyobjc-framework-Metal/) and with proper use can use it multithreaded. Performance seems good, maybe we should report back here on that.

PR #589, opened 2020-02-11
feat: let to_items/pandas_df/arrow_table return chunked iterators

maartenbreddels:
This is useful for outputting to csv or arrow very large dataframes
with virtual columns that cannot be kept into memory.

Example usage for csv:
```python
import vaex
df = vaex.open('s3://vaex/taxi/yellow_taxi_2009_2015_f32.hdf5?anon=true')[:1_000_000]
df = df[['passenger_count', 'fare_amount', 'pickup_datetime']]
for i1, i2, pdf in df.to_pandas_df(chunk_size=100_000):
    fn = f'taxi_{i1}-{i2}.csv'
    print(fn)
    pdf.to_csv(fn)
```

Or outputting to parquet
```python
schema = df[0:1].to_arrow_table().schema
with pq.ParquetWriter('example.parquet', schema) as writer:
    for i1, i2, table in df.to_arrow_table(chunk_size=100_000):
        print(i1, i2)
        writer.write_table(table)
```

Note that we probably want a nicer way to get the arrow schema. Also, we yield the row indices, should we also yield the chunk_index? e.g.:
```python
for chunk_index, row_begin, row_end, pdf in df.to_pandas_df(chunk_size=100_000):
```

We may want to put this into the docs @JovanVeljanoski , maybe open an item for this, basically we need a single page for IO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
       * Give proper NameError when using non-existing column names [#299](https://github.com/vaexio/vaex/pull/299)
    * Features
       * New lazy numpy wrappers: np.digitize and np.searchsorted [#573](https://github.com/vaexio/vaex/pull/573)
+      * df.to_arrow_table/to_pandas_df/to_items now take a chunk_size argument for chunked iterators [#589](https://github.com/vaexio/vaex/pull/589)
 
 # vaex-server 0.3.0-dev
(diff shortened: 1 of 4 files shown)


maartenbreddels:
We might be able to completely refactor/simplify the exporting code based on this. At least we should have the arrow and parquet exporting code use it and write chunks.

maartenbreddels:
> I tested it out locally and the performance I get when using the chunking parameter is often greater than x5!

what do you do, and compared to what?



> Seems a shame if by default this superpower is not used; although I see why this would be as it makes assumptions about the data, the available RAM etc..

I think we should merge this as is, so it does not break the API, but we first attack export and make it use these methods.

PR #590, opened 2020-02-12
Fixes/features for concat / concatenate 

maartenbreddels:
We not run all test also on the concatenated dataframe, and added some extra test.

Also, we now support filtered dataframe to be concatenated.


maartenbreddels:
Unrelated failure from travis, too many sockets open. Might have to do with the tests now running more servers and clients.

Issue #591, opened 2020-02-13
state_transfer fail after fillna of a virtual column


xdssio:
reproduce: 
```
from vaex.ml.datasets import load_titanic

train, test = load_titanic().split_random(0.7, random_state=42)
train['age_standard'] = (train.age - train.age.mean()) / train.age.std()  # by expression
train['age_standard'] = train['age_standard'].fillna(0)
test.state_set(train.state_get())
```

```
Traceback (most recent call last):
  File "/Users/yonatanalexander/opt/anaconda3/envs/xvaex/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3319, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-22-272bbd75908d>", line 1, in <module>
    test.state_set(df.state_get())
  File "/Users/yonatanalexander/opt/anaconda3/envs/xvaex/lib/python3.7/site-packages/vaex/dataframe.py", line 2297, in state_set
    self._rename(old, new)
  File "/Users/yonatanalexander/opt/anaconda3/envs/xvaex/lib/python3.7/site-packages/vaex/dataframe.py", line 3277, in _rename
    self.column_names.remove(old)
ValueError: list.remove(x): x not in list
```

xdssio:
possible solution:
in `veax.dataframe.DataFrame`- the function `_rename`, making a small test:
```
    def _rename(self, old, new, rename_meta_data=False):
        if old in self._dtypes_override:
            self._dtypes_override[new] = self._dtypes_override.pop(old)
        if old in self.columns:
            self.columns[new] = self.columns.pop(old)
        if rename_meta_data:
            for d in [self.ucds, self.units, self.descriptions]:
                if old in d:
                    d[new] = d[old]
                    del d[old]
        if old in self.virtual_columns:
            self.virtual_columns[new] = self.virtual_columns.pop(old)
            self._virtual_expressions[new] = self._virtual_expressions.pop(old)
        self._renamed_columns.append((old, new))
        if old in self.column_names:  ############# THIS IS THE CHANGE ###########
            self.column_names.remove(old)
        self.column_names.append(new)
        for key, value in self.selection_histories.items():
            self.selection_histories[key] = list([k if k is None else k._rename(self, old, new) for k in value])
        if hasattr(self, name):
            try:
                if isinstance(getattr(self, name), Expression):
                    delattr(self, name)
            except:
                pass
        self._save_assign_expression(new)
[...]

PR #592, opened 2020-02-14
Fix state transfer after rewriting virtual columns

JovanVeljanoski:
This PR handles the issue report in #591 

- [x] Write test that exposes the issue
- [x] Implement solution
- [x] review and test


maartenbreddels:
Note that this has nothing to do with fillna, this also exposes the bug:
```
    df_train['new_x'] = df_train.x + 1
    df_train['new_x'] = df_train.x + 1
```

PR #593, opened 2020-02-14
Make df.column_count to not count hidden columns

JovanVeljanoski:
This PR makes `df.column_count()` to not count hidden columns in the DataFrame

- [x] Add test
- [x] Make changes
- [ ] Review

### DIFF
--- a/tests/column_test.py
+++ b/tests/column_test.py
@@ -73,3 +73,14 @@ def test_dtype_object_with_arrays():
     df = vaex.from_arrays(z=z)
     assert df.z.tolist()[0].tolist() == x.tolist()
     assert df.z.tolist()[1].tolist() == y.tolist()
+
+
+def test_column_count():
+    x = np.array([1, 2, np.nan])
+    df = vaex.from_arrays(x=x)
+
+    df['new_x'] = df.x + 1
+    df['new_x'] = df['new_x'].fillna(value=0)
+
+    assert df.column_count() == 2
+    assert df.new_x.tolist() == [2, 3, 0]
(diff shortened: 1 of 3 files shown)


tests/column_test.py
```diff
@@ -64,3 +64,13 @@ def test_plain_strings():
     assert len(df[1:3]) == 2
     assert df[1:3].x.tolist() == x[1:3].tolist()
 
+
+def test_column_count():
+    x = np.array([1, 2, np.nan])
+    df = vaex.from_arrays(x=x)
+
+    df['new_x'] = df.x + 1
```
maartenbreddels, line 72: This was it's clear how many columns we start with
```suggestion
    df['new_x'] = df.x + 1
    assert df.column_count() == 2
```

tests/column_test.py
```diff
@@ -64,3 +64,13 @@ def test_plain_strings():
...
+    df['new_x'] = df['new_x'].fillna(value=0)
```
maartenbreddels, line 73:
```suggestion
    # 'overwriting' a column will rename a column to a hidden column
    df['new_x'] = df['new_x'].fillna(value=0)
```

tests/column_test.py
```diff
@@ -64,3 +64,13 @@ def test_plain_strings():
...
+
+    assert df.column_count() == 2
```
maartenbreddels, line 75:
```suggestion
    assert df.column_count() == 2  # which we do not account in column_count
    assert df.column_count(hidden=True) == 2  # unless we ask to
```

maartenbreddels:
Hmm, could you look for usage of this method internally in vaex? It seems to crash on this, so all usage should add this argument.

Not that this is a breaking change, so https://www.conventionalcommits.org/en/v1.0.0-beta.4/#commit-message-with-optional-to-draw-attention-to-breaking-change please make the commit msg reflect this.

Just make it 1 commit and squash/rebase.

PR #599, opened 2020-02-17
Bug in .isin() for string columns that have None

JovanVeljanoski:
This test uncovers a bug in `.isin` method: if a column is string, and has `None` values inside, `.isin` will return `False` for all samples. 

- [x] Update test to cover the issue
- [ ] Make test pass

### DIFF
--- a/tests/isin_test.py
+++ b/tests/isin_test.py
@@ -5,19 +5,22 @@ import vaex
 def test_isin():
     x = np.array([1.01, 2.02, 3.03])
     y = np.array([1, 3, 5])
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I think actually we should have:
```
assert df.sm.isin(['cat', 'dog']).tolist() == [True, True, None]
```
Since comparing missing values should lead to missing values, e.g.:

image

maartenbreddels:
I rebased this, i think this is now fixed, so this means this PR only adds a test

Issue #598, opened 2020-02-17
dropna() default on all columns


xdssio:
Currently:

```
df = vaex.from_arrays(i=range(12), c=['a'] * 3 + ['b'] * 3 + ['c'] * 3 + [None] * 3)
assert len(df.dropna()) == len(df.dropna(column_names=['c'])) # fails 
```
I think you would expect it to remove all na. 
So after you run dropna, you have no na's in your data.

maartenbreddels:
Good catch, the issue is here:
https://github.com/vaexio/vaex/blob/54d5d30e28360cbc0d35de9cba0c09c3c6ab016e/packages/vaex-core/vaex/dataframe.py#L4272
Should have been `|` and not `&`!

The tests failed to pick this up because we didn't test with 2 columns where 1 of them had missing, while the other did not. Want to try to make a PR?

maartenbreddels:
Fixed in #600, thanks Y & J!

PR #600, opened 2020-02-17
Fix: dropna issue when no columns are specified. 

JovanVeljanoski:
This test exposes the issue reported in #598

- [x] Add a test exposing the issue
- [ ] Make the test pass


maartenbreddels:
Can you add this to the changelog, and squash and reformat de commit as https://www.conventionalcommits.org/en/v1.0.0-beta.2/, that is e.g.
`fix(core): dropna/nan/missing only dropped rows when all column values were missing`

PR #609, opened 2020-02-22
Fix download link for the vaex.example() dataset

### LINKED ISSUE #169: vaex.example() is broken
`vaex.example()` is doesn't work because http://vaex.astro.rug.nl/data/helmi-dezeeuw-2000-10p.hdf5 returns a 403 (at the time of writing). This is with `vaex` 1.0.0b7 and `vaex-core` 0.6.1 from conda-forge.

JovanVeljanoski:
A simple fix for the download link of the `vaex.example()` dataset.

The vaex example dataset is (at least for now) [hosted on github](https://github.com/vaexio/vaex-datasets). This makes it easily accessible for download. 

This closes #169


maartenbreddels:
Great! Do you have the full dataset? If not I can upload that one.

maartenbreddels:
Btw, we could also make vaex lazily open and cache the file, similar to how we do it with s3.

JovanVeljanoski:
This is the 10% version. `vaex.example` always used the 10% version as far as I am aware. 

I think this is small enough that the lazy opening is not needed. I guess this is used mainly for learning/experimentation purposes, so it is nice if the users have it all at once, at it is small enough anyway.

maartenbreddels:
Hi Jovan,

I fixed both datasets,can you check these work and update the PR?
```
vaex.datasets.helmi_de_zeeuw.url = 'https://github.com/vaexio/vaex-datasets/releases/download/v1.0/helmi-dezeeuw-2000-FeH-v2.hdf5'
vaex.datasets.helmi_de_zeeuw_10percent.url = 'https://github.com/vaexio/vaex-datasets/releases/download/v1.0/helmi-dezeeuw-2000-FeH-v2-10percent.hdf5'
```

maartenbreddels:
als, vaex.example()'s code should be changed btw, it checks the old filename, I think it can simply always call fetch

JovanVeljanoski:
The links are updated for the released versions of the example datasets from the vaex-datasets repo. 

Note that the /data folder in vaex has also been removed.

Indeed, there is some redundancy with the code of `vaex.example()` I will take a look now

JovanVeljanoski:
@maartenbreddels I remove some code that I thought it is no longer needed. The tests pass (locally, should pass on the CI as well), but please give it a quick eye in case you recall if that function is used for other stuff...

packages/vaex-core/vaex/__init__.py
```diff
@@ -551,16 +551,13 @@ def connect(url, **kwargs):
     from vaex.server import connect
     return connect(url, **kwargs)
 
+
 def example(download=True):
```
maartenbreddels:
```suggestion
def example():
```

tests/virtual_columns_test.py
```diff
@@ -8,13 +8,13 @@ def test_add_virtual_columns_polar_velocities_to_cartesian():
 
     # With azimuth = None
     ds.add_virtual_columns_polar_velocities_to_cartesian(vx_out='vx_', vy_out='vy_')
-    np.testing.assert_almost_equal(ds.evaluate('vx'), ds.evaluate('vx_'), err_msg='error with converting polar to Cartesian velocities')
-    np.testing.assert_almost_equal(ds.evaluate('vy'), ds.evaluate('vy_'), err_msg='error with converting polar to Cartesian velocities')
+    np.testing.assert_almost_equal(ds.evaluate('vx'), ds.evaluate('vx_'), err_msg='error with converting polar to Cartesian velocities', decimal=3)
```
maartenbreddels: Hmm, this is because of float32.. ok

packages/vaex-core/vaex/__init__.py
JovanVeljanoski: Yup, fixed in the latest commit

Issue #585, opened 2020-02-11
Opening chunked array files fails


JovanVeljanoski:
An issue to track the support for chunked arrays (primarily for arrow files). 

@xdssio if you can provide some sample data for testing, that would be great.

xdssio:
[File example](s3://xdss-tmp/examples/simple_test_with_batch.arrow)

byaminov:
Here's another reproducible case. Try to read a small CSV with `pyarrow` and you get a `ChunkedArray` inside. The following script could be copied as a unit test:

```
import vaex
from pyarrow import csv

csv_path = 'vaex/tests/data/small3.csv'
table = csv.read_csv(csv_path)
df = vaex.from_arrow_table(table)
# this works
assert len(df) == 3
# this fails: there's only [5, 7]
assert df.columns['x'] == [5, 7, 9]
# this fails with ValueError: could not broadcast input array from shape (2) into shape (3)
assert (df.x.values == [5, 7, 9]).all()
```

As this happens not only for big tables, I'd re-categorize this issue to a "bug".

maartenbreddels:
I think this should be solved once #517 is merged since we then fully support arrow.

JovanVeljanoski:
This is now fixed. Thanks everyone.
Please reopen if you encounter new issues.

PR #611, opened 2020-02-24
Feature: read big CSV files

byaminov:
This changes `vaex.from_csv` to be able to read a CSV file that is too big to fit into memory. When called with `vaex.from_csv("big.csv", convert=True)`, Vaex will read CSV file in chunks of 5 million rows, save each chunk to "big.csv_chunk3.hdf5", and later merge all the intermediate files into one "big.csv.hdf5".

The reading speed with Pandas reading and Vaex conversion is like following on my Mac, when reading NYC Taxi CSV files from first half of 2019:
* around 500 000 rows per second with `convert=False`
* around 190 000 rows per second with `convert=True`, when several chunks had to be created (7 in my case). So 45 million rows, 4Gb CSV file is converted and read in 4 minutes. (This is measured with a fix in `dataframe.py` which I'll create as a different PR.)

Still to do in this PR:
- [x] @maartenbreddels wanted to discuss whether current place is a good one for this code. On one hand it would be logical to implement CSV as another _opener_, similar to `AsciiTable` opener. On the other hand reading a big CSV requires conversion, and `convert=True` is currently done solely by `vaex/__init__.py`. I wonder what's your opinion.
- [x] Try to get rid of code duplication for converting chunks and merging them together in `open(..)` and `from_csv(..)`, if we decide to keep `from_csv` logic in that module. (And from `read_csv_and_convert`.)
- [x] Try if `pyarrow.read_csv` would work faster when chunking through `skip_rows`.


maartenbreddels:
Hi Bulat,

looing good!
But, this got me thinking :)
Maybe we should make this behave more similar to: https://github.com/vaexio/vaex/pull/589

Meaning, that if you pass the chuck_size argument, you get an a generator which produces dataframes. But when the convert argument is given, it will do the whole conversion, sth like this:
```python
def from_csv(fn, chunk_size=None, **kwargs):
    def iterator():
        # a function cannot return OR be a generator, we use an inner function for that
        csv_reader = pd.read_csv(filename_or_buffer, chunksize=chunk_size, **kwargs)
        for pandas_df in csv_reader:
            yield from_pandas(pandas_df)
    if chunk_size is not None:
        if convert is not None:
            for i, df in from_csv(df, chunk_size=None, **kwargs)
                pass # write out chunk i
            # combine all chunks
        else:
            return iterator()
    else:
        return from_pandas(pd.read_csv(filename_or_buffer, **kwargs))
```

So only by passing chunk_size *without* passing convert will give a dataframe iterator.

This makes is easier to do whatever you want, e.g. filtering before writing it out:
```python
for i, df in enumerate(vaex.from_csv('big.csv', chunk_size=1_000_000)):
    df = df[df.passenger_count < 6]
    df.export(f'taxi_{i:02}.hdf5')
```

@JovanVeljanoski, what do you think?

cheers,

Maarten

byaminov:
@maartenbreddels ok I think it is a good idea. What do you think about the question I asked in the PR description: should the CSV reader (with the conversion capability) be an `opener` or a function where it is now?

JovanVeljanoski:
I am in agreement.

maartenbreddels:
I think it should be a function as it is now, that opener stuff is legacy stuff, and I hope to refactor that part.

benchmarks/from_csv.py
```diff
@@ -0,0 +1,124 @@
+import glob
+import os
+from datetime import datetime
+
+import vaex
+
+
+def test_open_two_big_csv_convert():
```
byaminov: These functions will be converted to use ASV in #620

byaminov:
As for the left TODOs:
> Try to get rid of code duplication for converting chunks and merging them together in `open(..)` and `from_csv(..)`

I'd like to do it in a different PR, together with trying to convert multiple CSVs to HDF5 in parallel. Seems like a lot of work to me now, therefore suggest to not include here. I can create a new issue for that if @maartenbreddels agrees.

> Try if `pyarrow.read_csv` would work faster when chunking through skip_rows.

Could not make it work because of #585, so this cannot be tried now.

benchmarks/from_csv.py
maartenbreddels: No, i'm ok having this in now

maartenbreddels:
> I'd like to do it in a different PR, together with trying to convert multiple CSVs to HDF5 in parallel. Seems like a lot of work to me now, therefore suggest to not include here. I can create a new issue for that if @maartenbreddels agrees.

Yes, I'm ok with that, I think it is best to make small steps first, as long as we don't break APIs, we can implement improvements in later minor releases.

maartenbreddels:
So, only a refactor using my code example as guideline till the next review?

byaminov:
> So, only a refactor using my code example as guideline till the next review?

That's already done, so you can read as an iterator: see [the test](https://github.com/vaexio/vaex/pull/611/files#diff-15bc6e079f7c6083f7e40b63badc2933R23-R29). So if you can review again, that would be great! I will check the failing builds in the meanwhile.

packages/vaex-core/vaex/__init__.py
```diff
@@ -474,13 +475,99 @@ def from_json(path_or_buffer, orient=None, precise_float=False, lines=False, cop
...
+        def iterator():
+            chunk_iterator = pd.read_csv(filename_or_buffer, chunksize=chunk_size, **kwargs)
+            for chunk_df in chunk_iterator:
+                yield from_pandas(chunk_df, copy_index=copy_index)
+        return iterator()
+
+
+def _from_csv_convert_and_read(filename_or_buffer, copy_index, maybe_convert_path, chunk_size, **kwargs):
+    # figure out the CSV file path
+    if isinstance(filename_or_buffer, six.string_types):
```
maartenbreddels: since we now use Python3 only, we don't have to do this anymore. I don't want to change the current code aggressively to pollute the git history, that's why you still see Python2 compatible code 
```suggestion
    if isinstance(filename_or_buffer, str):
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -474,13 +475,99 @@ def from_json(path_or_buffer, orient=None, precise_float=False, lines=False, cop
...
+        csv_path = filename_or_buffer
+    elif isinstance(maybe_convert_path, six.string_types):
```
maartenbreddels: same.
```suggestion
    elif isinstance(maybe_convert_path, str):
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -474,13 +475,99 @@ def from_json(path_or_buffer, orient=None, precise_float=False, lines=False, cop
...
+    combined_hdf5 = _convert_name(csv_path)
+    if os.path.exists(combined_hdf5):
+        return vaex.file.open(combined_hdf5)
+
+    # convert CSV chunks to separate HDF5 files
     import pandas as pd
-    return from_pandas(pd.read_csv(filename_or_buffer, **kwargs), copy_index=copy_index)
+    converted_paths = []
+    csv_reader = pd.read_csv(filename_or_buffer, chunksize=chunk_size, **kwargs)
+    for i, df in enumerate(csv_reader):
```
maartenbreddels:
```suggestion
    for i, df_pandas in enumerate(csv_reader):
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -474,13 +475,99 @@ def from_json(path_or_buffer, orient=None, precise_float=False, lines=False, cop
...
+        ds = from_pandas(df, copy_index=copy_index)
```
maartenbreddels: before vaex had a dataframe API, it's classname was a dataset, and we'd use `ds` a lot in the codebase. We're moving that to `df` everywhere when we refactor/change. I think it's a good chance to do this here as well, and give pandas dataframes a `_pandas` suffix.
```suggestion
        df = from_pandas(df, copy_index=copy_index)
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -474,13 +475,99 @@ def from_json(path_or_buffer, orient=None, precise_float=False, lines=False, cop
...
+        filename_hdf5 = _convert_name(csv_path, suffix='_chunk%d' % i)
+        ds.export_hdf5(filename_hdf5, shuffle=False)
```
maartenbreddels:
```suggestion
        df.export_hdf5(filename_hdf5, shuffle=False)
```

tests/from_csv_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+import pytest
+import vaex
+
+path = os.path.dirname(__file__)
+csv_path = os.path.join(path, 'data', 'small3.csv')
+
+
+def test_from_csv():
+    _cleanup_generated_files()
```
maartenbreddels: curious why you chose this code pattern, you prefer the files to be around after the test for manual inspection? Otherwise you could do a try/finally? I'm ok with both.

tests/from_csv_test.py
```diff
@@ -0,0 +1,81 @@
...
+        os.rename(csv_path + '_', csv_path)
+
+
+def _assert_csv_content(csv_df, with_index=False):
+    assert len(csv_df) == 3
+    if with_index:
+        assert csv_df.get_column_names() == ['x', 'y', 'index']
+    else:
+        assert csv_df.get_column_names() == ['x', 'y']
+    assert (csv_df.x.values == [5, 7, 9]).all()
```
maartenbreddels: I generally prefer:
```
df.x.tolist() == [5, 7, 9]
```
x might in the future be an arrow array, and I am not sure arrow will like comparisons to list.

tests/from_csv_test.py
```diff
@@ -0,0 +1,81 @@
...
+def test_from_csv_converting_in_chunks():
+    _cleanup_generated_files()
+
+    # can read several chunks with converting, intermediate files are deleted
+    df = vaex.from_csv(csv_path, copy_index=False, chunk_size=1, convert=True)
+    _assert_csv_content(df)
+    for filename in ['small3.csv_chunk0.hdf5', 'small3.csv_chunk1.hdf5', 'small3.csv_chunk2.hdf5']:
+        assert not os.path.exists(os.path.join(path, 'data', filename))
+    assert os.path.exists(os.path.join(path, 'data', 'small3.csv.hdf5'))
+    _cleanup_generated_files(df)
```
byaminov: @maartenbreddels I continue having troubles trying to make this test work on Windows. First I was not closing DataFrames properly, but now I'm getting a segmentation fault from deeper DataFrame threading logic: see [this run](https://github.com/vaexio/vaex/pull/611/checks?check_run_id=478452044#). Do you have any ideas how to fix it? I've seen that in the `open_test.py` the Windows tests [are just skipped](https://github.com/vaexio/vaex/blob/master/tests/open_test.py#L12), should I do the same here?

packages/vaex-core/vaex/__init__.py
byaminov: Oh nice, looks better! Will use `str` then.

packages/vaex-core/vaex/__init__.py
byaminov: Ok, will do.

tests/from_csv_test.py
maartenbreddels: Difficult to say if we should skip it, basically I didn't care about windows before, but I care a bit more now, as we do have serious users. Do you have access to a windows virtual machine maybe?
I am also not sure if the segfault is related, I may have seen that before (which is scary), maybe some external library got updated.

tests/from_csv_test.py
byaminov: Nice, I like it more too, will apply.

tests/from_csv_test.py
byaminov: At some point of time I spent a lot of time trying to see why all of a sudden the tests stopped passing locally, even reverting the code to `master` did not help. Then it appeared that on one of the previous runs had failed and left some generated files behind. Which made the test start failing even with the correct code.

So in general I don't mind cleaning up in the beginning of the test, to avoid these kind of unnecessary puzzles. But I understand why this might look annoying. Let me just remove it in this case.

byaminov:
@maartenbreddels I've implemented your review comments, please re-review when you have time.

tests/from_csv_test.py
byaminov: > Difficult to say if we should skip it, basically I didn't care about windows before, but I care a bit more now, as we do have serious users. Do you have access to a windows virtual machine maybe?
> I am also not sure if the segfault is related, I may have seen that before (which is scary), maybe some external library got updated.

it happened again, see [the build in appveyor](https://ci.appveyor.com/project/maartenbreddels/vaex-4gh4b/builds/31191907/job/648o04taq1w5hqre): with segfault in generic stacktrace, but seems to happen during/after the new `tests\from_csv_test.py`. The windows builds in GitHub Actions have succeeded. Do you have an idea how to proceed? Please advice, as I'm out of ideas :)

tests/from_csv_test.py
maartenbreddels: No, I kind of like it, because it allows you to inspect issues after the test run. Cleanups can be annoying.

byaminov:
@maartenbreddels what's up with this PR, do you want me to do any other changes? And also did the Windows problem hunting give any results?

maartenbreddels:
Hi Bulat, I still need to look into the windows issue. I'd like to do that before I lose track of that.

maartenbreddels:
I discussed this with @JovanVeljanoski, and I think we should disable the test on windows, because they were disabled before. I don't want to not merge this because of an unrelated issue. What do you think?

byaminov:
@maartenbreddels I concur.

byaminov:
@maartenbreddels the build is green, can you please re-review or stamp?

maartenbreddels:
Awesome work! Thanks

Issue #410, opened 2019-09-06
Adding a new column from a 1D numpy array of numpy arrays


kushalkolar:
Adding a new DataFrame column from a 1D numpy array of numpy array raises exceptions.  Currently this seems to put the data into the actual dataframe but exceptions are raised from the array comparison in the DataFrame.add_column() method.

This illustrates the current behavior:
https://nbviewer.jupyter.org/github/kushalkolar/vaex/blob/numpy-fix/numpy_fix_info/current_behavior.ipynb

Naively changing the array comparison doesn't results in exceptions and the dataframe can be printed:

https://github.com/kushalkolar/vaex/commit/2afd71be1482e4963ad86c6f22f9602c76dc2230

https://nbviewer.jupyter.org/github/kushalkolar/vaex/blob/numpy-fix/numpy_fix_info/numpy-fix.ipynb

However this breaks normal functionality:
https://nbviewer.jupyter.org/github/kushalkolar/vaex/blob/numpy-fix/numpy_fix_info/breaks_stuff.ipynb

maartenbreddels:
Hi Kushal,

thanks for opening the issue!
that line tries to find all types in a column, apart from nan (that's why the k == k) and None values/types. This is done to check if a column is actually a string type. I think that np.all(k == k) will do the trick.

You could open a PR if you'd like.

cheers,

Maarten

JovanVeljanoski:
Hi @kushalkolar 

Thank you for the issue and the PR! 
Just a word of warning: a `vaex` column should really only be a 1d array, and not nested or nd array. There is no official support for arrays or lists being elements of a column at the moment, especially if the length of those lists is not identical for all rows (like in your example). You may run into various unpredictable issues with that usage. 

Support for this may be added down the line (@maartenbreddels might answer this better), but unlikely to come in the immediate future. 

In addition, I would avoid using columns of `dtype=`O`` whenever possible. 

Happy vaexing :)

kushalkolar:
@JovanVeljanoski Thanks for the warning, I'll so how far I can integrate this into our stuff!

JovanVeljanoski:
Closed via #411

PR #612, opened 2020-02-24
Fix: more efficient check for value types in a column when creating a DataFrame

byaminov:
I noticed that when converting a Pandas data frame to Vaex it can take a lot of time. For example when working with NYC Taxi dataset and reading a CSV file of 7.5 million rows:
* `df = pd.read_csv("yellow_tripdata_2019-01.csv")` takes 12 seconds;
* `vaex.from_pandas(df)` takes on 1 minute 31 seconds on `master`, and only 3 seconds with the change from this PR.

I've traced the problem to the check `np.all(k == k)`. As I understand it's needed to filter out `np.NaN`s, and a `k == k` check should do the same but runs faster.

@maartenbreddels as you have more understanding of this code, please check if this change does not break anything. And I can add the benchmarks or performance tests described above somewhere to the repo, if you advise where.


maartenbreddels:
This was introduced in #411 to solve #410, which allowed to have individual array elements (ndarray of type object) to be arrays as well. The comparison is a more general isnan comparison, but I'm not super happy with it either.
I fail to understand why it is so slow actually, I'll look into it a bit more.

maartenbreddels:
Thanks, Bulat, this was a great catch. I've added the test that #411 was missing, and rewrote it slightly.

byaminov:
@maartenbreddels cool thanks! The build seems to be [failing on `master`](https://travis-ci.org/vaexio/vaex/jobs/654470218?utm_medium=notification&utm_source=github_status) as well. Can you please try to reproduce on your Linux machine? Otherwise I can help later, by setting up some virtual linux. Seems like that test [is already skipped for a bunch of versions](https://github.com/vaexio/vaex/blob/4771c562b7bb8b2aaafd0beb1555dad12894aad8/tests/ml/ml_test.py#L247), so maybe we need to add another one...

JovanVeljanoski:
> @maartenbreddels cool thanks! The build seems to be [failing on `master`](https://travis-ci.org/vaexio/vaex/jobs/654470218?utm_medium=notification&utm_source=github_status) as well. Can you please try to reproduce on your Linux machine? Otherwise I can help later, by setting up some virtual linux. Seems like that test [is already skipped for a bunch of versions](https://github.com/vaexio/vaex/blob/4771c562b7bb8b2aaafd0beb1555dad12894aad8/tests/ml/ml_test.py#L247), so maybe we need to add another one...

Hi @byaminov 

Yeah that is a problematic test.. the issue is on the numpy side so.. I'll take a look at it again. If it is `RobustScaler`  that is failing, feel free to ignore or complain to me.

maartenbreddels:
Yes, there are 2 issues in CI, the robustscales, which has to do with the percentile_approx, which hints at a numpy issue, or an issue in vaex. That part will be refactored, so I don't want to spend too much time on it.
Also, because of a refactor, for the remote dataframe part, we sometimes run out of file handles. This is very recent, and actively thinking about it.

maartenbreddels:
Great work, we should open an issue to convert the benchmarks to asv, and run them at least nightly, so we detect these regressions more easily.

byaminov:
@maartenbreddels yes I like the idea of printing to an output file (did you mean CSV, otherwise what is asv?). IMO we can make it easy to measure and keep the output file committed for the history. Running them nightly can be quite flaky from my experience, but we can work on that as well.

For this PR - should we merge, as long as it's green?

byaminov:
Created an issue #620 to track progress of setting up the automatic runs of the benchmarks with [ASV](https://asv.readthedocs.io/en/stable/).

byaminov:
@maartenbreddels let me know if anything else is needed on this PR.

JovanVeljanoski:
LGTM!

maartenbreddels:
Looks very good, thanks a lot!

PR #615, opened 2020-02-25
fix: copy sparse matrices

### LINKED ISSUE #556: Support sparse data
The data in many domains, such as scRNA, is sparse.

maartenbreddels:
Closes #556


xdssio:
This is the use case I am looking at:

```
from sklearn.feature_extraction.text import TfidfVectorizer
import vaex
ds = vaex.from_arrays(x=['I try this','but this is hard','I should buy a cat'])
tdfidf = TfidfVectorizer()
s = tdfidf.fit_transform(ds.x)
ds.add_columns(tdfidf.get_feature_names(), s)
assert "error" not in repr(ds)

```

Another example which will crash also because columns can't start with a digit:
```
ds = vaex.from_arrays(x=['I try this','this 1and and 2and','I should buy a cat'])
```

maartenbreddels:
Yes, those a new bugs, i have fixes for them.

> ```
> ds = vaex.from_arrays(x=['I try this','this 1and and 2and','I should buy a cat'])
> ```

This just works right?

PR #617, opened 2020-02-26
Fix invalid name rewrites

### LINKED ISSUE #378: make classification from pandas issue
to reproduce:
```
from sklearn.datasets import make_classification
X, y = make_classification(10)
df = pd.DataFrame(X)
df['target'] = y
vaex.from_pandas(df)
```

results:
```
could not convert column 0, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 0, error: TypeError('getattr(): attribute name must be string',)
could not convert column 1, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 1, error: TypeError('getattr(): attribute name must be string',)
could not convert column 2, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 2, error: TypeError('getattr(): attribute name must be string',)
could not convert column 3, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 3, error: TypeError('getattr(): attribute name must be string',)
could not convert column 4, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 4, error: TypeError('getattr(): attribute name must be string',)
could not convert column 5, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 5, error: TypeError('getattr(): attribute name must be string',)
could not convert column 6, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 6, error: TypeError('getattr(): attribute name must be string',)
could not convert column 7, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 7, error: TypeError('getattr(): attribute name must be string',)
could not convert column 8, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 8, error: TypeError('getattr(): attribute name must be string',)
could not convert column 9, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 9, error: TypeError('getattr(): attribute name must be string',)
could not convert column 10, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 10, error: TypeError('getattr(): attribute name must be string',)
could not convert column 11, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 11, error: TypeError('getattr(): attribute name must be string',)
could not convert column 12, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 12, error: TypeError('getattr(): attribute name must be string',)
could not convert column 13, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 13, error: TypeError('getattr(): attribute name must be string',)
could not convert column 14, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 14, error: TypeError('getattr(): attribute name must be string',)
could not convert column 15, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 15, error: TypeError('getattr(): attribute name must be string',)
could not convert column 16, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 16, error: TypeError('getattr(): attribute name must be string',)
could not convert column 17, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 17, error: TypeError('getattr(): attribute name must be string',)
could not convert column 18, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 18, error: TypeError('getattr(): attribute name must be string',)
could not convert column 19, error: TypeError('getattr(): attribute name must be string',), will try to convert it to string
Giving up column 19, error: TypeError('getattr(): attribute name must be string',)
AttributeError                            Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968                 d['include'] = include
    969                 d['exclude'] = exclude
--> 970                 return method(**d)
    971             return None
    972         else:

5 frames
/usr/local/lib/python3.6/dist-packages/vaex/dataframe.py in column_filter(name)
   3486             if not strings and (self.dtype(name) == str_type or self.dtype(name).type == np.string_):
   3487                 return False
-> 3488             if not hidden and name.startswith('__'):
   3489                 return False
   3490             return True

AttributeError: 'int' object has no attribute 'startswith'
TypeError                                 Traceback (most recent call last)
<ipython-input-15-a7b44e392414> in <module>()
      3 df = pd.DataFrame(X)
      4 df['target'] = y
----> 5 vaex.from_pandas(df)

2 frames
/usr/local/lib/python3.6/dist-packages/IPython/core/formatters.py in format(self, obj, include, exclude)
    150             return {}, {}
    151 
--> 152         format_dict, md_dict = self.mimebundle_formatter(obj, include=include, exclude=exclude)
    153 
    154         if format_dict or md_dict:

TypeError: 'NoneType' object is not iterable
```

### LINKED ISSUE #558: Better handle unicode column names
Creating a dataframe with unicode chars like this:
```python
df = vaex.from_dict({'ÚÚ': vaex.vrange(0, 100)})
```
Will replace the two characters with two underscores, which effectively hides the column (column names beginning with `__` are hidden).
This was mentioned here:
https://stackoverflow.com/questions/59738879/python-vaex-how-to-create-dataframe-from-a-csv-file

maartenbreddels:
Should
 * Fix #378 (the 2nd issue
 * Fix https://stackoverflow.com/questions/59738879/python-vaex-how-to-create-dataframe-from-a-csv-file and Fix #558

A demo:
image


xdssio:
Is the get_column_names() a bit confusing? You would expect the same columns you see in the table.

maartenbreddels:
Yes, I agree it is not ideal, but this API is also used internally, and I'm afraid it will trigger bugs, which means I need to spend a lot of time reviewing the code and the test to know that it can be safe to do.
So it's a time issue, combined with low priority.

xdssio:
maybe we can make _get_user_column for internal stuff, and get_column_names for users.

xdssio:
I think that having a renaming alert could be a nice adding, but then we need a _vaex.warning_flag_ that the user can set it off and on.

In any case, if we rename columns, the table when you print should reflect that. so the names should also be changed there. 

It's a real pickle!

maartenbreddels:
I'm trying with adding an alias keyword to get_column_names(), so that by default it returns the aliased names. I will not support it in evaluate, so for instance you cannot do this:
```
for name in df.get_column_names():
  df.evaluate(name)
```
Instead, this works
```
for name in df.get_column_names(alias=False):
  df.evaluate(name)
# or 
for name in df.get_column_names():
  df[name].evaluate()
```

But evaluate is kind of low level anyway.

maartenbreddels:
A way to translate column names: `[str(df[k]) for k in df.get_column_names()]`

maartenbreddels:
@JovanVeljanoski @xdssio let me know if you like this, and if we should merge this as is.

xdssio:
This looks good.

The fact that a high-level user can work without being aware of the changes in the background is exactly what's makes it easy to use.

I agree, `evaluate` is low level already. 

Maybe we want to have a column's property of its translation like `df.column._core_name` or `df.column._key_name`.

maartenbreddels:
That would be something for a separate PR. I want to merge this first.

JovanVeljanoski:
I am also happy with this, at least for now. There may be some trouble with wrappers such as `vaex.ml.sklearn.IncrementalPredictor` and the tensorflow wrappers that will come in the near future, but that is something we can discuss/attack in due time. For now this is good!

maartenbreddels:
🎉

Issue #616, opened 2020-02-26
Feat: Tensorflow integration via Apache Arrow


maartenbreddels:
It looks like the tensorflow+tensorflowio integration seems possible:
https://github.com/tensorflow/io/issues/799
I propose we start adding a dataframe accessor `ml.tensorflow` (to https://github.com/vaexio/vaex/blob/master/packages/vaex-ml/setup.py#L28 )
But place it on the incubator, and not make tensorflow a dependency.

I think `df.ml.tensorflow.to_dataset()` could be useful, but maybe also a method that returns a dataset with a dict of tensors, and label for regression etc.

PR #622, opened 2020-03-01
Feature: Tensorflow Integration

JovanVeljanoski:
This PR realises #616 
The idea is enable vaex DataFrames to be used as input data sources for tensorflow (tf) data sources. 

This is made possible in large part thanks to [`tensorflow-io`](https://github.com/tensorflow/io) via arrow. 

- [x] A method to convert vaex DataFrames to tf datasets, to be used as input for tensorflow Estimators. This is now possible via `df.ml.tensorflow.to_dataset()`
- [x] A convenience function meant as a direct data source for tf Estimators. Note that tf Estimator consume an input function that returns a dataset object, rather than the dataset object itself.
- [x] Enabling the `to_dataset()` method do make the dataset compatible with Keras, so that one can use the dataset as an input for the Keras model `.fit()` method.
- [ ] A high level wrapper for the tf Estimators, which will make them easily consumable on the vaex side, while also turning them into vaex transformers, and also serialise them, and make them compatible with the state transfer system.
- [ ] A high level wrapper that will turn a pre-defined keras model into a vaex transformer.
- [ ] Unit tests (providing enough coverage & passing)
- [ ] Update the CHANGELOG when/if this PR is considered to be accepted
- [ ] Code review approval

Important notes:
- It is not readily obvious what the API of the high level wrapper for the tf Estimators should be. Typically, such wrappers need a `features` argument, used to select which features/columns of the vaex DataFrame should be used. To instantiate tf Estimators one need to provide one or more (in the case of `DNNLinearCombinedEstimator`) lists of `feature_columns` which give more detail to tf about the features, i.e. options on whether the features and numerical, categorical, providing various options of handing categorical variables, as well as options for engineering new features from the primary ones. In addition, some tf Estimators require various kwargs to be instantiated. Thus, over the next few days I will add some example code as comments here, from the user side, so we can decide on the type of API we should go with. 
- The test currently named `test_to_dataset_keras_model_train` that tests the successful passing of a vaex DataFrame to keras model via  a tf dataset fails with a segmentation fault.  At this point, I do not know whether this arises due to problems in vaex, tensorflow, tensorflow-io, or the code committed in this PR. However, my independent testing have also found other segmentation faults happening when making an input function or a dataset to be used for tf Estimators. Thus it is important that we investigate this issue.  
- At this point in time, we are using `tensorflow-io.arrow` to pass vaex data to keras, but this is not strictly necessary. It might be useful to investigate performance when using the tensorflow-io.arrow connection vs simply using vaex iterators  to pass numpy arrays to the keras models `.fit` method, since it can consume generators that return numpy arrays,

### DIFF
--- a/packages/vaex-ml/setup.py
+++ b/packages/vaex-ml/setup.py
@@ -21,9 +21,11 @@ setup(name=name + '-ml',
       author=author,
       author_email=author_email,
       install_requires=install_requires_ml,
+      extras_require={'ml.tensorflow': ['tensorflow>=2.1.0', 'tensorflow-io>=0.12.0']},
       license=license,
       packages=['vaex.ml', 'vaex.ml.incubator', 'vaex.ml.datasets'],
       include_package_data=True,
       zip_safe=False,
-      entry_points={'vaex.dataframe.accessor': ['ml = vaex.ml:DataFrameAccessorML']}
+      entry_points={'vaex.dataframe.accessor': ['ml = vaex.ml:DataFrameAccessorML',
+                                                'ml.tensorflow = vaex.ml.tensorflow:DataFrameAccessorTensorflow']}
 )
--- a/packages/vaex-ml/vaex/ml/tensorflow.py
+++ b/packages/vaex-ml/vaex/ml/tensorflow.py
@@ -1,0 +1,70 @@
+from functools import partial
+
+
+import tensorflow as tf
+
+import tensorflow_io.arrow as arrow_io
+from tensorflow_io.arrow.python.ops.arrow_dataset_ops import arrow_schema_to_tensor_types
+
+
+class DataFrameAccessorTensorflow(object):
+    def __init__(self, ml):
+        self.ml = ml
+        self.df = self.ml.df
+
+    def _arrow_batch_generator(self, features, target=None, chunk_size=1024):
+        column_names = features + [target] if target is not None else features
+        for i1, i2, table in self.df.to_arrow_table(column_names=column_names, chunk_size=chunk_size):
+            yield table.to_batches(chunk_size)[0]
+
+    @staticmethod
+    def _get_batch_arrow_schema(arrow_batch):
+        output_types, output_shapes = arrow_schema_to_tensor_types(arrow_batch.schema)
+        return output_types, output_shapes
+
+    def to_dataset(self, features, target=None, chunk_size=1024, repeat=None, shuffle=False, keras=False):
+
+        # Set up the iterator factory
+        iterator_factory = partial(self._arrow_batch_generator, **{'features': features,
+                                                                   'target': target,
+                                                                   'chunk_size': chunk_size})
+        # get the arrow schema
+        output_types, output_shapes = self._get_batch_arrow_schema(next(iterator_factory()))
+
+        # Define the TF dataset
+        ds = arrow_io.ArrowStreamDataset.from_record_batches(record_batch_iter=iterator_factory(),
+                                                             output_types=output_types,
+                                                             output_shapes=output_shapes,
+                                                             batch_mode='auto',
+                                                             record_batch_iter_factory=iterator_factory)
+
+        # Reshape the data into the appropriate format
+        if keras:
+            if target is not None:
+                ds = ds.map(lambda *tensors: (tf.stack(tensors[:-1], axis=1), tensors[-1]))
+            else:
+                ds = ds.map(lambda *tensors: (tf.stack(tensors, axis=1)))
+        else:
+            if target is not None:
+                ds = ds.map(lambda *tensors: (dict(zip(features, tensors[:-1])), tensors[-1]))
+            else:
+                ds = ds.map(lambda *tensors: (dict(zip(features, tensors))))
+
+        # Repeating and shuffling the dataset if needed
+        if shuffle:
+            ds = ds.shuffle(chunk_size)
+        if repeat is not None:
+            ds = ds.repeat(repeat)
+
+        return ds
+
+    def make_input_function(self, features, target=None, chunk_size=1024, repeat=None, shuffle=False):
+        def tf_input_function():
+            return self.to_dataset(features=features,
+                                   target=target,
+                                   chunk_size=chunk_size,
+                                   repeat=repeat,
+                                   shuffle=shuffle)
+        return tf_input_function
+
+
--- a/tests/ml/tensorflow_test.py
+++ b/tests/ml/tensorflow_test.py
@@ -1,0 +1,112 @@
+import logging
+
+import pytest
+pytest.importorskip("tensorflow")
+pytest.importorskip("tensorflow_io")
+
+import tensorflow as tf
+
+import vaex
+import vaex.ml
+
+
+
+logger = tf.get_logger()
+logger.setLevel(logging.ERROR)
+
+
+# Set up the data to be used in this test file
+df = vaex.ml.datasets.load_iris()
+df_train, df_test = df.ml.train_test_split(test_size=0.2, verbose=False)
+features = df.column_names[:4]
+target = 'class_'
+
+
+def test_to_dataset_tensorflow():
+    ds = df_test.ml.tensorflow.to_dataset(features=features, target=target, chunk_size=10)
+    list_ds = list(ds)
+    assert len(list_ds) == 3  # The number of "batches" in the iterable, as defined by the chunk_size arg
+
+    idx_min = 0
+    idx_max = 10
+    for batch in list_ds:
+        assert len(batch) == 2  # check that each batch has feature matrix and target matrix (hence tuple of len 2)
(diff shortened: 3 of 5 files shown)


packages/vaex-ml/setup.py
```diff
@@ -21,9 +21,11 @@
       author=author,
       author_email=author_email,
       install_requires=install_requires_ml,
+      extras_require={'ml.tensorflow': ['tensorflow>=2.1.0', 'tensorflow-io>=0.12.0']},
```
maartenbreddels: This is not easy to type, let's call it all

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,70 @@
...
+        column_names = features + [target] if target is not None else features
+        for i1, i2, table in self.df.to_arrow_table(column_names=column_names, chunk_size=chunk_size):
+            yield table.to_batches(chunk_size)[0]
+
+    @staticmethod
+    def _get_batch_arrow_schema(arrow_batch):
+        output_types, output_shapes = arrow_schema_to_tensor_types(arrow_batch.schema)
+        return output_types, output_shapes
+
+    def to_dataset(self, features, target=None, chunk_size=1024, repeat=None, shuffle=False, keras=False):
```
maartenbreddels: I think keras should be called as_dict (inverting the Boolean logic)

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,70 @@
...
+            if target is not None:
+                ds = ds.map(lambda *tensors: (dict(zip(features, tensors[:-1])), tensors[-1]))
+            else:
+                ds = ds.map(lambda *tensors: (dict(zip(features, tensors))))
+
+        # Repeating and shuffling the dataset if needed
+        if shuffle:
+            ds = ds.shuffle(chunk_size)
+        if repeat is not None:
+            ds = ds.repeat(repeat)
```
maartenbreddels: I don't think we should do this, since it is too trivial, and we would need to explain this. The map to dict is less trivial I would say, so good reason to support that.

tests/ml/tensorflow_test.py
```diff
@@ -0,0 +1,112 @@
...
+def test_to_dataset_tensorflow():
+    ds = df_test.ml.tensorflow.to_dataset(features=features, target=target, chunk_size=10)
+    list_ds = list(ds)
+    assert len(list_ds) == 3  # The number of "batches" in the iterable, as defined by the chunk_size arg
+
+    idx_min = 0
+    idx_max = 10
+    for batch in list_ds:
+        assert len(batch) == 2  # check that each batch has feature matrix and target matrix (hence tuple of len 2)
+        # Assert the target
```
maartenbreddels: These comments don't add anything, they are only harmful, since they don't explain anything, and will lie in the future if we forgot to update it.

tests/ml/tensorflow_test.py
```diff
@@ -0,0 +1,112 @@
...
+        assert batch[1].numpy().tolist() == df_test[target][idx_min:idx_max].tolist()
+        # Assert each feature individually
```
maartenbreddels: Same

packages/vaex-ml/vaex/ml/tensorflow.py
JovanVeljanoski: Yeah, ok, these lines should be moved to `make_input_function()` method instead. There is it necessary.

packages/vaex-ml/vaex/ml/tensorflow.py
JovanVeljanoski: Yeah we could do that. I thought this way it is more obvious which option is for what. I guess we can have it as_dict as you suggest, and describe in the docstring what it is for?

tests/ml/tensorflow_test.py
JovanVeljanoski: Well, we disagree about the use of comments consistently so.. To me the code without the comments is a bit hard to follow is I naively would not know what to expect, so the comments are just reminders of that the lines are meant to accomplish. 

It is easy to remove, i have no problem with it.

packages/vaex-ml/setup.py
JovanVeljanoski: Done

packages/vaex-ml/vaex/ml/tensorflow.py
JovanVeljanoski: Now it is implemented `as_dict=True` (use to be `keras=False`)

tests/ml/tensorflow_test.py
JovanVeljanoski: Removed most of the comments

JovanVeljanoski:
Substantial update to this PR. 
Now we have support for multi-class classification and multi-target regression. One can pass `to_dataset()`  datasets to both tf estimators and keras models. For keras models, one can construct the dataset as dicts (similar as in the case of the tf estimators), or as tensors. 

For passing data to Keras models, now we can also just use vaex as the main generator, no need to go through tensorflow-io. (I have not checked which one is faster).

For comprehensive examples on how all this works, please see the unit-tests. There are now more detailed unit tests, which test both the creation of the data sources, and how they are passed on to models as well.

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,168 @@
...
+                                                             record_batch_iter_factory=iterator_factory)
+
+        # Reshape the data into the appropriate format
+        if as_dict:
+            if target is not None:
+                if n_target_cols == 1:
+                    ds = ds.map(lambda *tensors: (dict(zip(features, tensors[:-1])), tensors[-1]))
+                else:
+                    ds = ds.map(lambda *tensors: (dict(zip(features, tensors[:-n_target_cols])),
+                                                  # dict(zip(target, tensors[-n_target_cols:]))))
```
maartenbreddels:
```suggestion
```

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,168 @@
...
+import vaex
+
+
+class DataFrameAccessorTensorflow(object):
+    def __init__(self, ml):
+        self.ml = ml
+        self.df = self.ml.df
+
+    def _arrow_batch_generator(self, column_names, chunk_size=1024, parallel=True):
+        """Create a generator which yields arrow table batches, to use as datasoure for creating Tensorflow datasets.
```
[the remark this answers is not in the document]
JovanVeljanoski: Well, i wrote the docstrings for us more than to make the method public. But if you think we should expose it, i have no problem with it

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,168 @@
...
+    @staticmethod
+    def _get_batch_arrow_schema(arrow_batch):
+        """Get the schema from a arrow batch table."""
+        output_types, output_shapes = arrow_schema_to_tensor_types(arrow_batch.schema)
+        return output_types, output_shapes
+
+    def to_dataset(self, features, target=None, chunk_size=1024, as_dict=True, parallel=True):
+        """Create a tensorflow Dataset object from a DataFrame, via Arrow.
+
+        :param features: A list of column names.
```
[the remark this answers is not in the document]
JovanVeljanoski: Well, indeed we can. I guess I did it for consistency with the rest of vaex.ml. 
We include  `features` argument everywhere..  do we want to remove all of those as well? 

Or should keep this here, and then remove it in a separate PR consistently everywhere?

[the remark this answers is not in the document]
maartenbreddels: In any case, I think it should be an optional argument right?

packages/vaex-ml/vaex/ml/tensorflow.py
```diff
@@ -0,0 +1,168 @@
...
+        :param target: The dependent or target column, if any.
```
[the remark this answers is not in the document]
maartenbreddels: And when you only give target, I expect features to be all columns but the target

solalatus:
This would be such a killer feature! Any news on this?

JovanVeljanoski:
Hi @solalatus 

We are happy you think so!

Yeah, it is ready. Perhaps some cleanup is needed. We are doing some internal testing on big-ish data together with an involved (vaex based) pre-processing pipeline to make sure everything will actually work as intended. 

If you are comfortable with a dev install you can try this branch. Otherwise, if all goes well I estimate 1-2 weeks before it is merged and released.

solalatus:
Marvelous news! I will be talking about data iterators soon in a teaching context, so I am now more confident to hint at Vaex!
Keep pushing guys, it is cool stuff!

PR #639, opened 2020-03-11
Fix: inconsistent evaluation in to_arrow_table

JovanVeljanoski:
This is to fix a flaky/inconsistent evaluation of expressions using the `to_arrow_table` method. 
The better summarize the issue:
- It happens only (to the best of my knowledge) when an expression is make of masked arrays, or is a virtual column calculated on the basis of masked arrays
- Not sure if it happens if the evaluation is *not* in chunks
- Does not happen if the `chunk_size` is very small (like less than 5 or so)
- Happens more rarely if the `chunk_size` is large compared to the size of the dataset (like 50% of the df size or more), although I can not guarantee this in all cases. 
- This does not seem to be the case with other types of missing values like `None` and `np.nan`.

I made a unit-test exposing this. Not the most concise of test, but this is the most I could narrow the issue down. Suggestions welcome. 

Checklist:
- [x] Unit-test reproducing the issue
- [ ] Making the unit-test pass
- [ ] Manual testing

### DIFF
--- a/packages/vaex-arrow/vaex_arrow/convert.py
+++ b/packages/vaex-arrow/vaex_arrow/convert.py
@@ -8,6 +8,7 @@ def arrow_array_from_numpy_array(array):
     mask = None
     if np.ma.isMaskedArray(array):
         mask = array.mask
+        # arrow 0.16 behaves weird in this case https://github.com/vaexio/vaex/pull/639
         if mask is np.False_:
             mask = None
         elif mask is np.True_:
--- a/packages/vaex-arrow/vaex_arrow/convert.py
+++ b/packages/vaex-arrow/vaex_arrow/convert.py
@@ -8,6 +8,10 @@ def arrow_array_from_numpy_array(array):
     mask = None
     if np.ma.isMaskedArray(array):
         mask = array.mask
+        if mask is np.False_:
+            mask = None
+        elif mask is np.True_:
+            raise ValueError('not sure what pyarrow does with mask=True')
         array = array.data
     if dtype.kind == 'S':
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I am not sure this is a pyarrow feature or bug, we should check their Jira, but basically:
```python
>>> import pyarrow as pa
>>> pa.Array.from_pandas(np.arange(10), mask=True)
<pyarrow.lib.Int64Array object at 0x11a6b2168>
[
  null,
  1,
  2,
  3,
  4,
  5,
  6,
  null,
  null,
  null
]
```
It interprets the mask very strangely, actually i you pass a list (and not an array, you will get):
```python
>>> import pyarrow as pa
>>> pa.Array.from_pandas(range(10), mask=True)
ArrowInvalid                              Traceback (most recent call last)
<ipython-input-13-92f2033e9674> in <module>
      1 import pyarrow as pa
----> 2 pa.Array.from_pandas(range(10), mask=True)

~/miniconda3/envs/dev/lib/python3.7/site-packages/pyarrow/array.pxi in pyarrow.lib.Array.from_pandas()

~/miniconda3/envs/dev/lib/python3.7/site-packages/pyarrow/array.pxi in pyarrow.lib.array()

~/miniconda3/envs/dev/lib/python3.7/site-packages/pyarrow/array.pxi in pyarrow.lib._sequence_to_array()

~/miniconda3/envs/dev/lib/python3.7/site-packages/pyarrow/error.pxi in pyarrow.lib.check_status()

ArrowInvalid: Null mask must be NumPy array
```.

maartenbreddels:
It seems fixed in pyarrow 0.17 btw.

PR #640, opened 2020-03-11
Fix: evaluation of a dataframe containing masked arrays

JovanVeljanoski:
When a dataframe (the whole dataframe, not an expression) is evaluated with the `.values` method, if a column contains masked values, those values will be realized as some numeric values (i think the masked values is filled with some large number). In the case of `numpy.ma.MaskedArray`, i expect the masked values to be "translated" to `np.nan`. This is what the unit-test assumes. 

This may be related to #639 although I am not quite sure. 

Checklist:
- [x] Unit-test that catches the issue
- [x] Making the unit-test pass 
- [ ] Manual testing

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5108,7 +5108,10 @@ class DataFrameLocal(DataFrame):
                 if self.dtype(name) != dtype:
                     raise ValueError("Cannot cast %r (of type %r) to %r" % (name, self.dtype(name), dtype))
         chunks = self.evaluate(column_names, parallel=parallel)
-        return np.array(chunks, dtype=dtype).T
+        if any(np.ma.isMaskedArray(chunk) for chunk in chunks):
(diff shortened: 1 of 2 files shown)


maartenbreddels:
> i expect the masked values to be "translated" to `np.nan`

I would say, a masked array, not nan, since the underlying data is a masked array, would you agree?

maartenbreddels:
Should we say `.values` gives a masked array, and we can change the semantics of np.array(df) later?

JovanVeljanoski:
Yes that makes sense. 
So as we discussed elsewhere, is it possible to keep close (or follow exactly your proposal):
- `.values` will return the internal structure, if numpy array (masked) it will return a numpy array (masked if needed)
- to get a get a numpy array always we will keep to `np.array(df)`
- to get arrow array we will keep to `pa.array(df)`

maartenbreddels:
We forgot the changelog :(

maartenbreddels:
updated in master

PR #651, opened 2020-03-24
Fix: `count` issue with `limits='minmax'`

JovanVeljanoski:
This PR fixes a potential issue with `df.count(expression=some_expression, limit='minmax')`
This issue is especially apparent when there is a selection applied to the dataframe.

- [x] Unit-tests exposing the problem
- [ ] Unit-tests passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
       * Fix `colum_count`, now only counts hidden columns if expicitly specified [#593](https://github.com/vaexio/vaex/pull/593)
       * df.values respects masked arrays [#640](https://github.com/vaexio/vaex/pull/640)
       * Rewriting a virtual column and doing a state transfer does not lead to `ValueError: list.remove(x): x not in list` [#592](https://github.com/vaexio/vaex/pull/592)
+      * `df.<stat>(limits=...)` will now respect the selection [#651](https://github.com/vaexio/vaex/pull/651)
    * Features
       * New lazy numpy wrappers: np.digitize and np.searchsorted [#573](https://github.com/vaexio/vaex/pull/573)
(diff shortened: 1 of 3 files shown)


maartenbreddels:
The test does not expose an issue I think, the test is faulty, you were not passing in min and max, and the last bin for numpy.histogram is treated differently:
```patch
diff --git a/tests/count_test.py b/tests/count_test.py
index dda3d91d..e929c48a 100644
--- a/tests/count_test.py
+++ b/tests/count_test.py
@@ -33,13 +33,13 @@ def test_count_1d_verify_against_numpy(ds_local, limits):
 
     # bin with vaex
     xmin, xmax = df.limits(expression=expression, value=limits, selection=selection)
-    vaex_counts = df.count(binby=[expression], selection=selection, shape=shape, limits=limits)
+    vaex_counts = df.count(binby=[expression], selection=selection, shape=shape, limits=[xmin, xmax])
 
     # bin with numpy
     x_values = df[selection][expression].values
     numpy_counts, numpy_edges = np.histogram(x_values, bins=shape, range=(xmin, xmax))
 
-    assert vaex_counts.tolist() == numpy_counts.tolist()
+    assert vaex_counts[:-1].tolist() == numpy_counts[:-1].tolist() 
```

JovanVeljanoski:
Ok the test is partly faulty, or maybe my coding has confused you:
 - the calculation of `xmin, xmax = ..` is for the `np.histogram(... , range=(xmin, xmax))` method, not for the `df.count` method, and this is not a mistake. I expect that output from count to be the same where one calculates the limits via `df.limits` or via `df.count`. 

- Indeed the last bin is defined differently, and I will update the test to reflect this. 

Does this make sense to you?

Issue #631, opened 2020-03-05
`isin` Not Working Properly


Alon-Alexander:
Hi,
Putting an example usage here.
Do you have any idea why isn't it working as it is supposed to (like pandas `isin`)?

![image](https://user-images.githubusercontent.com/19497150/75974016-c7705580-5ede-11ea-8d43-9b68c0b3935d.png)

JovanVeljanoski:
Hi @Alon-Alexander 

Thanks for the report and for the reproducible example!

Curious.. if you instantiate a vaex dataframe via numpy, there is no such problem.. My guess is that this has something to do with the 'object' type of pandas, but I am not sure. It should have been converted to string..  (screenshot below).

@maartenbreddels should I make a test for this and classify as bug?

![image](https://user-images.githubusercontent.com/18574951/75976544-ed93f680-5eda-11ea-91c8-6b1e30e20928.png)

maartenbreddels:
yeah, please do, looks like a bug.

maartenbreddels:
Fixed in #669

PR #668, opened 2020-04-12
`df.isin` fix for data imported from a pandas dataframe

JovanVeljanoski:
Test for the issue exposed by #631 

- [x] Make unit-test
- [ ] Make the test pass

### DIFF
--- a/tests/isin_test.py
+++ b/tests/isin_test.py
@@ -1,5 +1,8 @@
-import vaex
 import numpy as np
+
+import pandas as pd
+
+import vaex
 
 
 def test_isin():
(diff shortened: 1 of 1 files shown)


tests/isin_test.py
```diff
@@ -17,3 +20,15 @@ def test_isin():
     assert df.w.isin([2, None]) == [True, False, True]
     assert df.m.isin([1, 2, 3]) == [False, False, True]
     assert df.n.isin([2, np.nan]) == [False, True, False]
+
+
+def test_isin_pandas_object():
+    df_pandas = pd.DataFrame({'x': np.array(['a', 'b', 'c']),
+                              'y': np.array([1, 2, 3], dtype='O')})
```
maartenbreddels, line 27: I think it has nothing to do with pandas, but with dtype=object? If so, could you rename everything? Now it is misleading as if we treat data from pandas differently. I understand that it shows up when using a pandas dataframe, but it is no the underlying issue.

PR #671, opened 2020-04-14
feat: support xarray format for quick aggregators (e.g. df.count(..))

maartenbreddels:
This is to be in line with #654 and the binby supporting xarray.
Example:
image

Also, this integrates well with plotly express' latest xarray support (cc @nicolaskruchten):
image


maartenbreddels:
@JovanVeljanoski should format be renamed to `array_type='xarray'`?

JovanVeljanoski:
> @JovanVeljanoski should format be renamed to `array_type='xarray'`?

Yes, i quite prefer this actually, great idea!

maartenbreddels:
@JovanVeljanoski should we merge this, and we do docs separately, by pushing to the docs branch?

JovanVeljanoski:
Let me test something out in the next 30 minutes or so, but it looks good. 

In any case we will do docs separately, so this will not be delayed due to that

PR #672, opened 2020-04-14
chore: setup benchmarks with ASV

### LINKED ISSUE #620: Chore: run benchmarks periodically to record improvements or detect regressions
There are some benchmarks in Vaex that let measure performance of various functions, for example the [`benchmarks/strings.py`](https://github.com/vaexio/vaex/blob/master/benchmarks/strings.py). It would be useful if we can run those benchmarks periodically (for example, nightly) and record the results. Then when there is a performance improvement (like for example #612), a benchmark run history can show the improvement. And if a certain change would decrease the performance of certain functions, we could detect it automatically.

Let's do the following:
- [ ] Setup [airspeed velocity](https://asv.readthedocs.io/en/stable/) in the repository to let run the benchmarks. 
- [ ] `strings.py` and `from_pandas.py` can be used as the first benchmark tests.
  - [ ] Look also for other existing benchmarks in the repository.
- [ ] The benchmark results should be published as GitHub pages in the Vaex repository.
- [ ] Document the process of running a benchmark locally and developing new benchmarks.
  - as a README file in the benchmarks folder.
- [ ] Configure a physical Linux laptop to run the benchmarks every night or every week.

byaminov:
This closes #620.


maartenbreddels:
Cool, is it maybe an idea to open a separate PR to get the tests and ASV configuration in? Then I can start adding more benchmarks soon. In the meantime, we can try out different options for running them.

byaminov:
@maartenbreddels done!

maartenbreddels:
I had an idea of reusing the big files each time, so they are generated only once (per machine). It seems that the working directory is different each time :(.

Can you find anything on this topic?

byaminov:
> Can you find anything on this topic?

Yes you can use the [`setup_cache`](https://github.com/vaexio/vaex/pull/672/files#diff-f012523e3ef85c7f27976c7540bf99a8R21) to make sure that a big file is downloaded/generated in a location you want (e.g. `~/.vaex/data` as you use in `datasets.py`), and use that location in the `setup` function to open a `df`.

benchmarks/strings.py
```diff
@@ -1,166 +1,119 @@
...
+class StringsSuite:
+    pretty_name = "Performance of string operations in DataFrames"
+    version = "1"
+    timeout = 3600  # in seconds
+
+    param_names = [
+        'n',  # log number of rows to use
+        'partitions',  # number of partitions to split
+    ]
+    params = ([6], [24])
```
byaminov, line 16: @maartenbreddels here I used parameters to showcase how they can be used, and to check if you want to run these benchmarks with different number of filtered rows or partitions. If not - then we can get rid of these parameters, as with single value in each of them it's an overkill.

benchmarks/strings.py
```diff
@@ -1,166 +1,119 @@
...
+
+    dataframe_nmax = 7  # log number of rows for test dataset
```
byaminov, line 18: @maartenbreddels here I chose a number that can be comfortably generated within a test, at least when running yourself. We can increase this if you want to run the benchmarks on a bigger scale. But from other projects seems like measuring milliseconds of execution on small amounts of data is a common practice.

maartenbreddels: Yes, that's exactly what I'm thinking of, 10**9 :)
I also wanna do msec benchmarking, but this is also a use case, i want to keep running the old string benchmarks before I refactor that.

benchmarks/strings.py
maartenbreddels: What I am experimenting with now is sth like this:
```
class Binning:
    params = ([10**k for k in [8, 9]], [np.float64])
    param_names = ['N', 'dtype']
    def setup(self, N, dtype):
        basedir = os.environ['VAEX_BENCHMARK_DIR']
        self.df = vaex.open(os.path.join(basedir, 'numerical.hdf5'))[:N]
```

so I generate 1 big file (maybe a setup script that runs outside of the benchmark suite), and then slice what I need from it. But I am not sure I like the VAEX_BENCHMARK_DIR env var. I'd rather make a symlink from benchmarks/data to a data directory specific to the machine. But the CWD doesn't seem to be the benchmarks directory.

benchmarks/strings.py
byaminov: ok, then it's probably better to pre-generate it and let the benchmark test download it. When I tried with 10**9 it took afair over 10 minutes and then I did not wait anymore.

benchmarks/strings.py
byaminov: 👍 As I mentioned in another comment I don't see a problem with `~/.vaex/data` as used in `datasets.py`.

maartenbreddels: I missed that, I like that idea!

maartenbreddels:
Shall I push something here to share these ideas in code?

byaminov:
> Shall I push something here to share these ideas in code?

Sure!

maartenbreddels:
I rebased and had to force push, what do you think of this approach?

byaminov:
> I rebased and had to force push, what do you think of this approach?

looks good! I guess this is not the final code so I'm not going to make a review.

I guess you'd have a `def setup_cache(self)` which calls the `generate_numerical()`?

For the tests themselves I trust you know what you're doing :) I did not fully understand what's being tested there, and why different combinations of columns and parameters matter. But I agree with the general approach anyway.

maartenbreddels:
> I guess you'd have a `def setup_cache(self)` which calls the `generate_numerical()`?

good idea.

> I did not fully understand what's being tested there, and why different combinations of columns and parameters matter.

It test binby/groupby for particular cases. For instance, having a small number of unique integers makes the hashtable much smaller, and will be better to parallelize (each thread has its own hashtable). https://github.com/vaexio/vaex/issues/684 exposes an issue with this, so it's good to have a baseline of our performance now.

maartenbreddels:
Do you want to refactor the string testing to be similar? with 10^9 testing? I'd like to run it on the laptop soon, since I want to take a part of https://github.com/vaexio/vaex/pull/345/files into a new PR and see the performance impact.

byaminov:
@maartenbreddels I've downsized the datasets to 10^8, as then the generated file sizes are within 10Gb, while I did not see much difference in measured performance. So it's practically more usable to run locally and develop new tests.

With the current setup it took 7 minutes to run the whole suite on my laptop (with pre-generated datasets).

Can you please have a look at the code, and comment if anything else is needed in this PR?

benchmarks/aggregates.py
```diff
@@ -0,0 +1,85 @@
+import vaex
+
+from benchmarks.fixtures import generate_numerical
+
+
+class Aggregates:
+    pretty_name = "Performance of aggregates: stats, binby etc"
+    version = "1"
+
+    params = ([10**7, 10**8, 5*10**8, 10**9],)
```
maartenbreddels, line 10:
```suggestion
    params = ([10**7, 10**8, 5*10**8],)
```
I guess this doesn't run with 10**9?

byaminov: right, thanks for the catch! Will fix now.

maartenbreddels:
Great work Bulat!

PR #674, opened 2020-04-15
fix(expresso): support floor div, fixes #634

### LINKED ISSUE #634: Error printing DataFrames with certain types of virtual columns
When trying to print or display a dataframe with a certain type of virtual columngs, but only using a subset of the columns crashes with `NameError: name 'do_not_understand_expression' is not defined`.

See a fully reproducible example below.

```
import vaex

df = vaex.from_arrays(x=['dog', 'cat', 'cat', 'dog', 'mouse'], time=[2230, 0, 1200, 1515, 945])
# Extract the hour from the time column
df['hour'] = df.time // 100 % 24

# Printing/displaying the  entire DataFrame works:
print(df)

# Printing/displaying a subset of columns, which includes the virtual column created above crashes
print(df[['x', 'hour']])
```

maartenbreddels:
This also includes yet uncommitted tests for the expresso module.

### DIFF
--- a/packages/vaex-core/vaex/expresso.py
+++ b/packages/vaex-core/vaex/expresso.py
@@ -357,6 +357,8 @@ class ExpressionString(ast.NodeVisitor):
                 return "({left} * {right})".format(left=left, right=right)
             if isinstance(node.op, ast.Div):
                 return "({left} / {right})".format(left=left, right=right)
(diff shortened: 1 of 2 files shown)


maartenbreddels:
GHA is a bit flakey, merging this.

Issue #661, opened 2020-04-07
concatenate with masked arrays


maartenbreddels:
```python
import vaex
import numpy as np
df1 = vaex.from_arrays(x=np.arange(3.))
df2 = vaex.from_arrays(x=np.ma.masked_all(4, dtype='f8'))
df1.concat(df2)
```
Gives:
```
~/src/vaex/packages/vaex-core/vaex/column.py in __init__(self, expressions, dtype)
    145                             self.fill_value = expressions[0][0:1].fill_value
    146                         except:
--> 147                             self.fill_value = expressions[0].values.fill_value
    148 
    149             any_strings = any([dtype == str_type for dtype in dtypes])

AttributeError: 'numpy.ndarray' object has no attribute 'fill_value'
```

More strangely:
```
import vaex
import numpy as np

@vaex.register_dataframe_accessor('mytool', override=True)
class mytool:
    def __init__(self, df):
        self.df = df

    def shift(self, column, n, cyclic=True, fill_value=None, inplace=False):
        # make a copy without column
        df = self.df.copy().drop(column)
        # make a copy with just the colum
        df_column = self.df[[column]]
        # slice off the head and tail
        if cyclic:
            df_head = df_column[-n:]
        else:
            df_head = vaex.from_dict({column: np.ma.masked_all(n, dtype=self.df[column].dtype)})
        df_tail = df_column[:-n]
        # stitch them together
        df_shifted = df_head.concat(df_tail)
        # and join (based on row number)
        return df.join(df_shifted, inplace=inplace)

x = np.arange(10)
y = x**2
df = vaex.from_arrays(x=x, y=y)
df['shifted_y'] = df.y
df2 = df.mytool.shift('shifted_y', 2, cyclic=False)
df2
```
This silently does not include the dropped column, (a non-cyclic version of #660 )

okapies:
Hi, I'm interested to use this feature in my product. What is a blocker for merging #925 into the mainline?

maartenbreddels:
All full-fledged version of shift requires a more significant effort than this code snippet, which should work btw! Did you already try this?

If you or your company has the budget to sponsor development for this feature, feel free to contact us via https://vaex.io/feature

PR #678, opened 2020-04-16
Feature: export_csv

JovanVeljanoski:
This PR adds `.export_csv` method to the dataframe.
For DataFrames that are too large to fit in memory, the export is done in chunks, so one only needs to have enough disk space to house the output file.

- [x] Implementation
- [x] Testing 
- [x] Review

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5767,7 +5767,7 @@ class DataFrameLocal(DataFrame):
     def export_csv(self, path, virtual=False, selection=False, progress=None, chunk_size=1_000_000, **kwargs):
         """ Exports the DataFrame to a CSV file.
 
-        :param str path: Path for filename
+        :param str path: Path for file
         :param bool virtual: If True, export virtual columns as well
         :param bool selection: {selection1}
         :param progress: {progress}
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5767,8 +5767,7 @@ class DataFrameLocal(DataFrame):
         :param str path: Path for filename
         :param bool virtual: If True, export virtual columns as well
         :param bool selection: If True, export the selection
-        :param progress: Progress callback that gets a progress fraction as argument and should return True to continue,
-                or a default progress bar when progress=True
+        :param progress: {progress}
         :param int batch_size: Number of rows to be written to disk in a single iteration
         :param **kwargs: Extra keyword arguments to be passed on pandas.DataFrame.to_csv()
         :return:
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5764,14 +5764,14 @@ class DataFrameLocal(DataFrame):
         vaex.export.export_fits(self, path, column_names, shuffle, selection, progress=progress, virtual=virtual, sort=sort, ascending=ascending)
 
     @docsubst
-    def export_csv(self, path, virtual=False, selection=False, progress=None, batch_size=1_000_000, **kwargs):
+    def export_csv(self, path, virtual=False, selection=False, progress=None, chunk_size=1_000_000, **kwargs):
         """ Exports the DataFrame to a CSV file.
 
         :param str path: Path for filename
         :param bool virtual: If True, export virtual columns as well
         :param bool selection: {selection1}
         :param progress: {progress}
-        :param int batch_size: {chunk_size_export}
+        :param int chunk_size: {chunk_size_export}
         :param **kwargs: Extra keyword arguments to be passed on pandas.DataFrame.to_csv()
         :return:
         """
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/dataframe.py
```diff
@@ -5747,6 +5751,40 @@ def export_fits(self, path, column_names=None, shuffle=False, selection=False, p
         import vaex.export
         vaex.export.export_fits(self, path, column_names, shuffle, selection, progress=progress, virtual=virtual, sort=sort, ascending=ascending)
 
+    def export_csv(self, path, virtual=True, selection=False, progress=None, batch_size=1_000_000, **kwargs):
+        """ Exports the DataFrame to a CSV file.
+
+        :param str path: Path for filename
+        :param bool virtual: If True, export virtual columns as well
```
maartenbreddels, line 5758: The default value for virtual is False for the others, should we remove it altogether. Or change all the other default to True?

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5747,6 +5751,40 @@ def export_fits(self, path, column_names=None, shuffle=False, selection=False, p
...
+        :param bool selection: If True, export the selection
+        :param progress: Progress callback that gets a progress fraction as argument and should return True to continue,
```
maartenbreddels, line 5760: I guess this is one of the doc snippets

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: Ah my bad, i was testing it and forgot the switch it back. I say for now let's keep everything consistent so virtual=False by default indeed. Thanks!

JovanVeljanoski:
Hi @maartenbreddels 

Thanks for your comments. I did some better testing (included in the PR). I also uncovered some issues when exporting columns that have masked values, so the implementation is changed a bit now.

Also the new test uncovers a bit of a deeper issue in vaex, not related to this PR: and error is being raised when using `evaluate_iterator` on a DataFrame that is concatenated and than filtered. This is why this PR is failing. 

I can open a separate PR with a test for that if you wish, let me know.

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: I don't understand what is the issue here, can you please explain?

maartenbreddels: like 
`        :param progress: {progress}`
We have to reuse the doc snippets not to go crazy :)

tests/export_test.py
```diff
@@ -41,6 +40,14 @@ def test_export_open_hdf5(ds_local):
     ds_opened = vaex.open(filename)
     assert list(ds) == list(ds_opened)
 
+def test_export_open_csv(ds_local):
+    df = ds_local
+    df = df.drop(df.obj)
+    filename = tempfile.mktemp(suffix='.csv')
```
maartenbreddels, line 46: Please use the pytest tmpdir fixture, you can find examples in this source file

maartenbreddels:
Yes, I opened that issue here: #661

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: Ah yeah, indeed. I just copied the stuff from the other `export_` methods. Can you please check that the explanation in the doc snippets and say `export_hdf5` is consistent and has not changed recently? (just in case).

tests/export_test.py
JovanVeljanoski: Done!
(I saw there are tests with and without this.. should we eventually switch to tmpdir for everything?)

maartenbreddels: I think in most cases we should.

packages/vaex-core/vaex/dataframe.py
maartenbreddels: I don't understand what you are saying, but also export_hdf5 should have used that snippet.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5747,6 +5753,41 @@ def export_fits(self, path, column_names=None, shuffle=False, selection=False, p
...
 
+    @docsubst
+    def export_csv(self, path, virtual=False, selection=False, progress=None, batch_size=1_000_000, **kwargs):
+        """ Exports the DataFrame to a CSV file.
+
+        :param str path: Path for filename
+        :param bool virtual: If True, export virtual columns as well
+        :param bool selection: {selection1}
+        :param progress: {progress}
+        :param int batch_size: {chunk_size_export}
```
maartenbreddels, line 5764:
```suggestion
        :param int chunk_size: {chunk_size_export}
```

maartenbreddels: In vaex-core we use chunk_size everywhere, it;'s also consistent wording with arrow.

maartenbreddels:
When green, please squash and merge (so it is 1 commit)

maartenbreddels:
Ok, I was a bit too fast, we first had some concat-df issues to solve in #688

maartenbreddels:
I think this is good to be, please squash and merge.

PR #697, opened 2020-04-25
Chore run benchmarks (WIP)

byaminov:
Configure GitHub actions workflow to run the ASV benchmarks and commit the results to `gh-pages`.


byaminov:
This approach did not work out for two reasons:
* The new workflow was running on the `self-hosted` runner even though it was created from a forked repository, but a contributor without write access on the base Vaex repo (me). So this is a security hole to run arbitrary code on the self-hosted hardware.
* The script of the runner had to be restarted once in a while; once decided to upgrade itself and was handing for hours until I `kill -9`ed it. So not sure how mature this service of GitHub is yet.

Abandoning this PR in favor of #701.

PR #709, opened 2020-04-30
chore: automate releases using GHA

maartenbreddels:
exploring if we can do nighly releases


maartenbreddels:
Instead of nighlies, I decided we just automate the releases, if we tag, we release. Better spend energy on making releases easily then have nightlies *and* doing releases.

We can test with pushing to test pypi, but I suggest we make a release_test branch, which then pushes to test pypi.

JovanVeljanoski:
I agree with this strategy.

PR #714, opened 2020-05-01
chore: setup mybinder + doit script to keep it up to date

maartenbreddels:


maartenbreddels:
By running doit we can keep the requirements file up to date it we have to.
I plan to automate more using this doit

maartenbreddels:
I made the mybinder point to the latest branch, which now is equal to the meta-v2.6.0 tag (the latest release).

PR #715, opened 2020-05-02
Fix: limits calculation with selection

### LINKED ISSUE #44: Calculating limits with selection
Hi,

There seems to be a problem when calculating the limits of a column while having specified a selection. 
For example:
`dataset.limits('param1', value='99%', selection='sel')`
fully ignores the selection. 
However
`dataset.limits('param1', value='minmax', selection='sel')`
works as expected with the specified selection.

Thanks, 
Jovan.

JovanVeljanoski:
This fixes #44. 

_Description:_
The selection seems to be ignored when calculating limits that are not `minmax`. For `minmax` the selection is obeyed, as shown by already implemented tests, and via manual testing.
```
limits = df.limits('x', value='90%', sel='x>5')  # selection is ignored
```
However, calculating limits on a filtered DataFrame gives the correct result:
```
df2 = df[df.x > 5]
limits = df2.limit('x', value='90%')  # give the correct result
```
_Checklist:_
- [x] Problem exposed via a unit-test
- [ ] Unit-test passes

### DIFF
--- a/tests/limits_test.py
+++ b/tests/limits_test.py
@@ -27,3 +27,12 @@ def test_limits(df):
 
     assert df.limits(['x', 'y'], 'minmax', shape=10)[1] == [10, 10]
     assert df.limits(['x', 'y'], 'minmax', shape=(10, 12))[1] == [10, 12]
+
+
+def test_limits_with_selection(df):
+    limits_selection_perc = df.limits('x', value='90%', selection='x > 5')
+
+    df_sliced = df[df.x > 5]
+    limits_sliced = df_sliced.limits('x', value='90%')
+
+    assert limits_sliced.tolist() == limits_selection_perc.tolist()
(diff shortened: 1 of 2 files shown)


tests/limits_test.py
```diff
@@ -27,3 +27,12 @@ def test_limits(df):
...
     assert df.limits(['x', 'y'], 'minmax', shape=(10, 12))[1] == [10, 12]
+
+
+def test_limits_with_selection(df):
+    limits_selection_perc = df.limits('x', value='90%', selection='x > 5')
+
+    df_sliced = df[df.x > 5]
+    limits_sliced = df_sliced.limits('x', value='90%')
+
+    np.testing.assert_almost_equal(limits_sliced, limits_selection_perc, decimal=3)
```
maartenbreddels, line 38: no reason for the almost right? (I'll fix it)

maartenbreddels: In any case, better tested with minmax, in case the percentile algo changes.

maartenbreddels: Ok, minmax did work!

Issue #606, opened 2020-02-21
importing vaex package with torch causes segmentation fault


jpujari-ctx:
We are experiencing an issue when importing the vaex package when torch is installed

```
root@73c8737039e5:/# python3.7
Python 3.7.6 (default, Dec 19 2019, 23:49:42)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> import vaex
Segmentation fault
root@73c8737039e5:/#
```

No error when just importing vaex
```
root@73c8737039e5:/# python3.7
Python 3.7.6 (default, Dec 19 2019, 23:49:42)
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import vaex
INFO:MainThread:matplotlib.font_manager:generated new fontManager
>>>
```

Error log:
```
root@a30a347a9c71:/# python3.7 -q -X faulthandler
>>> import torch
>>> import vaex
Fatal Python error: Segmentation fault
Current thread 0x00007f520a7b3700 (most recent call first):
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1043 in create_module
  File "<frozen importlib._bootstrap>", line 583 in module_from_spec
  File "<frozen importlib._bootstrap>", line 670 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 967 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 983 in _find_and_load
  File "/usr/local/lib/python3.7/dist-packages/vaex/strings.py", line 7 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 728 in exec_module
  File "<frozen importlib._bootstrap>", line 677 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 967 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 983 in _find_and_load
  File "/usr/local/lib/python3.7/dist-packages/vaex/column.py", line 9 in <module>
  File "<frozen importlib._bootstrap>", line 219 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 728 in exec_module
  File "<frozen importlib._bootstrap>", line 677 in _load_unlocked
[...]

PR #717, opened 2020-05-05
chore(core): fix compilation issue with gcc5

maartenbreddels:
As mentioned here: https://github.com/vaexio/vaex/issues/606#issuecomment-590407455

this should fix compilation with gcc5, this was still in my git stash. (cc @nugend )


maartenbreddels:
failure was unrelated.

PR #722, opened 2020-05-06
Docs for vaex-jupyter

maartenbreddels:
@JovanVeljanoski would be great if you can go through the tutorial

Maybe we want to refer to these images instead of putting them in the repo:
example_jupyter_ipyvolume
example_jupyter_plotly


JovanVeljanoski:
Few comments on the state of this so far:
 - If you go to the bottom of the vaex-jupyter tutorial, in the build your own viz section, the screenshots for both plotly and ipyvolume are repeating. The code seems to be correct, i.e. the notebook markdown cell is showing the correct paths to the image/source. If I remove one of them, the correct link/screenshot is shown, but not when both links are present, even in differen cells.

- I've re-structured to docs such that we have a Gallery page, which essentially shows the vaex-jupyter examples for now. These examples are also accessible from the Examples page.

PR #726, opened 2020-05-07
counts bug if np.any(nan_mask) in count_values()

And0k:
just strange code noticed


maartenbreddels:
Yeah, I discussed this with Jovan, and it looks wrong indeed. I am need to look more into this. My linter also noticed that mask is an undefined variable, meaning that that code path never ran.
Also, I think it cannot ever reach there, so I don't think it is a bug. But I have to go through the git logs to see what is going on.

Good catch, and thanks for letting us know, will dig into this later.

PR #730, opened 2020-05-08
Performance: groupby faster due to category and reusing count(*) 🔥

maartenbreddels:
Replaces #345, should run the benchmarks on asv because of the branch name

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Create env
       # if: steps.cache-conda-env.outputs.cache-hit != 'true'
       run: |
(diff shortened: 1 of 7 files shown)


maartenbreddels:
I hope you agree with https://github.com/vaexio/vaex/pull/730/commits/7913d0af3f5b55d3be937c2b08f00dc0b6df3c18

maartenbreddels:
Seeing a 2x speedup locally, from the category use, and 2x from not calculating count(*) twice. We may see 4x speedups 🔥

JovanVeljanoski:
> I hope you agree with [7913d0a](https://github.com/vaexio/vaex/commit/7913d0af3f5b55d3be937c2b08f00dc0b6df3c18) @JovanVeljanoski

I agree!

Issue #732, opened 2020-05-09
Assigning an join() expression to a non-current dataframe fails. And wrong where() output if it selects NA values


And0k:
```python
import vaex, pandas as pd
ds = vaex.from_arrays(a=['a', 'a', 'c', 'd', 'e', 'f', 'f'], row_index=range(7))
Out[4]: 
  #  a      row_index
  0  a              0
  1  a              1
  2  c              2
  3  d              3
  4  e              4
  5  f              5
  6  f              6
df = pd.DataFrame({'a': ['A', 'B', 'F', 'G']}, index=[0, 1, 5, 6])
dsj = ds.join(vaex.from_pandas(df), left_on='row_index', right_on='index', rsuffix='u')
dsj
Out[5]: 
  #  a      row_index  au    index
  0  a              0  A     0
  1  a              1  B     1
  2  c              2  --    --
  3  d              3  --    --
  4  e              4  --    --
  5  f              5  F     5
  6  f              6  G     6

# Replacing a with NA where it is in au:
combined = dsj['au'].isna().where(dsj['au'], dsj['a'])
combined
Out[6]: 
Expression = where(isna(au), au, a)
Length: 7 dtype: object (expression)
------------------------------------
0  a
1  a
2  G
3  G
4  G
5  f
6  f
```
First issue: why "G" in values?

Bug 2 - on assignment:
```python
ds['combined'] = combined
ds
Out[37]: ERROR:MainThread:vaex:error evaluating: combined at rows 0-7
Traceback (most recent call last):
  File "C:\Users\and0k\anaconda3\envs\py3.7x64h5togrid_no_pip1\lib\site-packages\vaex\scopes.py", line 97, in evaluate
    result = self[expression]
  File "C:\Users\and0k\anaconda3\envs\py3.7x64h5togrid_no_pip1\lib\site-packages\vaex\scopes.py", line 144, in __getitem__
    raise KeyError("Unknown variables or column: %r" % (variable,))
KeyError: "Unknown variables or column: 'where(isna(au), au, a)'"
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "C:\Users\and0k\anaconda3\envs\py3.7x64h5togrid_no_pip1\lib\site-packages\vaex\dataframe.py", line 2031, in dtype
    data = self.evaluate(expression, 0, 1, filtered=False, internal=True, parallel=False)
[...]

Issue #725, opened 2020-05-07
bug: deadlock in execute_test.py


maartenbreddels:
Seeing this in CI on windows, e.g.:
https://ci.appveyor.com/project/maartenbreddels/vaex-4gh4b/builds/32693435/job/9qywm58pj5xpljym

I think the issue is in `test_thread_safe`.

maartenbreddels:
It seems to be only affecting windows.

PR #736, opened 2020-05-11
Fix: Left join when the right dataframe is filtered

JovanVeljanoski:
Adds a test that exposes a bug with left join when the right dataframe is filtered. 
Originally reported in #732 

- [x] Add a test
- [x] Make the test pass


maartenbreddels:
The problem appeared to be related to the ColumnIndexed, join was fine, although it did expose the issue.

maartenbreddels:
The failure on travis should be fixed by #739, it seems the deadlock (#725) can occur on Linux as well (failure on GHA).
Both failures are unrelated though.

Issue #748, opened 2020-05-18
CI: xgboost fails to import


maartenbreddels:
https://github.com/vaexio/vaex/runs/683835668?check_suite_focus=true

My guess is that there has recently been an xgboost release.

maartenbreddels:
Note, this failure was triggered by the nightly CI run: https://github.com/vaexio/vaex/actions

JovanVeljanoski:
Hopefully fixed via #749

PR #749, opened 2020-05-18
Fix for the CI

JovanVeljanoski:
Yet another PR to fix the CI. Currently xgboost is giving trouble on mac.

Addresses #748

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Create env
       # if: steps.cache-conda-env.outputs.cache-hit != 'true'
       run: |
-        conda create -q -n vaex -c conda-forge python=${{ matrix.python-version }} numpy scipy xarray pyqt matplotlib pyopengl h5py astropy tornado cython pandas pytest numba pyarrow graphviz python-graphviz pcre catboost plotly notebook scikit-learn lightgbm py-xgboost
+        conda create -q -n vaex -c conda-forge python=${{ matrix.python-version }} numpy scipy xarray pyqt matplotlib pyopengl h5py astropy tornado cython pandas pytest numba graphviz python-graphviz pcre catboost plotly notebook scikit-learn lightgbm py-xgboost
     # - name: Restoring cached env
     #   if: steps.cache-conda-env.outputs.cache-hit == 'true'
     #   run: |
@@ -65,6 +65,10 @@ jobs:
       run: |
         conda init --all
         echo ::add-path::$CONDA/envs/vaex/bin
+    - name: Install OpenMP runtime (Mac-only)
+      if: matrix.os == 'macOS-latest'
+      run: |
+        brew install libomp
     - name: Fix python-dateutil/botocore version conflict
       if: matrix.os != 'windows-latest'
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/pythonpackage.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Create env
       # if: steps.cache-conda-env.outputs.cache-hit != 'true'
       run: |
-        conda create -q -n vaex -c conda-forge python=${{ matrix.python-version }} numpy scipy xarray pyqt matplotlib pyopengl h5py astropy tornado cython pandas pytest numba pyarrow graphviz python-graphviz pcre catboost plotly notebook scikit-learn lightgbm py-xgboost
```
xhochy, line 49: You should be able to revert that again, everything should be fixed now.

.travis.yml
```diff
@@ -36,12 +36,16 @@ before_install:
...
   - conda info -a
   # future: kapteyn should only be needed for astro package testing
-  - conda create -q -n test-environment -c conda-forge python=$PYTHON_VERSION setuptools==42.0.2 certifi=2019.6.16 pip numpy scipy xarray pyqt matplotlib pyopengl h5py astropy tornado cython pandas pytest pytest-asyncio numba pyarrow graphviz python-graphviz pcre catboost libcxx=9.0.0 notebook scikit-learn lightgbm py-xgboost # $OPTIONAL_PACKAGES
+  - conda create -q -n test-environment -c conda-forge python=$PYTHON_VERSION setuptools==42.0.2 certifi=2019.6.16 pip numpy scipy xarray pyqt matplotlib pyopengl h5py astropy tornado cython pandas pytest pytest-asyncio numba graphviz python-graphviz pcre catboost libcxx=9.0.0 notebook scikit-learn lightgbm py-xgboost # $OPTIONAL_PACKAGES
   - which pip
   - source activate test-environment
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       conda install gxx_linux-64 -y;
     fi
+  # Install OpenMP on osx with brew
```
xhochy, line 45: You can get that via `conda` as well, it is named `llvm-openmp`

maartenbreddels: Cool, that is much cleaner, how should we have known this btw? We tried 'omp' I believe, but was kind of a mystery.

xhochy: Checking in the jungle of the conda-forge documentation: https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp

Also it seems like you are only using the C-F compiler on Linux, might be that `llvm-openmp` is not compatible with the system compilers. Generally, you should use conda-forge packages only with the conda-forge provided compiler (you can get them either via the `gxx_linux-64` names or use the convenience package `compilers` which will get the correct ones platform-independently.

JovanVeljanoski: Thanks, we are trying of openmp via conda-forge now as you suggested. After that I will try pyarrow via conda-forge as well.

maartenbreddels: So we can replace `conda install gxx_linux-64 -y;` by `conda install -c conda-forge compilers -y;` for all OS'es?

xhochy: Yes, this will have the disadvantage that it will also install a fortran compiler but I think this is worth the comfort.

JovanVeljanoski: Cool, here we go again :)

maartenbreddels: Not sure why that led to a compilation issue with annoy :|

maartenbreddels:
Both CI issues are unrelated, I suggest we merge this first.

PR #757, opened 2020-05-22
fix(arrow): booleans were negated (fixes #562) and offsets ignored

### LINKED ISSUE #562: Bool values get flipped when converting Arrow table to DataFrame
Using the latest version:
`vaex==2.6.1`

Just realised that when converting an Arrow table to a DataFrame, bool columns get flipped and converted to integers:

```python
import vaex
from pyarrow import feather

bool_array = [False, True, True, False]
pdf = pd.DataFrame({"col1": bool_array})
pdf.to_feather("test_data.feather")
arrow_table = feather.read_table("test_data.feather")
vaex.from_arrow_table(arrow_table)
```

```
#  | col1
-- | --
0  | 1
1  | 0
2  | 0
3  | 1
```

maartenbreddels:


maartenbreddels:
https://github.com/vaexio/vaex/pull/757/checks?check_run_id=699083628 seems like a numba+windows new failure, never see that before. Unrelated, and lets see if disappears with a new release of numba. (also some vague network error, which I am not sure is related)

PR #760, opened 2020-05-25
chore(ci): try releasing wheel for windows

maartenbreddels:


maartenbreddels:
It looks like there are two runtimes installed:
```
14.16.27012
14.25.28508
```
Together with the GHA `ilammy/msvc-dev-cmd@v1` used in #627 we may be able to get rid of vcruntime140_1.dll

PR #762, opened 2020-05-25
Docs update

JovanVeljanoski:
Remove Dask stand-alone example which is now part of the I/O example notebook.


JovanVeljanoski:
This PR will be superseded by an upcoming docs update, so the changes done here are not relevant any more.

Issue #790, opened 2020-06-03
NameError using export_csv


BssMsi:
Hi,
I'm getting NameError, not sure why.
I have columns "Sub-Region" and "Global-Region", which I'm guessing is implicitly aliased to "Sub_Region", since I can access the column in both ways.
Even if I just read the file and export it in the next line, I get the error, that is,
oneDF = vaex.from_csv(inputFile)
oneDF.export_csv("test.csv")

NameError: Column or variable 'Region' does not exist. Did you mean: 'Sub-Region' or 'Global-Region'

My guess would be it might be some issue handling the hyphenated column while exporting.

Thanks

JovanVeljanoski:
Thank you for the report, we'll try to look into this soon.

maartenbreddels:
This should be fixed when #793 is merged/released, but I do recommend not using csv. Use hdf5, arrow or parquet, especially when exporting large amounts of data.

BssMsi:
Oh, that might be the issue, I had pip installed it, I'll clone the repo and set it up again and check.

Thanks

maartenbreddels:
Vaex 2.0.2 is almost released https://github.com/vaexio/vaex/actions/runs/124999191 :)

maartenbreddels:
This is released now, you can try it out with `$ pip install "vaex-core>=2.0.2"`

BssMsi:
Yeah, it's working now, thanks
Excited to get started with vaex

Cheers

maartenbreddels:
Great, thanks for letting us know, good luck!

PR #791, opened 2020-06-03
Fix: export_csv in presence of aliased columns

JovanVeljanoski:
This PR addresses #790, which exposed an issue with exporting data to csv when aliased columns are present in the dataframe. 

- [x] Write a test
- [x] Make test pass

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5817,14 +5817,15 @@ class DataFrameLocal(DataFrame):
         """
         import pandas as pd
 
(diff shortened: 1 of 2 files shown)


JovanVeljanoski:
CI issues seem unrelated, although I can't see why the macOS build is failing on GitHub actions

maartenbreddels:
Thanks, I checked if https://github.com/vaexio/vaex/pull/793 fixes it, and it does. Btw, not that we have a new column called '123456' in the general `df` fixture, so I've rewritten the csv test to just export that df, and test the presence when reading back.

Issue #776, opened 2020-05-31
Vaex fails to convert pandas df to vaex df using `from_pandas()`


voxmenthe:
vaex version: 3.0.0
numpy version: 1.17.3
pandas version: 0.23.4
python version: 3.7.4

Issue: vaex fails to convert a simple pandas dataframe to vaex

# Code example to reproduce:

pandas_df = pd.DataFrame({'x': [0,1,2,3,4], 'y': [0,1,4,9,16]})
vaex_df = vaex.from_pandas(pandas_df)

Fails with: 
AttributeError: module 'pandas.core.arrays' has no attribute 'integer'

Full error message:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-110-9fe4b9718db8> in <module>
      1 #doesn't work in vaex version 3.0.0
----> 2 vaex_df = vaex.from_pandas(pandas_df)

/.venv/lib/python3.7/site-packages/vaex/__init__.py in from_pandas(df, name, copy_index, index_name)
    400                 print("Giving up column %s, error: %r" % (name, e))
    401     for name in df.columns:
--> 402         add(name, df[name])
    403     if copy_index:
    404         add(index_name, df.index)

/.venv/lib/python3.7/site-packages/vaex/__init__.py in add(name, column)
    388     def add(name, column):
    389         values = column.values
--> 390         if isinstance(values, pd.core.arrays.integer.IntegerArray):
    391             values = np.ma.array(values._data, mask=values._mask)
    392         try:

AttributeError: module 'pandas.core.arrays' has no attribute 'integer'

JovanVeljanoski:
Hi, 

I believe this is caused by an issue on the pandas side, which has now been fixed. Can you please update to a more recent version of pandas?

@maartenbreddels we should probably require a more recent pandas version during install of vaex-core?

Issue #766, opened 2020-05-27
vaex.from_pandas fails with "AttributeError: module 'pandas.core.arrays' has no attribute 'integer'"


chaltik:
`(py3.6-tsse) wtisim@ip-10-0-0-38:~/wtisim$ ipython
Python 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 17:14:51)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import vaex

In [2]: import pandas as pd

In [3]: import numpy as np

In [4]: test_df = pd.DataFrame(np.random.randn(10000000,10))

In [5]: vaex_df = vaex.from_pandas(test_df)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-4d9420c13fea> in <module>()
----> 1 vaex_df = vaex.from_pandas(test_df)

~/wtisim/.venv/py3.6-tsse/lib/python3.6/site-packages/vaex/__init__.py in from_pandas(df, name, copy_index, index_name)
    400                 print("Giving up column %s, error: %r" % (name, e))
    401     for name in df.columns:
--> 402         add(name, df[name])
    403     if copy_index:
    404         add(index_name, df.index)

~/wtisim/.venv/py3.6-tsse/lib/python3.6/site-packages/vaex/__init__.py in add(name, column)
    388     def add(name, column):
    389         values = column.values
--> 390         if isinstance(values, pd.core.arrays.integer.IntegerArray):
    391             values = np.ma.array(values._data, mask=values._mask)
    392         try:

AttributeError: module 'pandas.core.arrays' has no attribute 'integer'

In [7]: pd.__version__
Out[7]: '0.23.3'

In [8]: np.__version__
Out[8]: '1.18.4'

In [9]: vaex.__version__
Out[9]:
{'vaex': '3.0.0',
 'vaex-core': '2.0.0',
 'vaex-viz': '0.4.0',
 'vaex-hdf5': '0.6.0',
 'vaex-server': '0.3.0',
 'vaex-astro': '0.7.0',
 'vaex-jupyter': '0.5.0',
 'vaex-ml': '0.9.0',
 'vaex-arrow': '0.5.0'}
`

JovanVeljanoski:
Hi, 

This is actually the same issue as in #608 

So in your example, if you "name" the column in the pandas dataframe, things should go smoothly:

```
[...]

PR #794, opened 2020-06-04
fix: support from_pandas_df with pandas 0.23

maartenbreddels:
Closes https://github.com/vaexio/vaex/issues/776
Closes https://github.com/vaexio/vaex/issues/766


maartenbreddels:
Note that our test suite doesn't pass with pandas 0.23, but I think we can do our best to support older versions.

PR #808, opened 2020-06-05
Fix join when there are None values in left index

JovanVeljanoski:
This PR addresses the following problems with join:
 - If the index column on the left dataframe contains `None` (and the right does not in this case), those rows on the right will be filled with the 1st group, rather than being themselves `None` (or masked), since there is no match.
- The same issue repeats when one is using `how=inner` with the same configuration as the previous point. 
- If one tries to "fix" this by `df_left['index'] = df_left['index'].fillna(value='missing_value')`, than join crashes for both left and inner. 

See screenshot below for a better explanation:
image

Expectation is that the last element in columns c, z1, z2 should be masked.

checklist:
- [x] Implement tests
- [ ] Make tests pass

### DIFF
--- a/tests/join_test.py
+++ b/tests/join_test.py
@@ -332,3 +332,29 @@ def test_join_f_c_inner_none():
     assert df.w2.tolist() == [True, False]
     assert df.z1.tolist() == [-1.0, -2.0]
     assert df.z2.tolist() == [True, False]
+
+
+def test_join_f_c_left_none_fillna():
+    df_f_copy = df_f.copy()
+    df_f_copy['f'] = df_f_copy.f.fillna(value='missing')
+    df = df_f_copy.join(df_c, left_on='f', right_on='c', how='left')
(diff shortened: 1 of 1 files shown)


tests/join_test.py
```diff
@@ -305,3 +310,51 @@ def test_with_masked_no_short_circuit():
...
+    assert df.w2.tolist() == [True, False]
+    assert df.z1.tolist() == [-1.0, -2.0]
+    assert df.z2.tolist() == [True, False]
+
+
+def test_join_f_c_left_none_fillna():
+    df_f_copy = df_f.copy()
+    df_f_copy['f'] = df_f_copy.f.fillna(value='missing')
+    df = df_f_copy.join(df_c, left_on='f', right_on='c', how='left')
+    assert df.shape == (2, 6)
```
maartenbreddels, line 341: this will be (3, 6), my guess is that you wanted to do a different join?

PR #814, opened 2020-06-08
cardinality implementation

xdssio:
implement a cardinality measure.
Simply put: 
```len(df.x.unique())/len(df)```

Notes:
Cardinality By definition is just the count of different elements, so maybe the naming is not optimal.

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -964,6 +964,14 @@ def f({0}):
     def is_masked(self):
         return self.ds.is_masked(self.expression)
 
+    def cardinality(self):
+        """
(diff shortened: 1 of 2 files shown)


maartenbreddels:
This has crossed my mind before, and something I thought about was `def cardinality_approx(self, count=None, assume_shuffled=None, relative_error=0.1):`

That (if we keep track of the metadata that a dataframe (or column!) is shuffled, could take the `count` first elements. If not, it would first do `df.sample(count)`. 

Count could be determined using poison statistics I guess. Basically, if you have a billion rows, shuffled, how rows would you need to do this on before you have a max error of 0.1.

@JovanVeljanoski : any thoughts on this?

PR #819, opened 2020-06-09
Feature: batch training for Catboost

xdssio:
Added four new params to *CatBoostModel*.
* **batch_size** - If provided, train in batches of that size.
* **batch_weights** - If provided and *batch_size* is provided, used to sum the models trained in batches and their feature importance. 
* **evals_result_** - Saving evals_result_ for all batches during training as the [sum_models](https://catboost.ai/docs/concepts/python-reference_sum_models.html) ignores it.

Implemented *get_feature_importance* to match the [Catboost method](https://catboost.ai/docs/concepts/python-reference_catboost_get_feature_importance.html).

Added feature_names to the data Pool such that the Catboost booster acknowledges the feature names.

Note - 
I designed it to run differently on a single run vs batch instead of a "cleaner" single batch version if batch_size isn't provided because the *sum_models* method loses all kind of information which we might want keep if we can.

### DIFF
--- a/packages/vaex-ml/vaex/ml/catboost.py
+++ b/packages/vaex-ml/vaex/ml/catboost.py
@@ -133,9 +133,14 @@ class CatBoostModel(state.HasState):
             models = []
             batch_weights = self.batch_weights if 0 < len(self.batch_weights) else None
 
+            # Set up progressbar
+            n_samples = len(df)
+            progressbar = vaex.utils.progressbars(progress)
+
             column_names = self.features + [self.target]
             iterator = df[column_names].to_pandas_df(chunk_size=self.batch_size)
             for i1, i2, chunk in iterator:
+                progressbar(i1 / n_samples)
                 data = chunk[self.features].values
                 target_data = chunk[self.target].values
                 dtrain = catboost.Pool(data=data, label=target_data, **self.pool_params)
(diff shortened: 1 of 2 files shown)


maartenbreddels:
As @JovanVeljanoski has done the sklearn wrapping, using the evaluate_iterator (https://github.com/vaexio/vaex/blob/0e1326b84f4baa7decb1695b89423db7bb5f060c/packages/vaex-ml/vaex/ml/sklearn.py#L225), I think it's best he reviews it.

JovanVeljanoski:
Hi, so I just pushed some changes. Most of it is just a clean-up. 
Here is a summary of the important stuff:
 - removed `get_feature_importance()` from the wrapper. It can be accessed from the underlying `booster` object which is the catboost model. We do not want to maintain this going forward.
- Given the [this explanation](https://github.com/catboost/catboost/issues/1344#issuecomment-653409209), a sensible default when the `batch_weighs` are not provided is to use `[1/num_models] * num_models`. 
- If the list of model weights is different than the list of models, be unforgiving: raise an explicit error.
- Add a progressbar to monitor the status of the batch training.

I still need to to a internal test on the big dataset, if that goes well, we can give this to @maartenbreddels for the final code review.

maartenbreddels:
Thanks Y&J!

Issue #836, opened 2020-06-16
df.head error after join


alrz199:
df = df1.join(df2,how='left',on='col-name',allow_duplication=True)
then when i use df.head(3)  i see this error:
array is of length 5012, while the length of the DataFrame is 3
df1 and df2 are csv files and i created them with pandas after clustering algorithms. also df2  has null values but df2.head(3) works (and also works for df1) and after joining i got the error. i tried df.fillna(0).head(3) but still see that error.

maartenbreddels:
Hi,

thanks for the report. It would help us a lot if you can provide a reproducible example!

Regards,

Maarten

alrz199:
thank you for your quick reply
example:
import pandas as pd 
import numpy as np
import vaex
df_test1 = pd.DataFrame(np.array([['a1', 2.0],
                                  ['a2', 1.0],
                                  ['a3', 3.0],
                                  ['a4', 5.0],
                                  ['a5', 6.0]]),columns =['a','b'])
df_test2 = pd.DataFrame(np.array([['a1',3],
                                  ['a2', 4],
                                  ['a2', 6],
                                  ['a1', None],
                                  ['a2', 6]]),columns =['a','c'])
df_test1.to_csv('test1.csv',header=True,index=False)
df_test2.to_csv('test2.csv',header=True,index=False)
df1 = vaex.from_csv('test1.csv', copy_index=False)
df2 = vaex.from_csv('test2.csv', copy_index=False)
df1.head():
#	a	b
0	a1	2
1	a2	1
2	a3	3
3	a4	5
4	a5	6
df2.head():
#	a	c
0	a1	3
1	a2	4
2	a2	6
3	a1	nan
4	a2	6
df = df1.join(df2,on='a',how='left',allow_duplication=True)
df.head(3):
....ValueError: array is of length 5, while the length of the DataFrame is 3

JovanVeljanoski:
Hi @alrz199 
Thanks for the reproducible example. I can also reproduce it and have already created a unit-test. We hope we can fix this soon. 

Thanks!

maartenbreddels:
Indeed, thanks for reporting this, it allowed us to find and fix a bug!

PR #839, opened 2020-06-16
Fix `df.head()` after join

JovanVeljanoski:
This PR fixes `df.head()` for a dataframe that is a result of the join operation.
Reported in #836 

checklist:
- [x] make unit-test
- [ ] make unit-test pass
- [ ] manual testing


maartenbreddels:
I don't think this has sth to do with join, but with concatenated dataframes.

PR #860, opened 2020-06-26
fix: concatenated columns trimmed to smaller then the length of the f…

### LINKED ISSUE #836: df.head error after join

maartenbreddels:
…irst chunk failed

Replaces #839
Closes #839 Closes #836


maartenbreddels:
Failure unrelated

Issue #864, opened 2020-06-26
from_sql and to_sql would be really helpful features in vaex


vishalhkd007:
Vaex has a huge advantage over Pandas when it comes to loading data from files into dataframe. With support for data read and write to RDBMS, Vaex can become a powerful tool for ETL processing where we currently are using Pandas (which is slow for large datasets).

The advantage of Pandas is its support for the database systems. I can create a dataframe from an SQL query as well as stream data to an RDBMS that creates a database table from the Pandas dataframe. This is quite helpful when it comes to using Pandas for ETL where processing of data including dealing with missing data, flagging incorrect data etc is done in Pandas and then streamed to database.

Vaex for its performance gains for large dataset over Pandas would really gain from basic database support making it the tool for such processing.

maartenbreddels:
Hi,

 I was thinking about it. And I'm open to the idea, possible via https://turbodbc.readthedocs.io/en/latest/ ->arrow -> vaex and possibly via arrow in the future directly.
What kind of usage do you see, a batch downloading to a arrow/hdf5 file after which you work with the data, or a 'lazy' fetch from the database, where no data is kept in memory?

cheers,

Maarten

vishalhkd007:
Thanks for replying on this issue.



The usage mentioned is one that is quite common for data transfers and ETL processing from legacy systems to RDBMS or NoSQL systems. We deal with a lot of clients who are trying to have this kind of data migration where data is dumped to CSV files and from there, the data needs to be processed and analyzed for any discrepancies and missing data. 



For small datasets, this can be done with Pandas. But our clients have data in sizes over 100GB+. In these situations, loading data in-memory for analysis is not an option. One way to deal with these is to split the data into smaller files (but many operations that we do on the data do not work with such splits). Vaex shows a lot of potential here where the lazy loading from arrow/hdf5 files works and allows for the ETL analysis.



Another use case is data consolidation - where we are dealing with data acquired from different sources (either through mergers or where our clients bring in external data sources to be mashed with what they already have). Now, we are dealing with multiple datasets which need to be analyzed for commonality along with the need for joins and merges. Again, based on the size of the data, this can be an extremely slow and painful in Pandas. I am hopeful that Vaex will be beneficial - in fact, I have been successful in creating a mash of over 5M records from 2 datasets using Vaex where it outshined Pandas by a huge margin.



[...]

PR #865, opened 2020-06-26
refactor: separate dataframe and dataset

maartenbreddels:
We're back to Dataset again 😄. An immutable mapping of name to column.

The idea is to abstract away the data part of a DataFrame, which gives us a few advantages.
 * We can have lazy loading of data, and reading of multiple columns in 1 go
   * e.g. SQL https://github.com/vaexio/vaex/issues/864
   * Parquet via Arrow Dataset API
 * Dataset and thus DataFrames can be serialized (e.g. only store the hdf5 or arrow path, not serialize the data), so we can efficiently pickle/serialize/transport them with use of Dask/Ray/Mars.

This also starts with the idea of identifying data with a hash key to quickly compare data, which makes caching easier (e.g. also when using Dask), or when we want to cache complex operations (groupby). 

We will not have subclasses of DataFrame anymore (except for DataFrameLocal and DataFrameRemote), all data specific 
parts are done in the Dataset.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -60,10 +60,10 @@ jobs:
         brew install libomp
     - name: Cache compiled binaries
       id: cache-compiled-binaries
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: packages/vaex-core/build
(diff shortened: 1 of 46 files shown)


maartenbreddels:
I can get it to work with both ray and dask, but dask crashes when doing distributed, not sure why. I think we want to ray/dask stuff on a different PR, but for now i keep this together. 2 commits in 1 PR would also be fine I think.

PR #867, opened 2020-06-26
fix(ml): weight of evidence encoder can handle edge cases

JovanVeljanoski:
This PR fixes an issue with the `WeightOfEvidence` encoder, that happens when a feature has no positive targets (binary classification case). 

Note: the case when a feature has no negative targets was already handled in the initial implementation. The same approach is used here for the other edge case.

 - [x] test covering these edge cases
 - [x] test passes


JovanVeljanoski:
Issues in Travis/Appveyor are unrelated.

PR #870, opened 2020-06-27
Fix: minmax method for different dtypes

JovanVeljanoski:
This PR enables the `df.minmax()` method to be applied on multiple expressions of different numeric types (e.g. int, float) at the same time.

Checklist:
 - [x] test
 - [ ] make test pass

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -415,10 +415,16 @@ class MinMaxScaler(Transformer):
         :param df: A vaex DataFrame.
         '''
 
-        assert len(self.feature_range) == 2, 'feature_range must have 2 elements only'
-        minmax = df.minmax(self.features)
-        self.fmin_ = minmax[:, 0].tolist()
-        self.fmax_ = minmax[:, 1].tolist()
+        fmin = df.min(self.features, delay=True)
+        fmax = df.max(self.features, delay=True)
+
+        @vaex.delayed
+        def assign(fmin, fmax):
+            self.fmin_ = fmin.tolist()
+            self.fmax_ = fmax.tolist()
+
+        assign(fmin, fmax)
+        df.execute()
 
     def transform(self, df):
         '''
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -415,15 +415,16 @@ class MinMaxScaler(Transformer):
         :param df: A vaex DataFrame.
         '''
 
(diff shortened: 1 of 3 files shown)


maartenbreddels:
I don't think we should support this. The idea is that you get back a single numpy array, so they cannot be of mixed type. If you agree, please close.

JovanVeljanoski:
Yeah, i understand what you are saying. 
Still, nothing stops users from trying to get minmax of columns of different types. So i propose either:
 - we explicitly catch this and throw a nice looking informative error
 - cast the final array in the "common denominator" type (i.e. float for everything). 

This is kind of needed for MinMaxScaler to work in one go with various columns. Not hard requirement, but the error message we get now is a bit misleading.  What do you think?

maartenbreddels:
I think that's a misuse of minmax, we should do minmax per feature, with delay=True, then call df.execute().
A better error message would be good!

JovanVeljanoski:
Refactored `MinMaxScaler` per your suggestions. 

Before merging this, we can simply add a more informative error message if somebody tries to use `df.minmax(...)` with columns of different types?

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -415,10 +415,16 @@ def fit(self, df):
         :param df: A vaex DataFrame.
         '''
 
-        assert len(self.feature_range) == 2, 'feature_range must have 2 elements only'
-        minmax = df.minmax(self.features)
-        self.fmin_ = minmax[:, 0].tolist()
-        self.fmax_ = minmax[:, 1].tolist()
+        fmin = df.min(self.features, delay=True)
+        fmax = df.max(self.features, delay=True)
```
maartenbreddels, line 419: Hmm, i am a bit worried that this passes the tests. This means it's being cast to a common type.

maartenbreddels: I think we should loop over each feature, we can still call minmax right? (which is faster than a call to min, and to max)

JovanVeljanoski: Both options would use 1 pass over the data right? Why would it be faster?
I can try to make a loop if you want, or if you have time/will you can also implement it.

maartenbreddels: There is a difference right. If I do a df.min(['somefloat', 'someint']), it will come back as float array, we don't want that.

JovanVeljanoski: Done. Now each feature preserves the dtype of its min/max values in MinMaxScaler

PR #871, opened 2020-06-27
Add unit-test for dropinf

JovanVeljanoski:
I have forgotten to commit the unit-test during the implementation of `.dropinf()`  #821, so here it is now.


maartenbreddels:
Thanks!

PR #872, opened 2020-06-27
Fix(core): not all operators where supported in expresso

JovanVeljanoski:
This PR fixes the issue that sometimes happens when chaining certain type of arithmetic operations. The issue is exposed via a unit-test.

Checklist:
- [x] Implement a unit-test
- [ ] Make the unit-test pass


maartenbreddels:
I've removed your test and replaced it by a more fundamental one, that revealed some new issues

maartenbreddels:
Awesome, thanks for  this!

PR #873, opened 2020-06-27
Feat: recognize pandas "string" dtype

JovanVeljanoski:
Enable `vaex.from_pandas()` to recognize the new `pandas.StringDtype` (or "string" for short when using it in pandas).
Also, update the test to cover these cases.

Checklist:
 - [x] Update tests
 - [x] Update functionality
 - [x] Review

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -387,6 +387,7 @@ def from_pandas(df, name="pandas", copy_index=False, index_name="index"):
     import six
     import pandas as pd
     import numpy as np
+    import pyarrow as pa
     columns = {}
 
     def add(name, column):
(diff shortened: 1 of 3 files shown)


packages/vaex-core/vaex/__init__.py
```diff
@@ -392,6 +392,10 @@ def add(name, column):
         # the first test is to support (partially) pandas 0.23
         if hasattr(pd.core.arrays, 'integer') and isinstance(values, pd.core.arrays.integer.IntegerArray):
             values = np.ma.array(values._data, mask=values._mask)
+        elif hasattr(pd.core.arrays, 'StringArray') and isinstance(values, pd.core.arrays.StringArray):
+            values = np.ma.array(values.to_numpy().astype(np.str), mask=values.isna())
```
maartenbreddels, line 396: I think we should go straight to arrow, my guess is the following should work
```suggestion
            values = pa.array(values)
```

tests/repr_test.py
```diff
@@ -73,9 +73,10 @@ def test_repr_from_pandas():
     dd_dict = {
         'boolean': [True, True, False, None, True],
         'text': ['This', 'is', 'some', 'text', 'so...'],
-        'numbers_1': [1, 30, -2, 1.5, 0.000],
-        'numbers_2': [1, None, -2, 1.5, 0.000],
-        'numbers_3': [1, np.nan, -2, 1.5, 0.000],
+        'text_missing': pd.Series(['Some', 'parts', None, 'missing', None], dtype='string'),
+        'float': [1, 30, -2, 1.5, 0.000],
+        'float_missing': [1, None, -2, 1.5, 0.000],
+        'int_missing': pd.Series([1, None, 5, 1, 10], dtype='Int64'),
```
maartenbreddels, line 79: I don't think this should go into repr_test. Internally, vaex should never see these types, this should be handled at a higher level. It would make sense to have `from_pandas_test.py` and check if vaex indeed converts the types.

PR #875, opened 2020-06-28
Fix: minmax operations on datetime columns

JovanVeljanoski:
This PR addresses an issue with computing `.min`, `.max` and `.minmax` on `np.datetime64` columns in the presence of `pd.NaT` (not a time) i.e. missing time values.

Checklist:
- [x] make test
- [ ] make test pass

### DIFF
--- a/tests/datetime_test.py
+++ b/tests/datetime_test.py
@@ -1,7 +1,12 @@
-from common import *
-import numpy as np
-import pandas as pd
 import datetime
+
+import numpy as np
+
+import pandas as pd
+
+import pytest
+
+import vaex
 
 
 def test_datetime_operations():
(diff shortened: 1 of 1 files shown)


tests/datetime_test.py
```diff
@@ -129,17 +136,37 @@ def test_create_datetime64_column_from_str():
...
+
+def test_datetime_nat_minmax_consistency():
+    pandas_df = pd.DataFrame({'t': [pd.NaT,
+                                    datetime.datetime(2019, 1, 1),
+                                    datetime.datetime(2019, 2, 1),
+                                    datetime.datetime(2019, 4, 1),
+                                    datetime.datetime(2019, 5, 1)]})
+    df = vaex.from_pandas(pandas_df)
+
+    t_min = np.datetime64(df.t.min())
```
maartenbreddels, line 164: This is not consistent, since we only ignore NaN, which is a special float number, NaT is kind of a cornercase (I wouldn't even know what 'value' is it actually, or where it is defined. I think this should be `df.t.dropna().min()`

JovanVeljanoski: Agreed.

maartenbreddels: should we close this PR?

JovanVeljanoski: I  wonder if we can catch / intercept the pd.NaT and mask it, something like that.
Otherwise, I guess we can close it.

maartenbreddels: I think we need to move to Arrow for datetime in the future, and then it will be masked values, till that time we need to live with it I think.

PR #880, opened 2020-06-30
Fix: unique on str column that has missing values

### LINKED ISSUE #879: LabelEncoder new columns are object, should be int.
After running LabelEncoder, we get new columns of type object, although all values are ints.
This creates a new level of challenge with missing values, as you get None values in a column with ints.

JovanVeljanoski:
This addresses  an issue of `.unique` when applied on a string column that has missing value.

I am not completely sure, but fixing this might fix #879 

Checklist
- [x] Make test
- [ ] Make em pass

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -160,7 +160,7 @@ class LabelEncoder(Transformer):
         '''
 
         for feature in self.features:
-            labels = df[feature].unique().tolist()
+            labels = vaex.array_types.tolist(df[feature].unique())
             self.labels_[feature] = dict(zip(labels, np.arange(len(labels))))
 
     def transform(self, df):
@@ -227,9 +227,15 @@ class OneHotEncoder(Transformer):
         uniques = []
         for i in self.features:
             expression = _ensure_strings_from_expressions(i)
-            unique = df.unique(expression)
-            unique = np.sort(unique)  # this can/should be optimized with @delay
-            uniques.append(unique.tolist())
+            unique_values = vaex.array_types.tolist(df.unique(expression))
+
+            if None in unique_values:
+                unique_values.remove(None)
+                unique_values.sort()
+                unique_values.insert(0, None)  # This is done in place
+            else:
+                unique_values.sort()
+            uniques.append(unique_values)
         self.uniques_ = uniques
 
     def transform(self, df):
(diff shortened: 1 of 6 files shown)


maartenbreddels:
Actually, this problem should go away in the future, this should not be dtype=object internally, but an Arrow array. Also the return of unique should be. Note that this actually tests dtype=object... *sigh*, anyway, we still want/need to support that though :)

maartenbreddels:
Arg, it wasn't dtype=object. Check my last commit, I think we should totally abandon using dtype=object for string types, we now return an arrow string array. What do you think?

We should only allow numpy for primitive types (int, double.. ) and datetime64. For string we always fall back to arrow, even if the input was dtype=object with only strings.

maartenbreddels:
@JovanVeljanoski there is a failure in the notebook that is not triggered in the unittests, maybe you can isolate the issue?

JovanVeljanoski:
If you are happy with my changes, feel free to merge this

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -226,8 +226,16 @@ def fit(self, df):
         for i in self.features:
             expression = _ensure_strings_from_expressions(i)
             unique = df.unique(expression)
-            unique = np.sort(unique)  # this can/should be optimized with @delay
-            uniques.append(unique.tolist())
+            unique = vaex.array_types.tolist(unique)
+
+            if None in unique:
+                unique = list(filter(None, unique))
```
maartenbreddels, line 232: this is dangerous, for instance `list(filter(None, [0, 1, 2]))` will tell you why. Surprised the tests didn't pick that up.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -242,8 +250,12 @@ def transform(self, df):
...
             for j, value in enumerate(self.uniques_[i]):
                 column_name = self.prefix + feature + '_' + str(value)
-                copy.add_virtual_column(column_name, 'where({feature} == {value}, {one}, {zero})'.format(
-                                        feature=feature, value=repr(value), one=self.one, zero=self.zero))
+                if value is None:
+                    copy[column_name] = copy.func.where(copy[feature].ismissing(), self.one, self.zero)
+                elif isinstance(value, np.float) and np.isnan(value):
+                    copy[column_name] = copy.func.where(copy[feature].isnan(), self.one, self.zero)
+                else:
+                    copy[column_name] = copy.func.where(copy[feature] == value, self.one, self.zero)
```
maartenbreddels, line 258: Yes, much better 👍

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -242,8 +250,12 @@ def transform(self, df):
         for i, feature in enumerate(self.features):
             for j, value in enumerate(self.uniques_[i]):
                 column_name = self.prefix + feature + '_' + str(value)
```
maartenbreddels, line 252: I don't like `None` as name, I think it should be, `_NaN` or `_missing` to be consistent with what we use in vaex.

packages/vaex-ml/vaex/ml/transformations.py
maartenbreddels: unique.remove(None) would also work

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -226,8 +226,16 @@ def fit(self, df):
         for i in self.features:
             expression = _ensure_strings_from_expressions(i)
             unique = df.unique(expression)
-            unique = np.sort(unique)  # this can/should be optimized with @delay
-            uniques.append(unique.tolist())
+            unique = vaex.array_types.tolist(unique)
+
+            if None in unique:
+                unique = list(filter(None, unique))
+                unique = np.sort(unique).tolist()
```
maartenbreddels, line 233: unique.sort() could also work

maartenbreddels:
We need to be sure to merge this, if we do, we should bump to a MAJOR version update, which means vaex v4, and vaex core v3. And if we do this, we want to make more breaking changes at the same time, I'd say go for full Arrow adoption where we can.

PR #882, opened 2020-07-01
fix: evaluate could non evaluate arrow types that could not be conver…

maartenbreddels:
…ted to numpy


maartenbreddels:
failures were due to the dataframe server being down

PR #884, opened 2020-07-02
feat(core): dispatch compute on arrow arrays to numpy

maartenbreddels:
A proof of concept that we can wrap Arrow arrays to dispatch calculations to Numpy.
The main point of this is that we can work towards running the full test suite in an
Arrow backed dataframe. We can gradually remove them over time when Arrow gets more
compute kernels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,7 +79,6 @@ catboost_info/
 .vscode
 
 # symlinks generated in setup.py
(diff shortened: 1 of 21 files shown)


maartenbreddels:
We can have some kernels from arrow:
https://github.com/apache/arrow/pull/7473

PR #885, opened 2020-07-03
Refactor: use Apache Arrow compute for string function

maartenbreddels:
This is a draft PR to check the status of arrow compute with vaex. I think we likely cherry pick from this branch as arrow makes new releases.


xhochy:
FYI: If you have a bit of patience (like 24h of patience), you could use the arrow conda packages in the arrow-nightlies channel instead of building it yourself.

maartenbreddels:
Great, I didn't know it existed, and was difficult to find, thanks a lot!

maartenbreddels:
@JovanVeljanoski would be great if you can add/finish the str->booleans added in https://github.com/apache/arrow/pull/7656
There are a few new ones (see for compute.py), and especially `binary_isascii` is something we may want to think about. Maybe we want to expose this under `Expression.str.isascii()` but also later on under `Expression.binary.isascii()` if we are going to add that accessor.

maartenbreddels:
@JovanVeljanoski I think I want to merge this early and leave it for you to do the rest in a different PR, we need some of this in #865 and I also want to merge that soon.
This means that very likely, master will be breaking, meaning the next release will be vaex v4, do you agree?

packages/vaex-core/vaex/dataframe.py
```diff
@@ -134,6 +134,7 @@ def get_main_executor():
 _doc_snippets['chunk_size_export'] = 'Number of rows to be written to disk in a single iteration'
 _doc_snippets['evaluate_parallel'] = 'Evaluate the (virtual) columns in parallel'
 _doc_snippets['array_type'] = 'Type of output array, possible values are None/"numpy" (ndarray), "xarray" for a xarray.DataArray, or "list" for a Python list'
+_doc_snippets['ascii'] = 'Transform only ascii character (usually faster).'
```
JovanVeljanoski, line 137: character -> characters

PR #886, opened 2020-07-03
New feature: KBinsDiscretizer

JovanVeljanoski:
This is an out-of-core implementation of KBinsDiscretizer. 

It helps to create discretized out of continuous features is a nice and elegant way. Plays very well with GroupByTransformer as well. 

Ofc, fully compatible with the state transfer concept of vaex.

Here is an example:
image
image

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -826,12 +826,12 @@ class KBinsDiscretizer(Transformer):
         if self.strategy == 'uniform':
             bin_edges = {feat: np.linspace(minmax[i, 0], minmax[i, 1], self.n_bins+1) for i, feat in enumerate(self.features)}
 
-        if self.strategy == 'quantile':
+        elif self.strategy == 'quantile':
             percentiles = np.linspace(0, 100, self.n_bins + 1)
             bin_edges = df.percentile_approx(self.features, percentage=percentiles)
             bin_edges = {feat: edges for feat, edges in zip(self.features, bin_edges)}
 
-        if self.strategy == 'kmeans':
+        else:
             from .cluster import KMeans
 
             bin_edges = {}
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -5,6 +5,7 @@ from . import generate
 from .state import HasState
 import traitlets
 from vaex.utils import _ensure_strings_from_expressions
+import numpy as np
 
 help_features = 'List of features to transform.'
 help_prefix = 'Prefix for the names of the transformed features.'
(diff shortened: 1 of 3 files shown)


packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+
+        :param df: A vaex DataFrame.
+        '''
+
+        # We need at least two bins to do the transformations
+        assert self.n_bins > 1, ' Kwarg `n_bins` must be greated than 1.'
+
+        # Find the extent of the features
+        minmax = df.minmax(expression=self.features)
+        epsilon = 1e-8
```
maartenbreddels, line 821: Please turn into a trait, that depends on the scale of the data, so it may require tweaking.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+        # minmax[:, 0] = minmax[:, 0] - epsilon
```
maartenbreddels, line 822:
```suggestion
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+        minmax[:, 1] = minmax[:, 1] + epsilon
+
+        # Bin enges and number of bins
```
maartenbreddels, line 825:
```suggestion
        # Maps features to bin edges and number of bins 
```
Your comment did not add information, we should not do that, since it's a maintenance burden if you change the code (then the comment lies, and it's super confusing). My comment adds a little bit of information, but I will still argue  that the code itself explains it much better. Second, there is no need to assign an empty duct, you don't modify it, you replace it.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+        bin_edges = {}
+        n_bins = {}
+
+        # Determine the bin edges:
```
maartenbreddels, line 829: Not true :)

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+        if self.strategy == 'uniform':
+            bin_edges = {feat: np.linspace(minmax[i, 0], minmax[i, 1], self.n_bins+1) for i, feat in enumerate(self.features)}
+
+        if self.strategy == 'quantile':
```
maartenbreddels, line 833:
```suggestion
        elif self.strategy == 'quantile':
```

packages/vaex-ml/vaex/ml/transformations.py
maartenbreddels, line 829:
```suggestion
        # Determine the bin edges and number of bins depending on the strategy:
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+            percentiles = np.linspace(0, 100, self.n_bins + 1)
+            bin_edges = df.percentile_approx(self.features, percentage=percentiles)
+            bin_edges = {feat: edges for feat, edges in zip(self.features, bin_edges)}
+
+        if self.strategy == 'kmeans':
```
maartenbreddels, line 838:
```suggestion
        elif self.strategy == 'kmeans':
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+
+            for i, feat in enumerate(self.features):
+
+                # Deterministic initialization with uniform spacing
+                uniform_edges = np.linspace(minmax[i, 0], minmax[i, 1], self.n_bins+1)
+                init = ((uniform_edges[1:] + uniform_edges[:-1]) * 0.5).tolist()
+                init = [[elem] for elem in init]
+
+                # KMeans strategy
+                km = KMeans(n_clusters=self.n_bins, init=init, n_init=1, features=[feat])
```
maartenbreddels, line 849: Do you agree init was a really bad name? I guess we copied it from sklearn? Can be make the variable here more meaningful?

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+                km.fit(df)
+                centers = np.sort(np.array(km.cluster_centers).flatten())
+                be = (centers[1:] + centers[:-1]) * 0.5
+                bin_edges[feat] = np.r_[minmax[i, 0], be, minmax[i, 1]]
```
maartenbreddels, line 853: This is a place where I'd put/expect a comment because here the code is not that trivial. What is the goal of this code, I don't know/never use `np.r_` for instance.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+
+        # Remove bins whose width are too small (i.e., <= 1e-8)
```
maartenbreddels, line 855: Those are the comments that help!

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+        for feat in self.features:
+            mask = np.ediff1d(bin_edges[feat], to_begin=np.inf) > 1e-8
```
maartenbreddels, line 857: Do we need ediff https://stackoverflow.com/questions/39014324/difference-between-numpy-ediff1d-and-diff#:~:text=0%20diff()%20supports%20padding,ediff1d()%20for%20boolean%20arrays. ?

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -776,6 +777,115 @@ def transform(self, df):
...
+            be = bin_edges[feat][mask]
+            if len(be) - 1 != self.n_bins:
+                print(f'Bins whose width are too small (i.e., <= '
+                      f'1e-8) in "{feat}" are removed. Consider '
+                      f'decreasing the number of bins.')
```
maartenbreddels, line 862: Not sure we should ever print, maybe use logging here, log.info(...). Also, maybe a good time to agree on a max line width, I'd say 120 or 140? I think it increases readability a lot.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -845,3 +955,103 @@ def transform(self, df):
                 join_name = self.rprefix + join_name + self.rsuffix
             df[join_name] = df[self.by].map(mapper, allow_missing=True)
         return df
+
+# class BinByTransformer(Transformer):
+#     ''' Binby transformer - work in progress '''
+#     by = traitlets.Unicode(allow_none=False, help='The feature on which to do the binning.')
+#     n_bins = traitlets.Int(allow_none=False, default_value=5, help='Number of bins. Must be greater than 1.')
```
maartenbreddels, line 962: I guess we can remove this all

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: `strategy` is a `traitlet.Enum` so it can only have the pre-defined 3 values. Given that, I find 3 separate `if` statements easier to read (personally). If you feel strongly about it, I can convert it to `elseif` form

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Yup, guilty as charged. Will change it

JovanVeljanoski: Although, the keyword in KMeans is also init so... :) perhaps not the worse choice

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: How about `warning.warn`? We do that elsewhere (e.g. train_test_split)

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Good point, done.

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Right!

packages/vaex-ml/vaex/ml/transformations.py
maartenbreddels: Yes, very strongly, this is very uncommon to see. Multiple if statements read as if multiple cases can happen, if/elif/elif/else is basically an exclusive set.

packages/vaex-ml/vaex/ml/transformations.py
maartenbreddels: Yes, warning sounds good,

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Done!

PR #889, opened 2020-07-07
Refactoring: if 1 -> if True

marload:
I think True is more intuitive than 1 when executing conditional statement unconditionally.

Thank You!


maartenbreddels:
I agree! Thanks

PR #893, opened 2020-07-09
Use print() function in both Python 2 and Python 3

cclauss:
@maartenbreddels Your review, please.

Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.


maartenbreddels:
Hi, yes, sound good. It's an old file that should probably be removed if the future though. Thanks!

Issue #608, opened 2020-02-22
csv file without header not supported


argenisleon:
Hi,
I am trying to load a CSV with no header  using

```
df = vaex.open('data/star0000-1.csv',sep=",", header=None, error_bad_lines=False)
```

but I get

```
could not convert column 0, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 0, error: TypeError('getattr(): attribute name must be string')
could not convert column 1, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 1, error: TypeError('getattr(): attribute name must be string')
could not convert column 2, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 2, error: TypeError('getattr(): attribute name must be string')
could not convert column 3, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 3, error: TypeError('getattr(): attribute name must be string')
could not convert column 4, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 4, error: TypeError('getattr(): attribute name must be string')
could not convert column 5, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 5, error: TypeError('getattr(): attribute name must be string')
could not convert column 6, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 6, error: TypeError('getattr(): attribute name must be string')
could not convert column 7, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 7, error: TypeError('getattr(): attribute name must be string')
could not convert column 8, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 8, error: TypeError('getattr(): attribute name must be string')
could not convert column 9, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 9, error: TypeError('getattr(): attribute name must be string')
could not convert column 10, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 10, error: TypeError('getattr(): attribute name must be string')
could not convert column 11, error: TypeError('getattr(): attribute name must be string'), will try to convert it to string
Giving up column 11, error: TypeError('getattr(): attribute name must be string')

```

Also it takes like 20x  the time to do nothing, any help?

JovanVeljanoski:
Hi @argenisleon 

[...]

PR #897, opened 2020-07-10
Feature 608 - fix loading csv without header

mkst:
Felt like the right place to apply the fix, but let me know if you think it should be done somewhere else :)

fixes https://github.com/vaexio/vaex/issues/608

**edit:**
@maartenbreddels - anything I need to do here?


maartenbreddels:
Nice work Mark, and indeed that seems like a good place. 
The tests though have moved to the `tests/` directory, and we now use pytest for testing. Would you mind moving it, and getting rid of the class (just using functions is ok for pytest)? I suggest also a csv file at `tests/data/noheader.csv` without a header and a test at `tests/from_csv_test.py` to make sure this keeps working.

What do you think.

If unclear, me or Jovan can help.

mkst:
@maartenbreddels, updated per your comments - hopefully good to go?

maartenbreddels:
Took a bit long to get to this, thanks! I've changed it a bit, because vaex now support column names that are not valid identifiers. So now the column names can be '0' and '1' etc.

mkst:
> Took a bit long to get to this, thanks! I've changed it a bit, because vaex now support column names that are not valid identifiers. So now the column names can be '0' and '1' etc.

No problem - had to go look at the original commit to refresh my memory! Thanks 👍

maartenbreddels:
Thanks Mark, finally merged :)

PR #898, opened 2020-07-10
Feat: stream hdf5 files directly from GCS

JovanVeljanoski:
This PR enables the streaming of HDF5 files directly from Google cloud storage. 
Vaex could already do this from S3, now we bring support for GCS as well. 

The usage is the same:
```
import vaex
df = vaex.open('gs://vaex-data/testing/xys.hdf5?cache=false&token=anon')
```

Note the slight differences compared to reading from S3. For reading data from GCP we are currently using `gcsfs` so the url parameters should be compatible with `gcsfs.GCSFileSystem`


maartenbreddels:
I'd like to see the s3 and gcs test merged into 1, say, fsspec_test that would use fixture to run both cases. I also want to refactor this, as part of #865. We could merge this as is, as these 'issues' will reveal itself, what do you think?

PR #908, opened 2020-07-17
Fix: value counts on added/concatenated strings

JovanVeljanoski:
This PR handles the issue of doing value counts on a virtual column which is created by adding or concatenating strings.

 - [x] Reproducing via a unit-test
 - [ ] Make test pass


maartenbreddels:
Note to future self: rebase after we merge #885

maartenbreddels:
CI failures are unrelated.

PR #923, opened 2020-08-04
Rename plotting methods to more standard 

JovanVeljanoski:
This PR simply renames the `df.plot()` and `df.plot1d` methods to `df.heatmap()` and `df.histogram()` respectively, to be more in line with what is naively expected. The old methods still work, you will get a deprecation warning. So this change is not breaking (yet).

In addition, it adds a convenient expression method, so that now one can create a histogram simply by:
`df.x.histogram()`.

- [x] Rename `plot()` -> `heatmap()`
- [x] Rename `plot1d()` -> `histogram()`
- [x] Add deprecation warnings to the old methods
- [x] Add an expression convenience method for fast calling of `histogram()`
- [ ] Review


xdssio:
Great idea

xdssio:
Maybe we can get the *plot()* to do the "right" plot per column/columns. maybe in another PR later.

maartenbreddels:
The meaning of histogram is kinda of ambiguous right, np.histogram gives you data for instance. I think this will be highly confusing.
df.x.viz.histogram() ? and only df.viz.histogram? (or s/viz/plot)

JovanVeljanoski:
I think the word "histogram" should always refer to the actual plot. Numpy have slightly mis-named that method, in fact that does binning, does not draw the histogram. I don't think there is ambiguity

I am also fine having `df.x.viz.histogram()` and  `df.viz.histogram()`. In fact that might open up space to try implement other kind of plots without cluttering.. so yeah, good idea.
Does this mean we need to use the dataframe accessors to make it like that? Should I give it a try?

maartenbreddels:
https://en.wikipedia.org/wiki/Histogram#Mathematical_definition

maartenbreddels (changes requested):
I hope you like my changes, so now it's df.viz.histogram(df.x) and df.x.viz.histogram(). Would be great if you can update the docs, running make in the docs folder revealed some issues.

JovanVeljanoski:
Yup, I will update the docs / notebooks to reflect the changes.

I also noticed that the methods that are now in `df.viz.` (heatmap, histogram, scatter), do not tab-complete the kwargs in a Jupyter notebook. On the other hand for `df.x.viz.histogram(...)` tab completion work very well! Is there something we can do about this? Heatmap has a lot of arguments..

maartenbreddels:
needs a vix->viz rename

PR #925, opened 2020-08-05
[Feat]: shift column

JovanVeljanoski:
This is an attempt at implementing `df.shift(...)`.  Currently, the methods shifts columns by a specified number of rows.

Some notes:
 - currently the test fails when a virtual column is shifted by a positive number of periods. It is most likely a concat related issue.
 - Ideally we should like to support shifting by a period related to a datetime column by a given frequency (as in pandas), thus being able to shift for a day/month/minute etc...
 - We need to think on how to add this to the state (or what the state will become in the future). Maybe this is not needed, but perhaps a discussion would be useful.

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4749,6 +4749,47 @@ class DataFrame(object):
                 del df[column]
         return df
 
+    @docsubst
+    def shift(self, column_names=None, periods=1, cyclic=False, fill_value=None):
+        """Shift the locations of the specified columns by a certain number of periods/rows.
+
+        :param:
+        """
+        def get_dtype(expr):
+            if self[expr].is_string():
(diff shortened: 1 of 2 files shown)


sreevanii:
hai jovan,
how to insert a virtual column in the specified index

JovanVeljanoski:
> hai jovan,
> how to insert a virtual column in the specified index

This is a wrong place to ask a random question that is unrelated to the PR. Please use the issues board.

Also, please read through the documentation to get a better understanding of how vaex works.

sreevanii:
oh! I'm sorry

> 
> 
> > hai jovan,
> > how to insert a virtual column in the specified index
> 
> This is a wrong place to ask a random question that is unrelated to the PR. Please use the issues board.
> 
> Also, please read through the documentation to get a better understanding of how vaex works.

oh! I'm Sorry

Issue #920, opened 2020-08-02
[FEATURE-REQUEST] Predict probability for sklearn models


matnguyen:
I have looked through the documentation but did not find any way to do this, so was wondering if it was implemented in any ways.

Basically, for some models such as logistic regression, the model can predict probabilities of an observation being in an class, rather than just the class itself. This may be useful to self-define probability thresholds for a class. Is this currently implemented for sklearn models through vaex.ml.IncrementalPredictor or vaex.ml.Predictor?

JovanVeljanoski:
Hi @matnguyen 

This is currently not possible out-of-the box. It is however something we would like to have, so it is very likely to be available soon! 

The good news is that it is rather easy to code this up yourself if needed. Check out the example below. 
You can easily modify the below example to work with `IncrementalPredictor`, or if you want to use something like `predict_log_proba`.

Let me know if you have any questions about this. 

Let's keep this issue open until we implement this properly. 

```
import vaex
import vaex.ml

from vaex.ml.sklearn import Predictor
from sklearn.ensemble import GradientBoostingClassifier

df = vaex.ml.datasets.load_iris_1e4()
df_train, df_test = df.ml.train_test_split(test_size=0.1, verbose=False)

features = df_train.column_names[:4]
target = 'class_'

my_predictor = Predictor(model=GradientBoostingClassifier(n_estimators=31), 
                         features=features, 
                         target=target, 
                         prediction_name='pred')
my_predictor.fit(df_train)
                         
# Predict proba in memory, get an numpy array
pred_in_memory = my_predictor.model.predict_proba(df_test[my_predictor.features].values)

# Predict proba as a transformer 
def transform_prob(df):
    # define the function to be lazy-fied
    def my_func(*args):
        X = np.vstack([arg.astype(np.float64) for arg in args]).T.copy()
        return my_predictor.model.predict_proba(X)   # you can change this to predict_log_proba if needed for example
    copy = df.copy()  # This is a shallow copy, no extra RAM used
    lazy_function = copy.add_function('sklearn_predict_proba_function', my_func, unique=True)
    expression = lazy_function(*my_predictor.features)
    copy.add_virtual_column(my_predictor.prediction_name, expression, unique=False)
    return copy

df_test_pred = transform_prob(df_test)
df_test_pred.head(5)

```

PR #927, opened 2020-08-07
Feat(ml): Support for `predict_proba` and `predict_log_proba` for sklearn classifiers

JovanVeljanoski:
Requested in #920 .

This PR adds support for obtaining probabilities and log probabilities from sklearn classifiers. 
The `Predictor` and `IncrementalPredictor` classes now have a `prediction_type` kwarg, which can be set to `predict` (default), `predict_proba` and `predict_log_proba`. Each option will trigger the appropriate method in the sklearn classifier at prediction time.

Note: in this PR we are also removing a deprecated class (`vaex.ml.sklearn.SKLearnPredictor`), as it's deprecation period has expired. 

Checklist:
 - [x] Implementation 
 - [x] Update of unit-tests
 - [ ] Review

### DIFF
--- a/docs/source/example_ml_iris.ipynb
+++ b/docs/source/example_ml_iris.ipynb
@@ -338,10 +338,10 @@
     "                                          random_state=42)\n",
     "\n",
     "# Make it a vaex transformer (for the automagic pipeline and lazy predictions)\n",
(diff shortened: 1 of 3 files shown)


maartenbreddels:
great work, nice clean commits

PR #940, opened 2020-08-18
Feat additional numpy functions

JovanVeljanoski:
Adding a couple of additional numpy functions.

- [x] Add `round`
- [ ] Add `argsort` - descoped for this PR (does not work across chunks)
- [x] Alphabetically sort numpy functions for better overview of that is there
- [x] Remove accidental duplicates
- [ ] Review

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -43,6 +43,9 @@ jobs:
       shell: bash
       run: |
         ./ci/02-create-vaex-dev-env.sh ${{ matrix.python-version }}
+        source activate vaex-dev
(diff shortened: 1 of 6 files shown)


maartenbreddels:
Argsort cannot be executed per chunk, like sort

maartenbreddels:
Ok, with mamba, we get a better solution to the environment, so for some reason conda downgrades dask a lot. OTOH, mamba fails in windows + py37.

maartenbreddels:
I don't really like the solution, this could be a conda issue, and maybe we're stretching conda to the limits of what it can do. This is showing mamba building on top of libsolv is probably a good idea! Lets hope mamba on GHA will work better in the future.

Issue #948, opened 2020-08-28
 fillna func with blank values convert column from sting type to object type 


salwaen:
Hi, 

### i have column with type string has none 

````
#       groupno    Name
0       0          name1
1       0          name1
2       0          name1
3       0          name1
4       1          name1
...     ...        ...
36,766  1          None
36,767  2          None
36,768  3          None
36,769  4          None
36,770  4          None
[----------------------------------------]  0.00% estimated time: unknown

````

````
print(vaex_df.Name.dtype)
````
### Output
````
<class 'str'>
````

###  with fillna func 
````
 vaex_df = vaex_df.fillna(value='', column_names=['Name'])
 print(vaex_df[['groupno', 'Name']])
 print(vaex_df['Name'].dtype))
````
### Output
````
#       groupno    Name
0       0            nam1
1       0            nam1
2       0            nam1
3       0            nam1
4       1            nam1
...     ...        ...
36,766  1
36,767  2
36,768  3
36,769  4
36,770  4
object
````
### export to HDF5 
````
df_vaex.export_hdf5('./data/io/output_data.hdf5')
````

### Output
````
TypeError: Object dtype dtype('O') has no native HDF5 equivalent
2020-08-28 10:38:59,196 root [ERROR] error creating dataset for 'Name', with type dtype('O') 
````
I know HDF5 doesn't supports object type, is there a solution?
thanks

JovanVeljanoski:
Yeah I think this is a bug..

How do you read in the data into vaex to begin with?

If you use something like `pandas.read_csv` or `vaex.read_csv` (which is the same), vaex will automatically try to convert dtype objects into to best possible primitive dtype. If it can't it will try to convert it to string. Sometimes (I've not been able to consistently reproduce this - might try later this week), this process does not go as planned, and the underlying data structure is still on some level considered of dtype object.

So in the meantime, there is a workaround. The best way is to read in the data as a pandas dataframe, and clean it there such that the data contains only primitive types (ints, floats, strings). Then you can pass it to vaex `vaex.from_pandas(..)` and all should be fine. 

I will mark this as bug, and hopefully we can reproduce it!

salwaen:
I read the data  with   `pd.read_sql_query` then into vaex  ` vaex .from_pandas()`

[...]

PR #953, opened 2020-08-31
Fix: Preserve dtype string after fillna() operation

JovanVeljanoski:
Issue #948 exposed a bug in which a column of type `str` gets accidentally converted to `object` after a `fillna` operation. 
This PR should address that.

This happens because, at the moment, we cast everything to numpy array in the processes of filling the N/A entries. If the columns / expression happens to be string, this is never converted back from numpy into the proper string/arrow format, resulting with a dtype `object`. I hope this to be an easy fix for @maartenbreddels.

- [x] Implement a unit test
- [ ] Make the test pass


maartenbreddels:
Ok, this will be simpler after the #865 is merged.

maartenbreddels:
This is going to be part of #865 actually.

maartenbreddels:
Thanks for the tests!

PR #961, opened 2020-09-04
Feature: adding ipython tab-completion

JovanVeljanoski:
Adding tab competion when accessing columns using the dictionary syntax.

example:
`df['hom' (tab)]` -> `df['home_dest']`

- [ ] Review


maartenbreddels:
Awesome, we've been sloppy with the changelog btw, we should pick that habit up again.

PR #966, opened 2020-09-10
chore: a failing install of a subpackage did not stop the installation

maartenbreddels:
the pip install should fail because vaex-ml has wrong pinned dependencies.


maartenbreddels:
mamba fails to create a py36 environment, i'm assuming it's a fluke.

PR #980, opened 2020-09-22
[Feat]: Pipeline

xdssio:
The new Pipeline object handle all the hurdles that moves from notebook to deployment.
Purpose: When the notebook is done, the data-science side of things is done in a way that translate between deployment processes.

Use case - When a single deployment setup is done, for example a local server, it would work for any data, training and inference (as long the requirements are there). This can be extended trivially to AWS Sagemaker and GCP.

## Key features.
1. **Sklearn API** - fit, fit_transform, predict apply and return what you would expect in a sklearn API. 

- `fit()` retrain the data from scratch (partial_fit can be added later) - can define any static function which returns a transformed dataframe. 
- `transform()` run all given transformations including filtering.
- `predict()` returns a single numpy array of a given columns as if it was an estimator.

2. **Inference** - the `inference()` function call in the same as transform but do not apply selections. It can also handle many different types of inputs like json, dicts, and pandas.

3. The pipeline can be saved and load with ease.

## Examples
Simple Usage:
```
  from sklearn.linear_model import LogisticRegression
  from vaex.ml.sklearn import SKLearnPredictor
  
  train, test = vaex.ml.datasets.load_iris().split_random(0.8)
  model = SKLearnPredictor(model=LogisticRegression(), features=['x', 'y', 'w', 'z'], target='class_')
  train = model.fit_transform(train)
 
  pipeline = Pipeline.from_dataaset(train)
  pipeline.save(<path>)

  # Server side
  pipeline = Pipeline.from_file(<path>)
  pipeline.inference(test)
  
```

Complicated example:
```
  
   def fit(df):    
        from sklearn.linear_model import LogisticRegression
        from vaex.ml.sklearn import SKLearnPredictor
        df['x'] = df.sepal_length * 1
        df['y'] = df.sepal_width * 1
        df['w'] = df.petal_length * 1
        df['z'] = df.petal_width * 1
        model = SKLearnPredictor(model=LogisticRegression(), features=['x', 'y', 'w', 'z'], target='class_')
        model.fit(df)
        df = model.transform(df)
        df['response'] = df['prediction'].map({0: 'class A', 1: 'class B', 2: 'class C'})
        return df

    train, test = vaex.ml.datasets.load_iris().split_random(0.8)


    pipeline = Pipeline.from_dataset(train, fit=fit)
    pipeline.fit(train) # can train anywhere
    pipeline.inference(test)
```

### DIFF
--- a/packages/vaex-jupyter/setup.py
+++ b/packages/vaex-jupyter/setup.py
@@ -14,7 +14,7 @@ author_email = 'maartenbreddels@gmail.com'
 license = 'MIT'
 version = version.__version__
 url = 'https://www.github.com/maartenbreddels/vaex'
-install_requires_jupyter = ['vaex-core>=2.0.0,<3', 'vaex-viz', 'bqplot>=0.10.1', 'ipyvolume>=0.4', 'ipyleaflet', 'ipympl', 'ipyvuetify>=1.2.2,<2']
+install_requires_jupyter = ['vaex-core>=2.0.0,<3', 'vaex-viz', 'bqplot>=0.10.1', 'ipyvolume>=0.4', 'ipyleaflet', 'ipympl', 'ipyvuetify>=1.2.2,<2', 'xarray']
 
 setup(name=name + '-jupyter',
       version=version,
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Thanks, as mentioned in private, this is a good start for a discussion about pipelines, which we should probably do via a high bandwidth channel (video)

JovanVeljanoski:
This is closed in favor of #1332

PR #983, opened 2020-09-24
chore(ci): make jupyter test more robust

maartenbreddels:


maartenbreddels:
Unclear what the error was exactly, or why it was caused, but the tests are more robust now, and pytest-timeout should help future hangs.

PR #996, opened 2020-10-03
Chore: LightGBM wrapper cleanup

JovanVeljanoski:
This PR allows safely usage of the lightgbm wrapper through vaex.ml for the latest version(s) of lightgbm (>=3.0.0).

Vaex.ml implemented a special `VaexDataset` class built on top of the `lightgbm.Dataset` class, and allowed for more memory efficient way of passing vaex data to lightgbm. This came at some performance penalty. 

Lightgbm 3.0.0 introduced breaking changes to their Dataset. We do not currently have the capacity to maintain the `VaexDataset`, so it is best to remove it for now. If anybody in the community is willing to pick this up, they are welcome to.

### DIFF
--- a/ci/conda-env.yml
+++ b/ci/conda-env.yml
@@ -5,7 +5,7 @@ geopandas
 graphviz
 h5py
 # libcxx=9.0.0
-lightgbm<3.0.0
+lightgbm
 matplotlib-base
 notebook
 numba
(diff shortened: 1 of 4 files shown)


maartenbreddels:
Awesome, thanks! Note to both of us: include changelog entries :)

PR #1003, opened 2020-10-09
refactor!: df.export had a lot of redundant arguments and chunked export

maartenbreddels:
Redudant/legacy argument:
 * column_names: use df[column_names] instead
 * shuffle: use df.shuffle() instead
 * virtual: use df[df.get_column_names(virtual=False)], but I don't think this argument makes sense any more
 * sort: use df.sort() instead
 * selection: use df.filter(selection) instead, export does not support multiple selections either

df.export_chunked('chunk-{i:05}.parquet') will output N files in parallel, it's quite convenient and fast already.


packages/vaex-core/vaex/dataframe.py
```diff
@@ -5868,153 +5872,163 @@ def mangle_name(prefix, name, suffix):
...
         else:
             write(path_or_writer)
 
-    def export_parquet(self, path, column_names=None, byteorder="=", shuffle=False, selection=False, progress=None, virtual=True, sort=None, ascending=True):
-        """Exports the DataFrame to a parquet file
+    @docsubst
+    def export_parquet(self, path, progress=None, parallel=True, **kwargs):
+        """Exports the DataFrame to a parquet file.
+
+        Note: This may require that all of the data fits into memory (mmapped data is an exception).
```
JovanVeljanoski, line 5944: I would prefer if this is spelled out (i.e. memory mapped). Also not quite clear when the exception happens. I tried to export 60GB hdf5 file to parquet with this method, and i could not do it. Might be other issue (see general comments).

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5868,153 +5872,163 @@ def mangle_name(prefix, name, suffix):
...
+            Use :py:`DataFrame.export_chunks` to write to multiple files in parallel.
```
JovanVeljanoski, line 5945: fix typo, also see comment addressing this method.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5868,153 +5872,163 @@ def mangle_name(prefix, name, suffix):
...
         """
-        import vaex.arrow.export
-        vaex.arrow.export.export_parquet(self, path, column_names, byteorder, shuffle, selection, progress=progress, virtual=virtual, sort=sort, ascending=ascending)
+        import pyarrow.parquet as pq
+        table = self.to_arrow_table(chunk_size=None, parallel=False)
+        pq.write_table(table, path, **kwargs)
 
-    def export_hdf5(self, path, column_names=None, byteorder="=", shuffle=False, selection=False, progress=None, virtual=True, sort=None, ascending=True):
+    @docsubst
+    def export_chunked(self, path, progress=None, chunk_size=default_chunk_size, parallel=True):
```
JovanVeljanoski, line 5959: This is a good idea! However I am not sure if this is a good name of this method. Maybe we should discuss this.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5868,153 +5872,163 @@ def mangle_name(prefix, name, suffix):
...
+        """Export the DataFrame to multiple files in parallel.
+
+        :param str path: Path for file, formatted by chunk index i (e.g. 'chunk-{{i:05}}.parquet')
+        :param int chunk_size: Number of rows for each file (except possibly the last)
+        :param bool parallel: {evaluate_parallel}
+        """
+        from .itertools import pmap, pwait, buffer, consume
```
JovanVeljanoski, line 5966: Specifying output name(s) like this is very uncommon. I suggest we keep it, but instead of raising an error if this convention is not met, to automatically determine the number of chunks needed based on the chunk size, and append the relevant chunk number to the "base" filename, with the most appropriate number formatting.

JovanVeljanoski (changes requested):
General feedback:

I tested this PR but using a single large CSV file of 70 GB (the Chicago taxi dataset). The csv is clean and consistent. 

Reading and exporting the file with `vaex.from_csv(file, convert=True)` crashes at the final step - concatenation and export of the multiple intermediate hdf5 files. The intermediate hdf5 files are exported correctly. Once can successfully open, concatenate and export them to hdf5 in a separate session. 

Manually iterating over the csv file with `vaex.open(file, chunk_size=25_000_000)`, and then opening the intermediate hdf5 files, concatenating them and and exporting them as a single df5 file works very well in the same script/notebook (tried both).  Which is surprising that the 'one-liner' fails since it is meant to do the very same thing.

Working with the single large hdf5 file created with the previous step, i tried to export it to `.arrow`. This should now be supported since chunking is supported. However attempting this crashes the kernel immediately (looks like it happens after the 1st chunk is done).

Exporting to `.parquet` also failed, some issue related to string support (long strings?).

Exporting using `export_chunked` also failed (unknown issues - more investigation needed), but I suspect is related to the same problems arrow/parquet suffer as explained above.

I can reproduce all of this issues with the big CSV/HDF5 file consistently. Please let me know if I should try to write unit-tests using smaller portions of this data.

JovanVeljanoski:
Once all this is working, it would be nice if we can update the docs right away (either in this, or in an accompanying PR).

packages/vaex-core/vaex/__init__.py
```diff
@@ -567,7 +570,7 @@ def _from_csv_convert_and_read(filename_or_buffer, copy_index, maybe_convert_pat
         logger.info('converting %d chunks into single HDF5 file %s' % (len(converted_paths), combined_hdf5))
         dfs = [vaex.file.open(p) for p in converted_paths]
```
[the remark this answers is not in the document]
maartenbreddels: Yes, I agree, lets remove it.

packages/vaex-core/vaex/dataframe.py
maartenbreddels: export_many ? export_parallel ?

packages/vaex-core/vaex/dataframe.py
maartenbreddels: done

maartenbreddels:
There was huge memory copying going on, because we were passing views of super large arrow arrays around. We now trim those, so memory usage should be better.

maartenbreddels:
Failure in test_thread_safe seem unrelated... we've seen that on appveyor a lot.

Issue #718, opened 2020-05-05
State can not be saved when fillna and mapping is applied to a DataFrame


JovanVeljanoski:
We are encountering a problem when trying to save the state to disk when a `fillna` is followed by a method that does `map`. This is an example:

```
import vaex 
import vaex.ml
df = vaex.ml.datasets.load_titanic()
df['embarked'] = df.embarked.fillna("NA")
le = vaex.ml.LabelEncoder(features=['embarked'])
le.fit(df)
df = le.transform((df))
df.state_write('tmp_state.json')   # Crash happens here
```

This is the end part of the stack-trace:
```
~/Work/vaex/packages/vaex-core/vaex/json.py in encode(obj)
    118     for serializer in serializers:
    119         if serializer.can_encode(obj):
--> 120             return serializer.encode(obj)
    121 
    122 

~/Work/vaex/packages/vaex-core/vaex/json.py in encode(obj)
     97             'data': {
     98                 'values': values,
---> 99                 'count': obj.count,
    100                 'nan_count': obj.nan_count,
    101                 'missing_count': obj.null_count

AttributeError: 'vaex.superutils.ordered_set_object' object has no attribute 'count'
```

(Issue found by @xdssio )

maartenbreddels:
The underlying issue is that we somehow get to dtype=object, I think we should avoid it, but for sure a bug.

JovanVeljanoski:
Ok i found out where the object happens. In short, if you do `fillna` on a string type it results in an object type. Fixing this will probably fix this issue. 

Here is an example:
<img width="399" alt="image" src="https://user-images.githubusercontent.com/18574951/81079877-67069000-8ef0-11ea-947a-28878c59bc60.png">

<img width="501" alt="image" src="https://user-images.githubusercontent.com/18574951/81079901-7259bb80-8ef0-11ea-8e6e-e3af4f076ece.png">

JovanVeljanoski:
This has been fixed.

PR #1007, opened 2020-10-13
Fix: various improvements related to string columns.

JovanVeljanoski:
This PR fixes several (hopefully) small issues that pop up when using string columns with vaex, now that it is primarily arrow based.

List of new / updated unit tests:
 - using `vaex.from_arrow_table(pa_table, as_numpy=True)` makes string columns be dtype 'O' in vaex
 - selections made on arrow/string columns
 - `df.x.fillna()` converts string type to 'O' type. This is not captured by `df.is_string(df.x)` but is captured by `df.x.dtype`
 - `df.func.where(df.x == 'string_value', 'new_value', 'df.x')` casts the resulting expression to dtype 'O'. This is captured by `df.is_string(...)`


JovanVeljanoski:
Also see https://github.com/vaexio/vaex/issues/718

PR #1017, opened 2020-10-21
Fix: datetime64 arithmetics with a constant

JovanVeljanoski:
This PR enables the subtraction a datetime64 expression from a constant np.datetime64 values.

Note: the reverse works, i.e. subtracting a constant np.datetime64 value from a datetime expression works well. The test is included for completeness.

- [x] Implement unit test
- [ ] Make tests pass

### DIFF
--- a/tests/datetime_test.py
+++ b/tests/datetime_test.py
@@ -1,3 +1,4 @@
+import datetime
 from common import *
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


maartenbreddels:
I dont think we can fix this, since NumPy gets a first try to perform this operation (the subtract), and fails, we don't get a chance. Possibly we could solve this with NEP13/18, should we look into this?

PR #1036, opened 2020-11-03
fix: concat and chunk_iterator for no columns would get stuck

maartenbreddels:


maartenbreddels:
Failures are unrelated.

PR #1040, opened 2020-11-03
Improve vaex.open docstrings

JovanVeljanoski:
This PR hope to achieve two things:
 - Improve the docstring of `vaex.open(...)`
 - Refactor a bit how we generate the docstrings for common keywords.
 - Remove `copy_index` from `vaex.open`. This kwarg is not general, but only applies when a csv file is read via `pandas.read_csv`. It can still be passed as part of the `**kwargs` so there is no loss of functionality here, and old code will **not** break.

Checklist
- [x] Put the common kwargs docstrings in a separate dedicated file to be better reused amongst different modules and packages
- [x] Add intersphinx mapping to the docs of s3fs, gcsfs, (py)arrow (groundworks for improving the docs & docstrings)
- [x] Improve the docstrings for `vaex.open`
- [x] Add missing docstring info on the various `df.export_{...}` methods.
- [x] Remove `copy_index` as a defined keyword in `vaex.open`. It can still be passed as part of `**kwargs` so this is not a breaking change.

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -101,43 +101,54 @@ def open(path, convert=False, shuffle=False, fs_options={}, *args, **kwargs):
     >>> df = vaex.open('somedata*.csv', convert='bigdata.hdf5')
 
     :param str or list path: local or absolute path to file, or glob string, or list of paths
-    :param convert: convert files to an hdf5 file for optimization, can also be a path
+    :param convert: Uses `dataframe.export` when convert is a path. If True, ``convert=path+'.hdf5'``
+                    The conversion is skipped if the path already exists.
     :param bool shuffle: shuffle converted DataFrame or not
-    :param bool copy_index: copy index when source is read via pandas
-    :param dict fs_options: coming soon...
+    :param dict fs_options: Extra arguments passed to an optional file system if needed:
+        * Amazon AWS S3
+            * :py.class:`pyarrow.fs.S3FileSystem` - If supported by Arrow.
+            * :py:class:`s3fs.core.S3FileSystem` - Used for globbing and fallback.
+        * Google Cloud Storage
+            * :py:class:`gcsfs.core.GCSFileSystem`
+        In addition you can pass the boolean "cache" option.
     :param args: extra arguments for file readers that need it
     :param kwargs: extra keyword arguments
     :return: return a DataFrame on success, otherwise None
     :rtype: DataFrame
 
-    S3 support:
+    Cloud storage support:
 
-    Vaex supports streaming of hdf5 files from Amazon AWS object storage S3.
-    Files are by default cached in $HOME/.vaex/file-cache/s3 such that successive access
-    is as fast as native disk access. The following url parameters control S3 options:
+    Vaex supports streaming of HDF5 files from Amazon AWS S3 and Google Cloud Storage.
+    Files are by default cached in $HOME/.vaex/file-cache/(s3|gs) such that successive access
+    is as fast as native disk access.
+
+    The following common fs_options are used for S3 access:
 
      * anon: Use anonymous access or not (false by default). (Allowed values are: true,True,1,false,False,0)
-     * use_cache: Use the disk cache or not, only set to false if the data should be accessed once. (Allowed values are: true,True,1,false,False,0)
-     * profile and other arguments are passed to :py:class:`s3fs.core.S3FileSystem`
+     * cache: Use the disk cache or not, only set to false if the data should be accessed once. (Allowed values are: true,True,1,false,False,0)
+     * profile and other arguments are passed to `pyarrow.fs.S3FileSystem` or :py:class:`s3fs.core.S3FileSystem`.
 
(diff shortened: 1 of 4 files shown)


packages/vaex-core/vaex/__init__.py
```diff
@@ -117,9 +117,10 @@ def open(path, convert=False, shuffle=False, copy_index=False, fs_options={}, *a
     :param str or list path: local or absolute path to file, or glob string, or list of paths
     :param convert: convert files to an hdf5 file for optimization, can also be a path
     :param bool shuffle: shuffle converted DataFrame or not
+    :param bool copy_index: copy index when source is read via pandas
```
maartenbreddels, line 120: I don't think we want to have this here, this means it will be part of the API, and with arrow csv reading we will never have this again. we should only have arguments that are common (or very common) to all the dataset reading.

JovanVeljanoski: That is not a new thing.. i just re-ordered it so that the order of the kwargs in the docstring matches the one in the function.

I suggest we remove it when `read_csv` is shifted from pandas to arrow. Also I don't know if we should fully drop `pandas.read_csv`, might depend on the type of options the arrow reader provides.

maartenbreddels: Maybe we should prepare for this, read_cvs_pandas, read_cvs_arrow, and a read_csv with a backend='pandas' argument?

JovanVeljanoski: Sounds good to me. I would propose `read_csv_pandas` and `read_csv_arrow` to be private method, and `read_csv` with a `backend` kwarg to be the one exposed to the world.

JovanVeljanoski:
Hey @kaburelabs  you might find this interesting / helpful.

packages/vaex-core/vaex/__init__.py
```diff
@@ -101,41 +101,54 @@ def open(path, convert=False, shuffle=False, fs_options={}, *args, **kwargs):
...
+    Files are by default cached in $HOME/.vaex/file-cache/(s3|gs) such that successive access
+    is as fast as native disk access.
+
+    The following common fs_options are used for S3 access:
 
      * anon: Use anonymous access or not (false by default). (Allowed values are: true,True,1,false,False,0)
-     * use_cache: Use the disk cache or not, only set to false if the data should be accessed once. (Allowed values are: true,True,1,false,False,0)
-     * profile and other arguments are passed to :py:class:`s3fs.core.S3FileSystem`
+     * cache: Use the disk cache or not, only set to false if the data should be accessed once. (Allowed values are: true,True,1,false,False,0)
+     * profile and other arguments are passed to `pyarrow.fs.S3FileSystem` or :py:class:`s3fs.core.S3FileSystem`.
```
maartenbreddels, line 129:
```suggestion
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -101,41 +101,54 @@ def open(path, convert=False, shuffle=False, fs_options={}, *args, **kwargs):
...
 
-    All arguments can also be passed as kwargs, but then arguments such as `anon` can only be a boolean, not a string.
+    All fs_options can also be encoded in the file path as a query string.
 
     Examples:
 
+    >>> df = vaex.open('s3://vaex/taxi/yellow_taxi_2015_f32s.hdf5', fs_options={'anon': True})
```
maartenbreddels, line 135:
```suggestion
    >>> df = vaex.open('s3://vaex/taxi/yellow_taxi_2015_f32s.hdf5', fs_options={'anonymous': True})
```

packages/vaex-core/vaex/__init__.py
```diff
@@ -196,8 +209,22 @@ def open(path, convert=False, shuffle=False, fs_options={}, *args, **kwargs):
...
+                        df = from_csv(path, convert=csv_convert, **kwargs)
+                    else:
+                        ds = vaex.dataset.open(path, fs_options=fs_options, *args, **kwargs)
+                        if ds is not None:
+                            df = vaex.from_dataset(ds)
+                        if convert and ds is not None:
+                            df = vaex.from_dataset(ds)
+                            df.export_hdf5(filename_hdf5, shuffle=shuffle)
+                            df = vaex.open(filename_hdf5)
+>>>>>>> chore(core): Remove copy_index as a defined kwarg from vaex.open
```
maartenbreddels, line 227:
```suggestion
```
oops, my bad :)

packages/vaex-core/vaex/__init__.py
```diff
@@ -196,8 +209,22 @@ def open(path, convert=False, shuffle=False, fs_options={}, *args, **kwargs):
                     )
                     ds = vaex.dataset.open(path_output, fs_options=fs_options)
                 else:
+<<<<<<< HEAD
```
maartenbreddels, line 212:
```suggestion
```

maartenbreddels (commented):
Awesome, since we now standardized the s3 argument, we should document them ourselves, hope you like my suggestions.

PR #1049, opened 2020-11-09
Docs: Improve i/o example

JovanVeljanoski:
Improvements to the I/O example to keep in track with the most recently developments in Vaex.


docs/source/example_io.ipynb
```diff
@@ -639,6 +602,197 @@
...
+    "For example: imagine that we have a file hosted on S3 that has 100 columns and 1 billion rows. Getting a preview of the DataFrame via `print(df)` for instance will download only the first and last 5 rows. If we then proceed to make calculations or plots with only 5 columns, only the data from those columns will be downloaded and cached to the local machine.\n",
+    "\n",
+    "By default, the data streamed from S3 and GCS is cached at `$HOME/.vaex/file-cache/s3` and `$HOME/.vaex/file-cache/gs` respectively, and thus successive access is as fast as native disk access. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Streaming Apache Arrow and Apache Parquet is just as simple. Caching is not yet available for these file formats yet, so the required data will be streamed from cloud each time an operation is executed. For applications that use such files, we suggest to place the compute instance in the same region as the storage bucket. \n",
```
maartenbreddels, line 732:
```suggestion
    "Streaming Apache Arrow and Apache Parquet is just as simple. Caching is available for these file formats, but using the Apache Arrow format will currently read all the data when opening the file, so less useful. For maximum performance, we always advise to use a compute instance at the same region as the bucket. \n",
```

docs/source/example_io.ipynb
```diff
@@ -1256,6 +1571,52 @@
     "The `df.export_csv` method is using `pandas_df.to_csv` behind the scenes, and thus one can pass any argument to `df.export_csv` as would to `pandas_df.to_csv`. The data is exported in chunks and the size of those chunks can be specified by the `chunk_size` argument in `df.export_csv`. In this way, data that is too large to fit in RAM can be saved to disk."
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Export to multiple files in parallel\n",
+    "\n",
+    "With the `export_many` method one can export a DataFrame to muliple files of the same type in parallel. This is likely to be more performant when exporting very large DataFrames compared to writing a single large Arrow of Parquet file, where each chunk is written in succession. The method also accepts the `fs_options` dictonary, and can be particularly convenient when exporting to cloud storage."
```
maartenbreddels, line 1580:
```suggestion
    "With the `export_many` method one can export a DataFrame to muliple files of the same type in parallel. This is likely to be more performant when exporting very large DataFrames to the cloud compared to writing a single large Arrow of Parquet file, where each chunk is written in succession. The method also accepts the `fs_options` dictonary, and can be particularly convenient when exporting to cloud storage."
```

maartenbreddels (commented):
Some small suggestions, hope you like them

Issue #1000, opened 2020-10-05
Casting PyArrow String Column to 'str' Returns 'object' Column w/ Wrong Values


kmcentush:
**Description**
See title. In short, if you try to cast an existing string column to a 'str' datatype, the result is an object column that doesn't convert properly.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`:
```
{'vaex': '3.0.0',
 'vaex-core': '4.0.0a3',
 'vaex-viz': '0.4.0',
 'vaex-hdf5': '0.6.0',
 'vaex-server': '0.3.1',
 'vaex-astro': '0.7.0',
 'vaex-jupyter': '0.5.2',
 'vaex-ml': '0.9.0',
 'vaex-arrow': '0.5.1'}
```
 - Vaex was installed via: pip
 - OS: Windows 10

**Additional information**
Here's a code example:
```
import pyarrow as pa
import vaex

dt = vaex.from_arrays(a=['test', None])
print(dt['a'].dtype == pa.string())  # is a PyArrow string
print(dt)

dt['b'] = dt['a'].astype('str')
print(dt['b'].dtype == 'O')  # is an object
print(dt)

True
  #  a
  0  'test'
  1  --
True
  #  a       b
  0  'test'  None
  1  --      None
```

JovanVeljanoski:
Hi, 

I am not 100% sure here (@maartenbreddels clarify better), but I think string types in vaex is meant to be arrow, and is not mean to be converted something else.

Is there any particular reason you are doing this conversion?

In any case, this behaviour is fixed in the new alpha ( i don't see a problem on master for example). 

Although I am surprised to see that a string type cast to string type is treated as an object on some level (@maartenbreddels )


<img width="601" alt="image" src="https://user-images.githubusercontent.com/18574951/95046439-7c245c80-06e4-11eb-9e60-b65ff8b9400a.png">

kmcentush:
Hi @JovanVeljanoski,

I'm on vaex-core 4.0.0a3 and seeing this issue. What versions are your other vaex libraries? See my original post for more details on the specific versions I'm running.

In my specific use case, I'm writing an API. Given that some results may include None, etc, I'm forcing a type conversion so that downstream code works. My actual intention here is to convert the column to whatever string type is natively used by vaex. However, in attempting to do so by using "str", I ran into this bug.

[...]

PR #1050, opened 2020-11-10
refactor!: Vaex DType wraps NumPy and Arrow DType in uniform way

maartenbreddels:
See the docstring of datatype.py for usage.
We should decide if we want this or not, and if so, we need to put this also in the docs:

TODO:
 * Add `py.test --doctest-modules packages/vaex-core/vaex/datatype.py` to CI
 * add `packages/vaex-core/vaex/datatype_test.py` to CI
 * Review naming

### DIFF
--- a/packages/vaex-core/vaex/datatype.py
+++ b/packages/vaex-core/vaex/datatype.py
@@ -245,10 +245,6 @@ class DataType:
         '''
         return vaex.array_types.is_string_type(self.internal)
 
-    # @property
-    # def is_str(self):
-    #     return self.is_string
-
     def upcast(self):
         '''Cast to the higest data type matching the type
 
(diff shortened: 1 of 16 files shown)


kmcentush:
+1, as I believe this would help to address the issues I noticed in https://github.com/vaexio/vaex/issues/1000

maartenbreddels:
Thanks for the feedback @kmcentush , glad you like it.

packages/vaex-core/vaex/datatype.py
```diff
@@ -0,0 +1,270 @@
...
+        >>> DataType(pa.large_string()).is_string
+        True
+        >>> DataType(pa.large_string()) == 'string'
+        True
+        '''
+        return vaex.array_types.is_string_type(self.internal)
+
+    # @property
+    # def is_str(self):
+    #     return self.is_string
```
JovanVeljanoski, line 250: Can this be removed?

JovanVeljanoski (approved):
Looks great! 

Did some light testing, runs as expected 👍 
Loving it <3

Issue #1023, opened 2020-10-23
[BUG-REPORT] Doc suggestion - note that if a file is _already_ converted, it won't be converted again


ianozsvald:
Here I'd assumed that if I played around with `from_csv` and changed the arguments, I'd get updated versions of the converted file: https://vaex.readthedocs.io/en/latest/api.html#vaex.from_csv

The code however checks to see if the converted result exists and if so, it silently exits without updating the file: https://github.com/vaexio/vaex/blob/master/packages/vaex-core/vaex/__init__.py#L549

For reference I'm working with a CSV file that lacks headers `df = vaex.from_csv(RAW_DATA, copy_index=False, chunk_size=None, convert=True, parse_dates=['date'], names=COLUMNS) ` so my first pass lacked the `names` column, then I couldn't understand _why_ I didn't head headers on the next pass.

I'd suggest it might be clearer in the documentation that the file is converted once and ignored on subsequent calls. I've just spent 20 mins wandering the internals of Vaex before I realised what was going on :-)

ianozsvald:
On reflection I'll suggest that if a user calls this function twice, they'll expect their data to be reprocessed. Possibly providing a flag to avoid this is sensible or maybe raise a "File Already Exists" exception (just as `os.makedirs` does if the target folder exists), but silently exiting without telling the user that the operation was short-circuited may confuse other new users. Thoughts?

JovanVeljanoski:
Hi @ianozsvald 

Thanks for raising this. It is a good point. I do think that both the documentation and perhaps the function itself needs to change a bit. I was made in a different time and with a specific use-case in mind a think. 

We are having an internal discussion how to proceed. Any ideas / suggestions are welcome. 

I hope to have this improved in the next stable release.

ianozsvald:
Perhaps a "from" function should just open what's there (maybe with a warning for non-performant choices) and a "convert" function should explicitly do the conversion?
I think guiding the first-time user from "here's how you get started, it isn't optimal but it works" to "and now we can drop a converted hdf5 file next to your original data" without surprises (i.e. clearer docs and a function that doesn't contain a caching surprise) will ease new people into Vaex. 
Thanks for following up :-)

maartenbreddels:
Hi Ian,

I agree with @JovanVeljanoski we should have better docs for this. I think thing, that if we cache, we should cache based on the argument of the functions and the mtime (time of last modification) of the file, and provide a `convert_force=True` override or something.

Regards,

Maarten

[...]

Issue #1054, opened 2020-11-14
[Bug report] Allow to specify custom path to store converted csv->hdf5 files


lrq3000:
When using `vaex.from_csv('somefile.csv', convert=True)`, the converted hdf5 file will always be stored in the same folder as the csv file. This may be inadequate if the user wants to keep separate the raw original data from the postprocessed hdf5 files. There is also the use case where the user wants to store the converted hdf5 files on another disk than the raw csv data (maybe to have more storage space?).

The argument `convert` accepts a path as a `str` instead of a boolean value (see [here](https://github.com/vaexio/vaex/blob/3a0785cf9265cec615c470172c8cd460c3bd17a8/packages/vaex-core/vaex/__init__.py#L538)) but it seems it's only meant to replace the `filename_or_buffer` argument (so I'm not sure the purpose?). It would be maybe more useful if the path would actually determine the folder where the converted files are meant to be stored to?

JovanVeljanoski:
Hi @lrq3000 

The behaviour your describe is exactly the behaviour that was intended. It looks like a bug indeed. I also found a related bug in `vaex.open(...)`. I will open a PR addressing this shortly. 
Thank you!

(let's keep this open until this is solved).

lrq3000:
That is a crazingly fast reply and PR! Thank you for having fixed it already!

byaminov:
Oh indeed, thanks for reporting! This workaround should help until the fix is released:
```
with open('somefile.csv') as f:
    vaex.from_csv(f, convert='/mnt/other/folder/converted.hdf5')
```

maartenbreddels:
This should be fixed in #1055, if not please reopen.

Released, to install:
```
$ pip install "vaex-core>=4.0.0a11"
# or for the while kitchsink:
$ pip install "vaex>=4.0.0a6"
```

PR #1055, opened 2020-11-14
Fix: correct conver behaviour in vaex.open and vaex.from_csv

JovanVeljanoski:
This addresses the issue raised in #1054: when `convert` contains a `str` it should act as a path for the converted file. This was ignored prior to this fix.


maartenbreddels:
Could you check why this is not covered in tests/open_test.py? I thought this was treated thoroughly.

JovanVeljanoski:
I got it. It was tested but not thoroughly enough. 
In the test when `convert` was a path, it is pointed to the same location where the source files are located. And that is where the converted file is saved.  And this is exactly the bug: if `convert` is string, the path was ignored, and the location of the source files is also the destination converted files (Just as if `convert` was `True`). 

I will update the test to reflect this (so that if would fail on current master for example).

JovanVeljanoski:
I think i got all cases. There was one problem when a single csv was converted in chunks - now this is tested as well. Should be fine I think, but please take a look.

packages/vaex-core/vaex/__init__.py
```diff
@@ -536,8 +541,7 @@ def iterator():
 
 
 def _from_csv_convert_and_read(filename_or_buffer, copy_index, maybe_convert_path, chunk_size, **kwargs):
-    # figure out the CSV file path
-    if isinstance(filename_or_buffer, str):
+    if isinstance(filename_or_buffer, str) & (maybe_convert_path is True):
```
byaminov, line 544: Maybe changing the order of checks would be a bit simpler to read, but LMK if you like it:
```
1. if `maybe_convert_path` is string => we know we have to use that as the file path
2. else if `filename_or_buffer` is a file path => use that
3. else fail.
```
Now the `maybe_convert_path is True` check works because we know that with the current code the `False` value would not be passed, but I'd rather not have to think about it.

byaminov (approved):
Thanks for fixing this @JovanVeljanoski ! Approved with one suggestion.

packages/vaex-core/vaex/__init__.py
JovanVeljanoski: Sounds good to me. 
Please push the improvement :)

byaminov: Argh, next time I'm not going to make suggestion then 😛 . Ok will do.

maartenbreddels: You can simply accept it Jovan

byaminov: Actually I'm not sure if I can update your PR. How can I?

byaminov: well I did not suggest it through a "Suggestion", so not sure if Jovan can. Let me try..

packages/vaex-core/vaex/__init__.py
```diff
@@ -536,8 +541,7 @@ def iterator():
...
         csv_path = filename_or_buffer
     elif isinstance(maybe_convert_path, str):
         csv_path = re.sub(r'\.hdf5$', '', str(maybe_convert_path), flags=re.IGNORECASE)
```
byaminov, line 547:
```suggestion
    if isinstance(maybe_convert_path, str):
        csv_path = re.sub(r'\.hdf5$', '', str(maybe_convert_path), flags=re.IGNORECASE)
    elif isinstance(filename_or_buffer, str):
        csv_path = filename_or_buffer
```

tests/open_test.py
```diff
@@ -40,15 +40,24 @@ def test_open():
     _cleanup_generated_files()
 
     # convert can also be a path
-    target = os.path.join(path, 'data', 'convert.hdf5')
+    target_path = os.path.join(path, 'data', 'output')
+    target = os.path.join(target_path, 'convert.hdf5')
```
byaminov, line 44: BTW there are more tests for CSV operations in the `from_csv_test.py`. But the test for the `convert=file_path` is passing a `filename_or_buffer=file_descriptor`, so it works :/

packages/vaex-core/vaex/__init__.py
JovanVeljanoski: Done.

JovanVeljanoski:
I forced-pushed because I rebased on master. Should be good to go now.

maartenbreddels:
I have no idea why this fails on appveyor, it cannot find the csv file all the sudden?

maartenbreddels:
I've refactored a bit, and added a mechanism such that if the input file changes, or the input parameters, the conversion is done again.

This fixes https://github.com/vaexio/vaex/issues/1023

JovanVeljanoski:
WIth the most recent refactor: if one uses `from_csv` or `read_csv`, and with the `convert=True` argument, but the original text file does not have an explicit `.csv` extension (which can  easily happen for ascii or tsv or others), an exception is being raise (since in those cases vaex will try to open such a file as if it was a memory mappable one).

maartenbreddels (approved):

Issue #827, opened 2020-06-12
When opening a bigger than memory CSV, using chunksize and converting to HDF5, the memory is still fully used until the machine hangs and the kernel dies.


nf78:
Following the recommendation here:
https://vaex.readthedocs.io/en/latest/faq.html

I have been able to verify that the small files (from each chunk) are being generated. 

However, although the data is being chunked and the small files are created, the memory usage increases until all memory is used, and the machine hangs until the kernel dies.

When reading the dataset using the chunksize argument, into pandas.read_csv(), the memory is kept as low as the chunksize. Perhaps the memory leak is happening in the process of conversion to hdf5?

To reproduce the issue, just try to read a bigger than memory csv file with this code (as recommended on the faq link):

df = vaex.from_csv('./my_data/my_big_file.csv', convert=True, chunk_size=5_000_000)

nf78:
Also, I want to add that, once all memory is fully used, no more small hdf5 files are created during conversion

byaminov:
@nf78 thanks for raising the issue!

I've tried converting 2.2Gb CSV file on a docker container limited to 1Gb memory, and it worked. I had to use smaller `chunk_size=100_000`.

However I could reproduce the issue when the CSV file contains non-number columns: the chunks are created successfully, but when trying to create one HDF5 file from the chunks the process crashes in that docker machine.

Does your CSV file also contain string columns? If not, can you try reducing the `chunk_size`? If yes, then it's probably the same issue that I could reproduce. We'll be looking into it further.

byaminov:
For the record, the command I used to start the Docker image:
```
docker run -it \
    --memory="1g" \
    --memory-swap="1g" \
    -v /my/datasets/test_yellow_tripdata:/data/ \
    python:3.7 /bin/bash
```

nf78:
Hi @byaminov ! Thanks for looking into this.

I should also update that my working environment is on an EC2 with 32GB and the dataset is a CSV with 54GB, running Ubuntu 18.04.2 LTS and python 3.6.9. The version of vaex is the latest available through pip install.

With regards to your questions:
a) Does your CSV file also contain string columns? 
     > Yes, it is mostly text.

b) If yes, then it's probably the same issue that I could reproduce. We'll be looking into it further.
      > Thank you

nf78:
[...]

PR #1056, opened 2020-11-14
Fix deletion of chunk files on csv->hdf5 conversion

lrq3000:
Use `df.close_files()` instead of `df.close()` to ensure the file lock on hdf5 chunks gets released and hence can be deleted.

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -575,7 +575,7 @@ def _from_csv_convert_and_read(filename_or_buffer, copy_index, maybe_convert_pat
         logger.info('deleting %d chunk files' % len(converted_paths))
         for df, df_path in zip(dfs, converted_paths):
             try:
-                df.close()
+                df.close_files()
                 os.remove(df_path)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


maartenbreddels:
close files is not used anymore, but close does not close the mmaped file on osx and Linux (see comment here https://github.com/vaexio/vaex/issues/827#issuecomment-727951692 )
If we do close them, it should have the effect you want, but it led to crashing CI in the past.

lrq3000:
Oh I see. It's weird because I am using vaex-core 2.0.3 and `close_files` is still an available method, I guess it was removed afterward? I'll close this issue then. BTW I am using Windows 10 64bit, and it was necessary for me to `close_files` in my code to be able to remove the filelock on hdf5 files opened with vaex.

maartenbreddels:
Indeed it is removed, i just pushed a change to fully implement close, lets see if it fails or not

maartenbreddels:
Great, I think previously I first closed the files, and then the mmaps. Great to finally solve this one. Thanks for bringing this up.

lrq3000:
Thank you for taking the time to implement a much cleaner fix than mine! :D

maartenbreddels:
Your timing was right 🙂

PR #1060, opened 2020-11-17
chore(ci): fix ci

maartenbreddels:


maartenbreddels:
Ubuntu 3.8 times out, I think it's due to compilation, and we did not see this before due to the cache.

Issue #1061, opened 2020-11-18
Unclear deprecation warning for {label, ordinal}_encode


pietrolesci:
https://github.com/vaexio/vaex/blob/f600ad06b9c1c05da309945cfc7496f6a8d6e87e/packages/vaex-core/vaex/dataframe.py#L5068

Hi all,

In the documentation for the DataFrame methods `label_encode` and `ordinal_encode` there is the warning "Deprecated: use is_category". However, it is unclear how this can be handled by the referenced method.

I hope this helps.

Best,
Pietro

JovanVeljanoski:
Hi, 

Yeah, looks like the deprecation message is wrong, should be "use ordinal_encode instead".

I don't know why the docs for `ordinal_encode` have this deprecation warning.. and I can't notice what in the code does his. I will take a closer look.

@maartenbreddels Do you think we can just remove `label_encode` in the next major version?

JovanVeljanoski:
Removing the `label_encode` method removes the deprecation warning in `ordinal_encode` for some reason..

pietrolesci:
Hi @JovanVeljanoski,

Thanks a lot for your quick feedback on that. 
I will use `ordinal_encode` then :)

Best,
Pietro

PR #1062, opened 2020-11-18
Chore: remove deprecated method (fixes docs)

JovanVeljanoski:
Proposal: since we are soon moving to a new major version, I propose we remove some of the deprecated methods, like `label_encode`, especially since there are methods superseding them (`ordinal_encode` in this case). 

This also fixes the documentation issue raised in #1061


maartenbreddels:
I don't think we should ever remove anything if it does add maintenance burden, I suggest we just correct the message.

PR #1066, opened 2020-11-20
fix: support alias in hdf5 files

maartenbreddels:


maartenbreddels:
duplicate, not needed anymore

PR #1078, opened 2020-11-24
feat: add export_feather function to allow compressed arrow files

picoDoc:
So this PR is part suggestion and part question 😄 

I recently put together a [comparison of vaex and kdb+ for finance type analyses](https://www.aquaq.co.uk/q/comparing-columnar-data-formats-arrow-vaex-and-kdb/) - vaex comes out pretty well considering the difference is maturity - and one of the more important features vaex is currently lacking for this kind of workload is compression.  

[Feather (v2)](https://ursalabs.org/blog/2020-feather-v2/) adds compression support to arrow files, but otherwise files saved as feather seem identical to the arrow IPC format, and so work identically when opened in vaex (or appear to do so).  I've added a simple `export_feather` function here that allows for saving compressed arrow files.  So far as I can tell feather is currently the only way to do this (I think [ARROW-8470](https://issues.apache.org/jira/browse/ARROW-8470) is related to possibly expanding this).  For example:

```
import pandas as pd
import numpy as np
import vaex

data = pd.DataFrame({
  "A":np.array([1,2,3,4]),
  "B":["A","B","C","D"],
  "C":[np.datetime64('2019-10-07 09:00:00'),np.datetime64('2019-10-07 10:00:00'),np.datetime64('2019-10-07 11:00:00'),np.datetime64('2019-10-07 12:00:00')]
  })

vaex.from_pandas(data).export_feather('example.arrow', compression='zstd')

data = vaex.open('example.arrow')

>>> data
  #    A  B    C
  0    1  'A'  Timestamp('2019-10-07 09:00:00')
  1    2  'B'  Timestamp('2019-10-07 10:00:00')
  2    3  'C'  Timestamp('2019-10-07 11:00:00')
  3    4  'D'  Timestamp('2019-10-07 12:00:00')
```

Performance in general seems as fast as standard arrow files, although one thing I've noticed using larger files (in the GB range) is that the memory usage of the process goes up when opening the compressed file (before running any queries) so it's possible it's cheating.  Do you think this is a limitation of using compression, that it needs to load more into memory rather than just memory mapping?

Like I said I'm raising this PR as much to get your opinions as to get this actually merged (although it would be cool to get some form of this functionality in vaex!)

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5920,7 +5920,7 @@ class DataFrameLocal(DataFrame):
             write(to)
 
     @docsubst
-    def export_feather(self, to, parallel=True, reduce_large=True, compression='uncompressed'):
+    def export_feather(self, to, parallel=True, reduce_large=True, compression='lz4', fs_options=None):
         """Exports the DataFrame to an arrow file using the feather file format version 2
 
         Feather is exactly represented as the Arrow IPC file format on disk, but also support compression.
@@ -5929,12 +5929,15 @@ class DataFrameLocal(DataFrame):
         :param to: filename or file object
         :param bool parallel: {evaluate_parallel}
         :param bool reduce_large: If True, convert arrow large_string type to string type
-        :param to: Can be one of 'zstd', 'lz4' or 'uncompressed'
+        :param compression: Can be one of 'zstd', 'lz4' or 'uncompressed'
+        :param fs_options: Placeholder for now
         :return:
         """
         import pyarrow.feather as feather
         table = self.to_arrow_table(parallel=False, reduce_large=reduce_large)
-        feather.write_feather(table, to, compression=compression)
+        fs_options = fs_options or {}
+        with vaex.file.open(path=to, mode='wb', fs_options=fs_options) as sink:
+            feather.write_feather(table, sink, compression=compression)
 
     @docsubst
     def export_parquet(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, fs_options=None, fs=None, **kwargs):
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5920,6 +5920,23 @@ class DataFrameLocal(DataFrame):
             write(to)
 
     @docsubst
+    def export_feather(self, to, parallel=True, reduce_large=True, compression='uncompressed'):
+        """Exports the DataFrame to an arrow file using the feather file format version 2
+
+        Feather is exactly represented as the Arrow IPC file format on disk, but also support compression.
+            see also https://arrow.apache.org/docs/python/feather.html
+
+        :param to: filename or file object
+        :param bool parallel: {evaluate_parallel}
+        :param bool reduce_large: If True, convert arrow large_string type to string type
+        :param to: Can be one of 'zstd', 'lz4' or 'uncompressed'
+        :return:
+        """
+        import pyarrow.feather as feather
+        table = self.to_arrow_table(parallel=False, reduce_large=reduce_large)
+        feather.write_feather(table, to, compression=compression)
+
+    @docsubst
     def export_parquet(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, fs_options=None, fs=None, **kwargs):
         """Exports the DataFrame to a parquet file.
 
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5922,7 +5922,7 @@ class DataFrameLocal(DataFrame):
             write(to)
 
     @docsubst
-    def export_feather(self, to, parallel=True, reduce_large=True, compression='lz4', fs_options=None):
+    def export_feather(self, to, parallel=True, reduce_large=True, compression='lz4', fs_options=None, fs=None):
         """Exports the DataFrame to an arrow file using the feather file format version 2
 
         Feather is exactly represented as the Arrow IPC file format on disk, but also support compression.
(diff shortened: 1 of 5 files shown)


JovanVeljanoski:
Hi @picoDoc 

This looks interesting for sure! We would be happy to support this if we can get it onto the right shape. Let me try and look/test at some point today/tomorrow and will try to provide some feedback. 

Talking about compression, vaex has out-of-core support for parquet files, although I have no idea how it compares to feather.

maartenbreddels:
Hi Matt,

> I recently put together a [comparison of vaex and kdb+ for finance type analyses](https://www.aquaq.co.uk/q/comparing-columnar-data-formats-arrow-vaex-and-kdb/) - vaex comes out pretty well considering the difference is maturity - and one of the more important features vaex is currently lacking for this kind of workload is compression.

took a while for me to give a proper reply, because I wanted to read your article properly first. And I like it! @JovanVeljanoski want to improve datetime support after 4.0, so your timing is perfect.

As Jovan said, we have support for parquet, but that comes at CPU cost.

As Wes McKinney says in https://ursalabs.org/blog/2020-feather-v2/ I think Feather is attractive for data that does not fit into memory, and where there is an IO bottleneck (say a slow RAID array, or a single nvme drive). For nvme raid arrays I think hdf5 or arrow is still best.

The problem is (or was) that these files would probably use memory (as you mentioned), and I did not know of a way to use the Arrow API to scan over the file https://issues.apache.org/jira/browse/ARROW-10736 but as @jorisvandenbossche kindly pointed out, we should be able to do that.

But even if we can't support .feather like we do parquet, having an export method is certainly on our wishlist (I'll review now).

cheers,

Maarten

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5894,6 +5894,23 @@ def write(writer):
         else:
             write(to)
 
+    @docsubst
+    def export_feather(self, to, parallel=True, reduce_large=True, compression='uncompressed'):
```
maartenbreddels, line 5898:
```suggestion
    def export_feather(self, to, parallel=True, reduce_large=True, compression='lz4'):
```

Given the results here https://ursalabs.org/blog/2020-feather-v2/ and the use case (slightly bigger than parquet, but faster in reading/writing), I think this makes for a better default

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5894,6 +5894,23 @@ def write(writer):
...
+
+        :param to: filename or file object
+        :param bool parallel: {evaluate_parallel}
+        :param bool reduce_large: If True, convert arrow large_string type to string type
+        :param to: Can be one of 'zstd', 'lz4' or 'uncompressed'
+        :return:
+        """
+        import pyarrow.feather as feather
+        table = self.to_arrow_table(parallel=False, reduce_large=reduce_large)
+        feather.write_feather(table, to, compression=compression)
```
maartenbreddels, line 5912: Too bad we have to put this in memory, but it doesn't seem like there is a streaming way to get the data in feather from https://arrow.apache.org/docs/python/api/formats.html

packages/vaex-core/vaex/dataframe.py
maartenbreddels: Would be great if you could mirror as much as possible the features of `export_arrow`, for instance, the fs_options, and how we open the file. That gives us the option to write to cloud storage.

maartenbreddels (changes requested):
Looks good, do you want to continue to work on this? 
For instance, we can put some feather files on s3 so we can run https://github.com/vaexio/vaex/blob/628c03e7069d9c194586264cda16f1d7c94bcc06/tests/cloud_dataset_test.py#L24 on it.
Also, adding an entry in https://github.com/vaexio/vaex/blob/628c03e7069d9c194586264cda16f1d7c94bcc06/packages/vaex-core/setup.py#L154
and some boilerplate to read in the data would be nice.

picoDoc:
Hey, sorry I'm slow replying but yes I'm happy to continue working on this.  Will try to find time in the next few days.

packages/vaex-core/vaex/dataframe.py
picoDoc: not yet anyway!

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5894,6 +5894,26 @@ def write(writer):
...
+        :param bool reduce_large: If True, convert arrow large_string type to string type
+        :param compression: Can be one of 'zstd', 'lz4' or 'uncompressed'
+        :param fs_options: Placeholder for now
+        :return:
+        """
+        import pyarrow.feather as feather
+        table = self.to_arrow_table(parallel=False, reduce_large=reduce_large)
+        fs_options = fs_options or {}
+        with vaex.file.open(path=to, mode='wb', fs_options=fs_options) as sink:
+            feather.write_feather(table, sink, compression=compression)
```
picoDoc, line 5915: I was pleasantly surprised that this seems to work.  In the [docs](https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html) it's only meant to be a string, but seems to support sinks just fine so we can write to s3 etc.

picoDoc:
I've added the cloud test, it just needs the appropriate feather file added to `gs://vaex-data` and `s3://vaex`, can you add it?  Are there any other places it would be useful to add tests for this?  export_test.py maybe?

That's great that we might potentially not have to load this all into memory, it would make this a lot more useful.  As you say it could be great for IO limited situations.

I've also added the boilerplate to open `*.feather` files, although because these are still arrow format we can save them as *.arrow` and that works fine too.

Any other changes or additions to get this ready to go?

packages/vaex-core/vaex/dataframe.py
maartenbreddels: Yeah, just tested that, and was able to write to s3.

maartenbreddels (commented):
Looking good, I'll push one more change, with export recognizing it, which should dispatch to export_parquet (for convenience).

maartenbreddels:
Thanks @picoDoc !

Issue #1074, opened 2020-11-24
How do I tell vaex to use multiple CPU cores?


aburkov:
I run the following code:

```
df = vaex.open('s3://...parquet')

df = df[["original_description"]]

extractor = Extractor()# an external class that has the "match" method that extracts something from a given text inptut.

def extractor_function(input):
    return json.dumps(extractor.match(input))

df["keywords"] = df.apply(extractor_function, arguments=[df.original_description])

df.export_csv('/home/username/output_data.csv')
```

While it's running, I run the top command and observe that only one python process is running and it consumes one CPU core at 100% on a machine that has 8 CPU cores. I also don't see an increase in processing speed as compared to my previous machine that only had 2 cores.

I tried to open three parquet files using open_many, but the processing time tripled and only one CPU core was still used.

Is there a way to "tell" vaex to use multiple CPU cores?

maartenbreddels:
If possible, vaex will use all cores. However, apply can never use more than 100% cpu, because of the GIL. I have thought of running apply in a process, possibly by using dask.
Were you expecting something like that to happen?

aburkov:
In my case, extractor_function is called to extract keywords from each text document in the dataframe. extractor_function uses the library hyperscan. hyperscan applies regular expressions to each document in a very efficient way: it can match 100,000 regexes in a fraction of a millisecond.

If apply is not the best way to do that, how can I transform a row in a custom way (by applying some external Python module) without using apply?

maartenbreddels:
The best way is to add a new function https://vaex.io/docs/tutorial.html#Adding-functions

instead of the apply, where each row will be fed into the Python function, this UDF will be called with an array (each thread will invoke it with a slice of the original column). If the function down the line accepts arrays (hyperscan), it should perform much better.

Let me know if you have trouble with this, happy to try this out myself. Hyperscan was on my list of libraries to use for the string regex in vaex, but we chose for PCRE instead.

If it does not accept an array, i may know of ways with numba to make that work.

aburkov:
I implemented it this way:

```
extractor = Extractor()
@vaex.register_function()
def extractor_function(input):
    return extractor.MatchVectorized(input)

df["skills"] = df.original_description.extractor_function()
```

The method MatchVectorized of the "extractor" object takes an Arrow StringArray as input, iterates over its values, and then returns a numpy array of strings as output. I know that looping over elements in StringArray is not optimal either, but I expected to gain because multiple cores would be used in parallel.

[...]

PR #1079, opened 2020-11-25
fix: chunk_size was ignored

maartenbreddels:
Related #1074 

confirmed it's working by running in a 1GB docker container:
```python
import vaex
import numpy as np

N = 1024*1024//8*500
x = np.ones(N)
df = vaex.from_arrays(x=x)
df['y'] = df.x*2
print(df.y.sum())
```
this will crash
`$ VAEX_CHUNK_SIZE=10000000 python /home/maartenbreddels/github/vaexio/vaex/memusage.py`
this will not
`$ VAEX_CHUNK_SIZE=10000 python /home/maartenbreddels/github/vaexio/vaex/memusage.py`


maartenbreddels:
The appveyor crash seem to be related to xgboost.

PR #1080, opened 2020-11-25
feat: apply with multiprocessing

maartenbreddels:
POC from #1074

### DIFF
--- a/docs/source/tutorial.ipynb
+++ b/docs/source/tutorial.ipynb
@@ -3687,6 +3687,294 @@
    "source": [
     "df.scale.mul(2)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# The escape hatch: apply\n",
+    "In case a calculation cannot be expressed as a Vaex expression, one can use the `apply` method as a last resort. This can be useful if the function you want to apply is written in pure Python, a third party library, and is difficult or impossible to vectorize.\n",
+    "\n",
+    "We think apply should only be used as a last resort, because it needs to use multiprocessing (which spawns new processes) to avoid the Python Global Interpreter Lock (GIL) to make use of multiple cores. This comes at a cost of having to transfer the data between the main and child processes.\n",
+    "\n",
+    "Here is an example which uses the `apply` method:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2020-11-27T18:59:23.880238Z",
+     "start_time": "2020-11-27T18:59:23.282738Z"
+    }
+   },
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<table>\n",
+       "<thead>\n",
+       "<tr><th>#                            </th><th style=\"text-align: right;\">  x</th><th>is_prime  </th></tr>\n",
+       "</thead>\n",
+       "<tbody>\n",
+       "<tr><td><i style='opacity: 0.6'>0</i></td><td style=\"text-align: right;\">  0</td><td>False     </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>1</i></td><td style=\"text-align: right;\">  1</td><td>False     </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>2</i></td><td style=\"text-align: right;\">  2</td><td>True      </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>3</i></td><td style=\"text-align: right;\">  3</td><td>True      </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>4</i></td><td style=\"text-align: right;\">  4</td><td>False     </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>5</i></td><td style=\"text-align: right;\">  5</td><td>True      </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>6</i></td><td style=\"text-align: right;\">  6</td><td>False     </td></tr>\n",
+       "<tr><td><i style='opacity: 0.6'>7</i></td><td style=\"text-align: right;\">  7</td><td>True      </td></tr>\n",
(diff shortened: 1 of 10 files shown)


docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# The escape hatch: apply\n",
+    "In case a calculation cannot be expressed in a Vaex expression, the last resort method is to use apply. This can be useful if the function you want to apply is written in pure Python, and difficult or impossible to vectorize.\n",
+    "\n",
+    "We think apply should only be used as a last resort, because it needs to use multiprocessing (which spawns new processes) to avoid the Python Global Interpreter Lock (GIL) to make use of multiple cores. This comes at a cost of having to transfer the data between the main and child processes.\n",
+    "\n",
+    "In case you really want, or need to use it, here is how you use it:"
```
JovanVeljanoski, line 3700:
```suggestion
    "Here is an example which uses the `apply` method:"
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+    }
+   ],
+   "source": [
+    "import vaex\n",
+    "\n",
+    "def slow_is_prime(x):\n",
+    "    return x > 1 and all((x % i) != 0 for i in range(2, x))\n",
+    "\n",
+    "df = vaex.from_arrays(x=vaex.vrange(0, 100_000, dtype='i4'))\n",
+    "# you need to explicitly tell which arguments you need\n",
```
JovanVeljanoski, line 3760:
```suggestion
    "# you need to explicitly specify which arguments you need\n",
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "There are 9592 prime number between 0 and 100000\n"
+     ]
+    }
+   ],
+   "source": [
+    "prime_count = df.is_prime.sum()\n",
+    "print(f'There are {prime_count} prime number between 0 and {len(df)}')"
```
JovanVeljanoski, line 3785:
```suggestion
    "print(f'There are {prime_count} prime numbers between 0 and {len(df)}')"
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+    "df['is_prime'] = df.apply(slow_is_prime, arguments=[df.x])\n",
+    "# but this form only works for a single argument\n",
+    "df['is_prime'] = df.x.apply(slow_is_prime)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## When not to use apply\n",
```
JovanVeljanoski, line 3809:
```suggestion
    "## When not to use `apply`\n",
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+    "When your function can easily be vectorized, you should not use apply. When you use Vaex' expression system, we know what you do, we see the expression, and can manipulate it. An apply function is like a black box, we cannot do anything with it, like JIT-ting for instance."
```
JovanVeljanoski, line 3810:
```suggestion
    "You should not use `apply` when your function can be vectorized. When you use Vaex' expression system, we know what you do, we see the expression, and can manipulate it in order to achieve optimal performance. An `apply` function is like a black box, we cannot do anything with it, like JIT-ting for instance."
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+   "execution_count": 11,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2020-11-27T18:59:37.287763Z",
+     "start_time": "2020-11-27T18:59:37.286036Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# bad idea: same performance as just dy['y'], but we lose the information about what you have done\n",
```
JovanVeljanoski, line 3952:
```suggestion
    "# bad idea: same performance as just dy['y'], but we lose the information about what was done\n",
```

docs/source/tutorial.ipynb
```diff
@@ -3687,6 +3687,294 @@
...
+   "execution_count": 7,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2020-11-27T18:59:27.623180Z",
+     "start_time": "2020-11-27T18:59:27.621623Z"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# will transfer the data to child processes, and execute the ** in Python it for each element\n",
```
JovanVeljanoski, line 3876:
```suggestion
    "# will transfer the data to child processes, and execute the **  operation in Python for each element\n",
```

packages/vaex-core/vaex/dataframe.py
```diff
@@ -384,6 +384,7 @@ def apply(self, f, arguments=None, dtype=None, delay=False, vectorize=False):
 
         :param f: The function to be applied
         :param arguments: List of arguments to be passed on to the function f.
+        :param bool multiprocessing: Use multiple processes to avoid the GIL (Global interpreter lock)
```
JovanVeljanoski, line 387: Should we take this opportunity to document the rest of the kwargs here?

packages/vaex-core/vaex/expression.py
```diff
@@ -856,7 +856,7 @@ def apply(self, f):
         :param f: A function to be applied on the Expression values
```
JovanVeljanoski, line 856: Should we document all kwargs here?

docs/source/tutorial.ipynb
maartenbreddels: this never format correctly with nbsphinx

packages/vaex-core/vaex/dataframe.py
maartenbreddels: yeah, vectorize at least

packages/vaex-core/vaex/expression.py
maartenbreddels: yeah

PR #1088, opened 2020-11-30
feat: join improvements for high cardinality data and particular data.

maartenbreddels:
With particular data, we mean data (primotives) for which the lower
bits don't change much.
See https://tessil.github.io/hopscotch-map/#growth-policy

In those cases (since we use an identify hash function), it is better
to have a prime growth strategy.

Also, based on the cardinality, we use less hash maps.


maartenbreddels:
Particular helpful on Gaia data:
image

Almost 100x faster

maartenbreddels:
At a small performance (5-10%) we can have a hash function that is non-identity that performs better in these edge cases. It might be good to keep in the prime growth strategy. 
I was sidetracked because apparently also boost::hash gives the identity function for primitives (at least int64!)

maartenbreddels:
The memory usage on the Gaia data is excessive, but doable with the new hash function and passing `cardinality=len(df)`.

PR #1090, opened 2020-12-01
feat: DataFrame pickle support for Dask, Ray and multiprocessing.

maartenbreddels:
Note that this not means you get free distributed computing, but it's possible to very efficiently pass around a vaex dataframes in Dask, Ray or multiprocessing (which use pickle).

Pickling `hdf5`, `parquet` and 'arrow ipc' backed dataframes pass the file path around, which means it is quite fast. CSV backed files currently pass around the data, which is slower (This will change).

# Dask example
```
# terminal 1
$ dask-scheduler --host localhost
# terminal 2:
# note that vaex does its own multi threading, so in this case we have 2 processes with 1 threads, and vaex will swap 16 threads in each of those processes.
$ VAEX_NUM_THREADS=16 dask-worker localhost:8786 --nthreads=1 --nprocs=2 --memory-limit=100G
```

Dask complains about memory usage, but that's all memory mapped memory it is reading, so we pass `--memory-limit=100G`.

```python
import dask
import vaex
from dask.distributed import Client
client = Client('127.0.0.1:8786')


@dask.delayed
def partial_sum(df):
    sub_total = df.total_amount.sum()
    return sub_total


if __name__ == '__main__':
    df = vaex.open('/data/taxi/yellow_taxi_2009_2015_f32_app.hdf5')
    N = len(df)
    print(f'{N:,}')
    print(df.get_column_names())

    df1, df2 = df[:N//2], df[N//2:]
    for i in range(10):
        total = (partial_sum(df1) + partial_sum(df2)).compute()
        print(total)

```

### DIFF
--- a/packages/vaex-core/src/agg.hpp
+++ b/packages/vaex-core/src/agg.hpp
@@ -188,6 +188,10 @@ public:
     void set_data(StringSequence* string_sequence, size_t index) {
         this->string_sequence = string_sequence;
     }
+    void clear_data_mask() {
+        this->data_mask_ptr = nullptr;
+        this->data_mask_size = 0;
+    }
     void set_data_mask(py::buffer ar) {
         py::buffer_info info = ar.request();
         if(info.ndim != 1) {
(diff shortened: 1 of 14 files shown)


maartenbreddels:
@aburkov you might be interested in this (pickling support), just released:
```
pip install vaex-core==4.0.0a10
or
pip install vaex==4.0.0a5
```

PR #1093, opened 2020-12-01
feat: df.split takes an integer, to split into n equal length dataframes

maartenbreddels:


packages/vaex-core/vaex/dataframe.py
```diff
@@ -4000,19 +3999,31 @@ def split(self, frac):
...
             # make sure it is normalized
-            total = sum(frac)
-            frac = [k / total for k in frac]
+            total = sum(into)
+            into = [k / total for k in into]
         else:
-            assert frac <= 1, "fraction should be <= 1"
-            frac = [frac, 1 - frac]
-        offsets = np.round(np.cumsum(frac) * len(self)).astype(np.int64)
+            assert into <= 1, "intotion should be <= 1"
```
JovanVeljanoski, line 4024:
```suggestion
            assert into <= 1, "when a float, `into` should be <= 1"
            assert into > 0, "`into` must be > 0."
```

packages/vaex-core/vaex/dataframe.py
```diff
@@ -4000,19 +3999,31 @@ def split(self, frac):
...
+                yield self[i1:i2]
+                i1, i2 = i2, i2 + step
+            return
+
+        if _issequence(into):
             # make sure it is normalized
-            total = sum(frac)
-            frac = [k / total for k in frac]
+            total = sum(into)
+            into = [k / total for k in into]
```
JovanVeljanoski, line 4022: If the user specifies a list of floats, the some of which is not 1, should we raise a warning notifying the user that the relative fractions will be normalized? 

I was expecting an error there, when I entered bunch of numbers whose sum was much greater than 1. 

(or just flat out ask for a sum of 1?).

packages/vaex-core/vaex/dataframe.py
```diff
@@ -4000,19 +3999,31 @@ def split(self, frac):
         [2 3 4]
         [5 6 7 8 9]
 
-        :param int/list frac: If int will split the DataFrame in two portions, the first of which will have size as specified by this parameter. If list, the generator will generate as many portions as elements in the list, where each element defines the relative fraction of that portion.
+        :param int, float or list of floats into: If float will split the DataFrame in two, the first of which will have a relative length as specified by this parameter.
```
JovanVeljanoski, line 4002:
```suggestion
        :param int, float or list of floats into: A float will split the DataFrame in two, the first of which will have a relative length as specified by this parameter.
```

JovanVeljanoski (changes requested):
Good PR! I like it :)

Can you please also propagate the changes to the `split_random` method? Maybe update the docs?

packages/vaex-core/vaex/dataframe.py
maartenbreddels: no, i think it's convenient like this.

JovanVeljanoski: it is convenient indeed, but i think a warning of the re-normalization (if it happens) is not a bad idea.

maartenbreddels: but how to get rid of the warning then? I mean, I think it's expected behavior.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -3956,18 +3955,20 @@ def split_random(self, frac, random_state=None):
...
         ...     print(dfs.x.values)
         [8 1]
         [5 0 7]
         [2 9 4 3 6]
 
-        :param int/list frac: If int will split the DataFrame in two portions, the first of which will have size as specified by this parameter. If list, the generator will generate as many portions as elements in the list, where each element defines the relative fraction of that portion.
+        :param int/float/list into: If float will split the DataFrame in two, the first of which will have a relative length as specified by this parameter.
+            When a list, will split into as many portions as elements in the list, where each element defines the relative length of that portion. Note that such a list of fractions will always be re-normalized to 1.
+            When an int, split DataFrame into n dataframes of equal length (last one may deviate), if len(df) < n, it will return len(df) DataFrames.
         :param int random_state: (default, None) Random number seed for reproducibility.
```
maartenbreddels, line 3972: Why did you edit this?

maartenbreddels: Got it!

maartenbreddels:
Thanks @JovanVeljanoski !

PR #1094, opened 2020-12-02
feat: Hive based partitioned datasets support. Read parquet, write any

### LINKED ISSUE #834: Vaex fails opening partitioned Parquet file
I have a local parquet files splitted in 28 partitions (~370M rows, 4.1Gb in snappy):
```
activities.parquet/
|___ activities-e502fb49-a1e7-49bd-8dbf-75468732cb45.snappy.parquet
|___ activities-d522fcf6-19f6-4855-8541-032bd5c0fbd5.snappy.parquet
|___ ...
```

I'm able to open it but `df.head()` with `vaex` raises:
```
ValueError: array is of length 13514680, while the length of the DataFrame is 370889861
```
I don't have any problem reading it with `arrow` or `pandas`.

I haven't investigated that much to be honest!

Vaex versions:
```
In [9]: vaex.__version__
Out[9]:
{'vaex': '3.0.0',
 'vaex-core': '2.0.3',
 'vaex-viz': '0.4.0',
 'vaex-hdf5': '0.6.0',
 'vaex-server': '0.3.1',
 'vaex-astro': '0.7.0',
 'vaex-jupyter': '0.5.2',
 'vaex-ml': '0.9.0',
 'vaex-arrow': '0.5.1'}
```
Tested under Python38 & Python37

Complete traceback:
```
ValueError                                Traceback (most recent call last)
<ipython-input-8-c42a15b2c7cf> in <module>
----> 1 df.head()

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in head(self, n)
   3449     def head(self, n=10):
   3450         """Return a shallow copy a DataFrame with the first n rows."""
-> 3451         return self[:min(n, len(self))]
   3452
   3453     def tail(self, n=10):

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in __getitem__(self, item)
   4624                 assert stop != -1
   4625                 stop = stop+1  # +1 to make it inclusive
-> 4626             df = self.trim()
   4627             df.set_active_range(start, stop)
   4628             return df.trim()

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in trim(self, inplace)
   3857         :rtype: DataFrame
   3858         '''
-> 3859         df = self if inplace else self.copy()
   3860         if self._index_start == 0 and self._index_end == self._length_original:
   3861             return df

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in copy(self, column_names, virtual)
   5034                 if name in self.columns:
   5035                     column = self.columns[name]
-> 5036                     df.add_column(name, column, dtype=self._dtypes_override.get(name))
   5037                     if isinstance(column, ColumnSparse):
   5038                         df._sparse_matrices[name] = self._sparse_matrices[name]

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in add_column(self, name, data, dtype)
   6051         #     self._length_original = len(data)
   6052         #     self._index_end = self._length_unfiltered
-> 6053         super(DataFrameArrays, self).add_column(name, data, dtype=dtype)
   6054         self._length_unfiltered = int(round(self._length_original * self._active_fraction))
   6055         # self.set_active_fraction(self._active_fraction)

/usr/local/lib/python3.8/site-packages/vaex/dataframe.py in add_column(self, name, f_or_array, dtype)
   2940                     if len(self) == len(ar):
   2941                         raise ValueError("Array is of length %s, while the length of the DataFrame is %s due to the filtering, the (unfiltered) length is %s." % (len(ar), len(self), self.length_unfiltered()))
-> 2942                 raise ValueError("array is of length %s, while the length of the DataFrame is %s" % (len(ar), self.length_original()))
   2943             # assert self.length_unfiltered() == len(data), "columns should be of equal length, length should be %d, while it is %d" % ( self.length_unfiltered(), len(data))
   2944             valid_name = vaex.utils.find_valid_name(name, used=self.get_column_names(hidden=True))

ValueError: array is of length 13514680, while the length of the DataFrame is 370889861
```

maartenbreddels:
Fixes #834

Similar to groupby, we can write those partitions to disk:
```python
import vaex
countries = ['US', 'US', 'NL', 'FR', 'NL', 'NL']
years = [2020, 2021, 2020, 2020, 2019, 2020]
values = [1, 2, 3, 4, 5, 6]
df = vaex.from_dict({
    'country': countries,
    'year': years,
    'value': values,
})
df.export_partitioned('./partitioned', by=['country', 'year'])
```
```
$ ls -R ./partitioned/
./partitioned/:
'country=FR'  'country=NL'  'country=US'

'./partitioned/country=FR':
'year=2020'

'./partitioned/country=FR/year=2020':
c543a46a-8188-4a0d-bd97-609c2adea36f.parquet

'./partitioned/country=NL':
'year=2019'  'year=2020'

'./partitioned/country=NL/year=2019':
545776eb-16dc-46d6-9882-fd68114b42be.parquet

'./partitioned/country=NL/year=2020':
c4211f02-10ef-499c-9175-5ffbcd274770.parquet

'./partitioned/country=US':
'year=2020'  'year=2021'

'./partitioned/country=US/year=2020':
e5d2ec3d-23f4-441e-aa7b-1a143c4c5b6b.parquet

'./partitioned/country=US/year=2021':
469bf3c2-7df7-4dad-b4b7-cc8f1cb7409a.parquet
```
If you don't like the uuid names:
```python
df.export_partitioned('./partitioned_fixed/{subdir}/vaex.parquet', by=['country', 'year'])
```

```
$ ls -R ./partitioned_fixed/
./partitioned_fixed/:
'country=FR'  'country=NL'  'country=US'

'./partitioned_fixed/country=FR':
'year=2020'

'./partitioned_fixed/country=FR/year=2020':
vaex.parquet

'./partitioned_fixed/country=NL':
'year=2019'  'year=2020'

'./partitioned_fixed/country=NL/year=2019':
vaex.parquet

'./partitioned_fixed/country=NL/year=2020':
vaex.parquet

'./partitioned_fixed/country=US':
'year=2020'  'year=2021'

'./partitioned_fixed/country=US/year=2020':
vaex.parquet

'./partitioned_fixed/country=US/year=2021':
vaex.parquet
```

Reading only supports Hive format, with parquet.
```
import vaex
df = vaex.open('./partitioned_fixed/')
print(df)
```
```
  #    value  country      year
  0        4  'FR'         2020
  1        5  'NL'         2019
  2        3  'NL'         2020
  3        6  'NL'         2020
  4        1  'US'         2020
  5        2  'US'         2021
```

Here country and year are not stored on disk, they get created on the fly.

### DIFF
--- a/packages/vaex-core/vaex/arrow/dataset.py
+++ b/packages/vaex-core/vaex/arrow/dataset.py
@@ -1,8 +1,11 @@
 __author__ = 'maartenbreddels'
+from collections import defaultdict
 import logging
 
+import numpy as np
 import pyarrow as pa
 import pyarrow.dataset
+import pyarrow.parquet as pq
 
 import vaex.dataset
 import vaex.file
@@ -33,6 +36,9 @@ class DatasetArrowBase(vaex.dataset.Dataset):
         self._row_count = row_count
         self._columns = {name: vaex.dataset.ColumnProxy(self, name, type) for name, type in
                           zip(self._arrow_ds.schema.names, self._arrow_ds.schema.types)}
+        for name, dictionary in self._partitions.items():
+            # TODO: make int32 dependant on the data?
+            self._columns[name] = vaex.dataset.ColumnProxy(self, name, pa.dictionary(pa.int32(), dictionary.type))
 
     def hashed(self):
         raise NotImplementedError
@@ -63,6 +69,17 @@ class DatasetArrowBase(vaex.dataset.Dataset):
         import pyarrow.parquet
         pool = get_main_io_pool()
         offset = 0
+        columns = tuple(columns)
+        columns_physical = []
+        columns_partition = []
+        for column in columns:
+            if column in self._partitions:
+                columns_partition.append(column)
+            else:
+                columns_physical.append(column)
+        columns_physical = tuple(columns_physical)
+        columns_partition = tuple(columns_partition)
+
         for fragment_large in self._arrow_ds.get_fragments():
             fragment_large_rows = sum([rg.num_rows for rg in fragment_large.row_groups])
             fragments = [fragment_large]
@@ -83,8 +100,15 @@ class DatasetArrowBase(vaex.dataset.Dataset):
                     # assert False
                     break
                 def reader(fragment=fragment):
-                    table = fragment.to_table(columns=columns, use_threads=False)
-                    chunks = dict(zip(table.column_names, table.columns))
+                    table = fragment.to_table(columns=list(columns_physical), use_threads=False)
+                    chunks_physical = dict(zip(table.column_names, table.columns))
+                    chunks_partition = {}
+                    partition_keys = self._partition_keys[fragment.path]
+                    for name in columns_partition:
+                        partition_index = partition_keys[name]
+                        partition_index_ar = pa.array(np.full(len(table), partition_index, dtype=np.int32))
+                        chunks_partition[name] = pa.DictionaryArray.from_arrays(partition_index_ar, self._partitions[name])
+                    chunks = {name: chunks_physical.get(name, chunks_partition.get(name)) for name in columns}
                     return chunks
                 assert length > 0
                 if start > chunk_start:
@@ -131,6 +155,16 @@ class DatasetArrowBase(vaex.dataset.Dataset):
         i1 = 0
         chunks_ready_list = []
         i1 = i2 = 0
+        # TODO: merge this with DatsetConcatenated.chunk_iterator
+        if not columns:
+            end = self.row_count if end is None else end
+            length = end - start
+            i2 = min(length, i1 + chunk_size)
+            while i1 < length:
+                yield i1, i2, {}
+                i1 = i2
+                i2 = min(length, i1 + chunk_size)
+            return
 
         workers = get_main_io_pool()._max_workers
         for chunks_future in buffer(self._chunk_producer(columns, chunk_size, start=start, end=end or self._row_count), workers+3):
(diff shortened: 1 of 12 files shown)


packages/vaex-core/vaex/arrow/dataset.py
```diff
@@ -151,19 +185,68 @@ def chunk_iterator(self, columns, chunk_size=None, reverse=False, start=0, end=N
...
-        file_system, path = vaex.file.parse(self.path, self.fs_options, for_arrow=True)
-        self._arrow_ds = pyarrow.dataset.dataset(path, filesystem=file_system)
+        file_system, source = vaex.file.parse(self.path, self.fs_options, for_arrow=True)
+        # we need to mix both arrow.parquet and arrow.dataset since the latter does not expose
+        # the partitions
+        if file_system is not None:
+            file_system_legacy = LegacyWrapper(file_system)
+        else:
+            file_system_legacy = LegacyWrapper(pa.fs.LocalFileSystem())
+        pds = pq.ParquetDataset(source,
```
jorisvandenbossche, line 235: Is there a specific reason (feature) for which you need to use the pq.ParquetDataset instead of `ds.dataset(..)` ? 

(since `pyarrow.dataset` is meant as a replacement, that's interesting to know)

maartenbreddels: Yes, I didn't open an issue on this yet for pyarrow. The partition information is not exposed AFAIK, so I cannot access the equivalent of `pds.partitions`.

jorisvandenbossche: It is available, but a bit hidden .. 
Each fragment has a `partition_expression`. Now, because the public API of those expression objects is limited, you can't really get the needed information out of this object. 
For that reason (and because dask also needs it), we added a private helper function for now:

```
In [1]: import pyarrow.dataset as ds

In [3]: dataset = ds.dataset("notebooks-arrow/nyc-taxi-data/dask-partitioned/", partitioning="hive")

In [4]: fragment = list(dataset.get_fragments())[0]

In [5]: fragment.partition_expression
Out[5]: <pyarrow.dataset.Expression ((year == 2016:int32) and (month == 10:int32))>

In [6]: ds._get_partition_keys(fragment.partition_expression)
Out[6]: {'month': 10, 'year': 2016}
```

It's "private-ish", mainly because we probably should come up with a better interface for this, but since advanced users of pyarrow.dataset like dask or vaex need it on the short term, it was added (and we can commit to keep it working)

jorisvandenbossche: (I know we need to improve this API .. any feedback on the information you need from the dataset is very welcome! We don't want to copy exactly the ParquetDataset API with its `pieces`, but at least we should somehow enable similar functionality)

maartenbreddels: That is awesome, I explored all those options but never got anywhere. Thanks a lot, Joris! I was ' ' this close to reaching out, but there is not low barrier way or reaching out, I find the mailing list a bit high barrier for some reason.

I kinda liked the ParquetDataset, with the dictionary, because now I have to assemble it myself.

jorisvandenbossche:
@maartenbreddels BTW, another question: assume that `ds.write_dataset(..)` could consume an iterator of tables/record batches lazily, and write those out in parallel, would that be of interest for vaex? 
(at the moment it requires a dense *list*, but we could improve that to also accept an iterable/generator).

I don't know if it would give you that much, though, as the `export_partitioned` in this PR is also not that much code / complicated

maartenbreddels:
I was thinking about this actually. And I wonder if we could have sth like lazy python-based datasets which could be fed back to arrow.

maartenbreddels:
Note that it's not a fast implementation, but a working one that does not require much memory, so certainly room for improvement.

jorisvandenbossche:
I opened https://issues.apache.org/jira/browse/ARROW-10882 for accepting iterators in `write_dataset()`, and for better exposing the partitioning information, I reopened https://issues.apache.org/jira/browse/ARROW-8655

PR #1099, opened 2020-12-04
refactor(hdf5): we now evaluate and write all columns in parallel

maartenbreddels:
We left the old code in place, but the new code is much easier to work
with, and is about 3-5x faster in some cases, and with many virtual
columns should be much faster.

### DIFF
--- a/benchmarks/export.py
+++ b/benchmarks/export.py
@@ -1,0 +1,36 @@
+import tempfile
+import os
+
+import numpy as np
+import vaex
+
+
+df = None
+
+def setup_df(N, M):
+    global df
+    x = [np.arange(N, dtype=np.float64) for _ in range(M)] 
+    df = vaex.from_dict({
+        f'c{i}': x[i] for i in range(M)
+    })
+
+def time_export_plain(N, M):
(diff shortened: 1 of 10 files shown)


maartenbreddels:
Writing the full gaia edr3 data seem to go from 3h to 25 minutes on Nyx.

maartenbreddels:
`$ asv run --python=$(which python) -b time_export_corr`

Before:
```
========== ============ ========== ============
              --                          M                  
              ---------- ------------------------------------
                  N           1           4           16     
              ========== ============ ========== ============
               1048576    77.0±0.9ms   413±20ms   3.38±0.02s 
               16777216   1.16±0.01s   6.06±0s      failed   
              ========== ============ ========== ============
```

After 
```
 ========== ============ ============ ============
              --                           M                   
              ---------- --------------------------------------
                  N           1            4            16     
              ========== ============ ============ ============
               1048576    18.3±0.5ms   45.4±0.2ms    186±1ms   
               16777216   194±0.5ms     431±9ms     1.39±0.02s 
              ========== ============ ============ ============

```

PR #1106, opened 2020-12-07
feat: support fsspec and arrow filesystem by passing around fs argument

maartenbreddels:
Instead of us creating the filesystem, we let uses create it, and pass
it down to vaex.

Example:
```python
import vaex
import fsspec
fs = fsspec.filesystem('https')
vaex.open('https://github.com/vaexio/vaex-datasets/releases/download/v1.0/helmi-dezeeuw-2000-FeH-v2-10percent.hdf5',
          fs=fs)
```

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -2273,7 +2273,7 @@ class DataFrame(object):
                 selection = selections.selection_from_dict(selection_dict)
             self.set_selection(selection, name=name)
 
-    def state_write(self, file, fs_options=None):
+    def state_write(self, file, fs_options=None, fs=None):
         """Write the internal state to a json or yaml file (see :meth:`DataFrame.state_get`)
 
         Example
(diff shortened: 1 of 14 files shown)


packages/vaex-core/vaex/dataframe.py
```diff
@@ -5850,7 +5849,7 @@ def mangle_name(prefix, name, suffix):
         return left
 
     @docsubst
-    def export(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, fs_options=None):
+    def export(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, fs_options=None, fs=None):
```
JovanVeljanoski, line 5852: we need to update the docstrings to all of these export methods i guess

packages/vaex-core/vaex/docstrings.py
```diff
@@ -38,4 +38,5 @@ def docsubst(f):
 _doc_snippets['evaluate_parallel'] = 'Evaluate the (virtual) columns in parallel'
 _doc_snippets['array_type'] = 'Type of output array, possible values are None/"numpy" (ndarray), "xarray" for a xarray.DataArray, or "list" for a Python list'
 _doc_snippets['ascii'] = 'Transform only ascii characters (usually faster).'
-_doc_snippets['fs_options'] = 'see :func:`vaex.open` e.g. for S3 {"profile": "myproject"}'
\ No newline at end of file
+_doc_snippets['fs_options'] = 'see :func:`vaex.open` e.g. for S3 {"profile": "myproject"}'
+_doc_snippets['fs_options'] = 'Pass a file system object directly, see :func:`vaex.open`'
```
JovanVeljanoski, line 42: I guess the 2nd one should be just fs?

maartenbreddels: indeed

PR #1135, opened 2020-12-28
Implement pandas like rename

### LINKED ISSUE #80: rename_column to excepts dict
should be `ds.rename({old_column:new_column})`

janakact:
Intended to fix #80


maartenbreddels:
Hi,

thanks for your first-time contribution! I like the idea of this, but I'm not sure of the method name. @JovanVeljanoski what do you think. Should we mayne rename the existing rename, or have 1 rename method that takes a dict or two arguments?

Regards,

Maarten

janakact:
Hi @maartenbreddels, Thank you for this marvelous library.
Yeah. Of course name should be changed I think, it should be similar to pandas I guess. May I add an  if condition to check whether the input is a dict to support both vaex and pandas API?

PR #1141, opened 2021-01-03
parse_expression: improve assert message

seregasheypak:
more context about  what exactly is "not an expression"


maartenbreddels:
Thanks!

PR #1147, opened 2021-01-07
fix: support for use with pandas 1.2

maartenbreddels:
It seems pandas 1.2 eagerly converts to dtype=object, e.g.:

```
delta = np.array([17658720110, 11047049384039, 40712636304958, -18161254954], dtype='timedelta64[s]')
pdf = pd.DataFrame({'delta': delta})
pdf.delta.dtype == 'O'
```

I couldn't find this in the pandas issues (cc @jorisvandenbossche)

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -16,7 +16,7 @@ jobs:
       max-parallel: 10
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
 
     steps:
     - uses: maxim-lobanov/setup-xcode@v1
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-    - PYTHON_VERSION: 3.6
-      MINICONDA: C:\Miniconda36-x64
     - PYTHON_VERSION: 3.7
       MINICONDA: C:\Miniconda37-x64
     - PYTHON_VERSION: 3.8
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -309,6 +309,11 @@ def _pandas_dt_fix(x):
         x = x.copy()
     return x
 
+def _to_pandas_series(x):
+    import pandas as pd
+    # pandas seems to eager to infer dtype=object for v1.2
+    return pd.Series(_pandas_dt_fix(x), dtype=x.dtype)
+
 @register_function(scope='dt', as_property=True)
 def dt_date(x):
     """Return the date part of the datetime value
@@ -336,7 +341,7 @@ def dt_date(x):
     2  2015-11-12
     """
     import pandas as pd
-    return pd.Series(_pandas_dt_fix(x)).dt.date.values.astype(np.datetime64)
+    return _to_pandas_series(x).dt.date.values.astype(np.datetime64)
 
 @register_function(scope='dt', as_property=True)
 def dt_dayofweek(x):
@@ -365,7 +370,7 @@ def dt_dayofweek(x):
     2  3
     """
     import pandas as pd
-    return pd.Series(_pandas_dt_fix(x)).dt.dayofweek.values
+    return _to_pandas_series(x).dt.dayofweek.values
 
 @register_function(scope='dt', as_property=True)
 def dt_dayofyear(x):
@@ -394,7 +399,7 @@ def dt_dayofyear(x):
     2  316
     """
     import pandas as pd
-    return pd.Series(_pandas_dt_fix(x)).dt.dayofyear.values
+    return _to_pandas_series(x).dt.dayofyear.values
 
 @register_function(scope='dt', as_property=True)
 def dt_is_leap_year(x):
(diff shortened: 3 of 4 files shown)


jorisvandenbossche:
The problem is that your timedeltas are out-of-bound for nanosecond resolution (the only resolution that pandas currently supports for datetime64/timedelta64, but I hope to change that in the near future).

So if you force the dtype (as it seems you are doing in this PR), you actually get wrong data:

```
In [1]: pd.__version__
Out[1]: '1.2.0'

In [2]: delta = np.array([17658720110, 11047049384039, 40712636304958, -18161254954], dtype='timedelta64[s]')

In [3]: pd.Series(delta)
Out[3]: 
0        204383 days, 8:01:50
1    127859367 days, 20:53:59
2     471211068 days, 8:15:58
3       -210200 days, 6:57:26
dtype: object

In [4]: pd.Series(delta, dtype=delta.dtype)
Out[4]: 
0    -9121 days +08:27:16
1   -29518 days +10:51:26
2      7779 days 07:58:01
3      3304 days 06:31:59
dtype: timedelta64[ns]

In [5]: pd.Series(delta, dtype="timedelta64[ns]")
Out[5]: 
0    -9121 days +08:27:16.290448384
1   -29518 days +10:51:26.978582016
2      7779 days 07:58:01.019583488
3      3304 days 06:31:59.709551616
dtype: timedelta64[ns]
```

That's actually a bug, though, as we shouldn't silently do such a wrong conversion. 
(although this silent conversion is actually coming from numpy's `astype` timedelta64[s] -> timedelta64[ns] not doing any boundschecking, but in pandas we should do this IMO).

But so on previous versions, we actually did this by default:

```
In [7]: pd.__version__
Out[7]: '1.1.5'

In [8]: pd.Series(delta)
Out[8]: 
0    -9121 days +08:27:16.290448384
1   -29518 days +10:51:26.978582016
2      7779 days 07:58:01.019583488
3      3304 days 06:31:59.709551616
dtype: timedelta64[ns]
```

So I would consider the change in pandas a bug fix (didn't check if it's listed in that section of the whatnew, though)

maartenbreddels:
Hmm, thanks for that Joris, I'll need to take a closer look.

This change was triggered by this:

```
delta = np.array([17658720110, 11047049384039, 40712636304958, -18161254954], dtype='timedelta64[s]')
df = pd.DataFrame({'delta': delta})
df.delta.dt.days
AttributeError                            Traceback (most recent call last)
<ipython-input-22-e1464cc7b97a> in <module>
      1 delta = np.array([17658720110, 11047049384039, 40712636304958, -18161254954], dtype='timedelta64[s]')
      2 df = pd.DataFrame({'delta': delta})
----> 3 df.delta.dt.days

~/miniconda/envs/dev/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5454             or name in self._accessors
   5455         ):
-> 5456             return object.__getattribute__(self, name)
   5457         else:
   5458             if self._info_axis._can_hold_identifiers_and_holds_name(name):

~/miniconda/envs/dev/lib/python3.7/site-packages/pandas/core/accessor.py in __get__(self, obj, cls)
    178             # we're accessing the attribute of the class, i.e., Dataset.geo
    179             return self._accessor
--> 180         accessor_obj = self._accessor(obj)
    181         # Replace the property with the accessor object. Inspired by:
    182         # https://www.pydanny.com/cached-property.html

~/miniconda/envs/dev/lib/python3.7/site-packages/pandas/core/indexes/accessors.py in __new__(cls, data)
    492             return PeriodProperties(data, orig)
    493 
--> 494         raise AttributeError("Can only use .dt accessor with datetimelike values")

AttributeError: Can only use .dt accessor with datetimelike values
```

Is this expected?

jorisvandenbossche:
It's expected in the sense that `.dt` is not supported for object dtype, only for timedelta/datetime dtype columns (in thus an expected consequence of the other change, i.e. that it is now object type, which in itself could have been unexpected of course ;))

But so, be aware that you previously simply got a wrong result from this `df.delta.dt.days`. Using pandas 1.1.5:

```
In [1]: delta = np.array([17658720110, 11047049384039, 40712636304958, -18161254954], dtype='timedelta64[s]')
   ...: df = pd.DataFrame({'delta': delta})
   ...: df.delta.dt.days
Out[1]: 
0    -9121
1   -29518
2     7779
3     3304
Name: delta, dtype: int64

In [2]: pd.Series(delta, dtype=object)
Out[2]: 
0        204383 days, 8:01:50
1    127859367 days, 20:53:59
2     471211068 days, 8:15:58
3       -210200 days, 6:57:26
dtype: object
```

So it's giving wrong "days" as result, because the `timedelta64[s]` array was incorrectly converted (due to integer overflow) to `timedelta64[ns]`. 
(so in that sense, it might be good that you are now aware of that wrong result ...)

maartenbreddels:
> (so in that sense, it might be good that you are now aware of that wrong result ...)

indeed!

maartenbreddels:
@JovanVeljanoski apart from the CI issues, I think we have to live with this inherited bug, for now, just want to make you aware of this.

JovanVeljanoski:
Can we specify pandas < 1.2 for the time being? Or is this not just 1.2 issue?

maartenbreddels:
No need to, this is not a new bug, it's that pandas catches an issue with overflow, and deals with that. We now avoid that new behavior by explicitly telling it not to do that, so pandas 1.2 solved a 'bug', we did not.
Formally I think this is a breaking change in Pandas, so it should have been a 2.0 thing, but that's life :)

jorisvandenbossche:
>  Formally I think this is a breaking change in Pandas, so it should have been a 2.0 thing, but that's life :)

I am typically more on the "safe side" within the pandas team when discussing possible breakages, but here I think avoiding complete data corruption can be considered a bug fix ;)
(but in the end, every bug fix is an API change when you rely on the buggy behaviour, but we need to be able to do some fixes before 2.0 ;))

PR #1159, opened 2021-01-14
chore: Remove duplicate for loop in GroupBy

kmcentush:
A for loop in GroupBy is unnecessarily duplicated


maartenbreddels:
Thanks!

PR #1162, opened 2021-01-15
Fix typo in _get_column_meta_data function

meredith-durbin:
Hello! I found a small typo that was causing an error when reading in FITS tables with `TUNITn` header keywords.


maartenbreddels:
Thanks for your first-time contribution @meredith-durbin !

PR #1170, opened 2021-01-20
feat: improved support for list (unique/value_count/data_type)

maartenbreddels:
image

and dtype support
image


maartenbreddels:
As discussed with @JovanVeljanoski and pushed for by @xdssio we now return a list by default:
image

By passing `as_python=False` you get the underlying type:
image

PR #1179, opened 2021-01-21
fix: allow executor to be in a different thread

maartenbreddels:
Because self.local._aggregator_nest_count is set from the calling thread
and the thread doing the execution may not have set the _aggregator_nest_count
attributes, we could get the following exception:
```
>       self.local._aggregator_nest_count += 1
E       AttributeError: '_thread._local' object has no attribute '_aggregator_nest_count'
```


maartenbreddels:
cc @mariobuikhuizen

PR #1181, opened 2021-01-22
feat: support Arrow lists in Expression.map

maartenbreddels:


packages/vaex-core/vaex/dataframe.py
```diff
@@ -472,23 +472,24 @@ def reduce(a, b):
...
     @docsubst
-    def unique(self, expression, return_inverse=False, dropna=False, dropnan=False, dropmissing=False, progress=False, selection=None, flatten=True, delay=False, array_type='python'):
+    def unique(self, expression, return_inverse=False, dropna=False, dropnan=False, dropmissing=False, progress=False, selection=None, axis=None, delay=False, array_type='python'):
         """Returns all unique values.
 
         :param dropmissing: do not count missing values
         :param dropnan: do not count nan values
         :param dropna: short for any of the above, (see :func:`Expression.isna`)
-        :param bool flatten: Flatten lists before finding the unique elements.
+        :param bool axis: Axis over which to determine the unique elements (None will flatten arrays or lists)
```
JovanVeljanoski, line 481: I wonder if this can be done more generally so we can use it in the docs with those shortcuts:
 like "Axis over which to do the operation (None will flatten arrays or lists)". Something like that, so we can re-use in `.sum` etc.. 

Not asking for this now, just an idea to think about

JovanVeljanoski (approved):

packages/vaex-core/vaex/dataframe.py
maartenbreddels: Indeed, I was thinking we need to do a pass over list features/docstrings, and add it to the tutorial.

PR #1185, opened 2021-01-25
feat: sum over list

maartenbreddels:
Example
```python
    data = [1, 2, None], None, [], [1, 3, 4, 5]
    df = vaex.from_arrays(i=pa.array(data).slice(offset))
    assert df.i.sum() == [16, 13][offset]
    assert df.i.sum(axis=1).tolist() == [3, None, 0, 13]
    assert df.i.sum(axis=1, fill_missing=99).tolist() == [3, 99, 0, 13]
    assert df.i.sum(axis=1, fill_missing_item=10).tolist() == [13, None, 0, 13]
    assert df.i.sum(axis=1, fill_empty=10).tolist() == [3, None, 10, 13]
    assert df.i.sum(axis=[0, 1]) == [16, 13][offset]
    # assert df.i.sum(axis=0)  # not supported, not sure what this should return
```

### DIFF
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -538,12 +538,51 @@ class Expression(with_metaclass(Meta)):
         kwargs['expression'] = self.expression
         return self.ds.count(**kwargs)
 
-    def sum(self, binby=[], limits=None, shape=default_shape, selection=False, delay=False, progress=None):
-        '''Shortcut for ds.sum(expression, ...), see `Dataset.sum`'''
-        kwargs = dict(locals())
-        del kwargs['self']
-        kwargs['expression'] = self.expression
-        return self.ds.sum(**kwargs)
+    def sum(self, axis=None, binby=[], limits=None, shape=default_shape, selection=False, delay=False, progress=None):
+        '''Sum elements over given axis.
+
+        If no axis is given, it will sum over all axes.
+
+        For non list elements, this is a shortcut for ds.sum(expression, ...), see `Dataset.sum`.
+
+        >>> list_data = [1, 2, None], None, [], [1, 3, 4, 5]
+        >>> df = vaex.from_arrays(some_list=pa.array(list_data))
+        >>> df.some_list.sum().item()  # will sum over all axis
+        16
+        >>> df.some_list.sum(axis=1).tolist()  # sums the list elements
+        [3, None, 0, 13]
+
+        :param int axis: Axis over which to determine the unique elements (None will flatten arrays or lists)
+        '''
+        expression = self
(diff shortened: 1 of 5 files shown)


maartenbreddels:
Note the 3 arguments, fill_missing/fill_missing_item/fill_empty. The first two could be done using fillmissing/fillna (which then then also support the axis argument), and possibly a new `Expression.fillempty()`, in any case, this will be much better performance.
Another thing we could think of: query optimization. e.g. `df.x.fillmissing(0).fillempty([0]).sum(axis=1)` could be optimized under water by this implementation, but we may not want to expose this as an API.

packages/vaex-core/vaex/expression.py
```diff
@@ -538,12 +538,40 @@ def count(self, binby=[], limits=None, shape=default_shape, selection=False, del
         kwargs['expression'] = self.expression
         return self.ds.count(**kwargs)
 
-    def sum(self, binby=[], limits=None, shape=default_shape, selection=False, delay=False, progress=None):
+    def sum(self, axis=None, binby=[], limits=None, shape=default_shape, selection=False, delay=False, fill_missing=None, fill_missing_item=0, fill_empty=0, progress=None):
```
maartenbreddels, line 541: @JovanVeljanoski I wonder if we should change the API and remove the binby/limits/shape arguments?

JovanVeljanoski: I was kind of hoping that the "flagship" methods like count, sum, mean, std would keep the arguments they have for the time being. 

If we remove them, maybe we can "port" them to the `binby` method, kind of like the plotting, the include what/limits/shape etc.. (whatever makes sense).

But for now I would keep it.. has been a core feature for a long time..

maartenbreddels: happy to keep them, was wondering if you had the same idea (df.sum will still have it btw)

packages/vaex-core/vaex/expression.py
```diff
@@ -538,12 +538,40 @@ def count(self, binby=[], limits=None, shape=default_shape, selection=False, del
...
         '''Shortcut for ds.sum(expression, ...), see `Dataset.sum`'''
```
JovanVeljanoski, line 542: This docstring should be updated. 

Probably the `df.sum()` should also be updated to include an explanation of the new kwargs.

JovanVeljanoski (commented):
Not fully convinced of the API, but probably this is a necessary step forward.

maartenbreddels:
After disscussing with Jovan, we indeed stripped down the API, all other cases should be dealt with using fillmissing/count/where etc

Issue #537, opened 2019-12-30
`where` is underperforming when dealing with string columns


JovanVeljanoski:
I am noticing a large performance slowdown when using `df.func.where` on virtual string column. 

This only happens for string columns, I have not been able to reproduce such performance slowdown for numeric data. 

Here is a reproducible example:

```python
import vaex
import numpy as np

# Create a couple of textual features
abc = 'abcdefghijklmnopqrstuvwxyz'
num = '0123456789'

# Text elements
x_elems = [''.join([i] * 5) for i in abc]
y_elems = [''.join([i] * 5) for i in num]

# Set probabilities - the 1st element should appear ~80% of the time
p_x = np.ones(26) * 0.008
p_x[0] = 0.8

# Create the data
size = 10_000_000
x = np.random.choice(x_elems, size=size, p=p_x)
y = np.random.choice(y_elems, size=size)

# Similar data but now numerical
z_elems = np.arange(9)
p_z = np.ones(9) * 0.025
p_z[0] = 0.8

z = np.random.choice(z_elems, size=size, p=p_z)
w = np.random.choice(np.arange(100), size=size)

# Create DataFrame
df = vaex.from_arrays(x=x, y=y, z=z, w=w)

# Add a virtual columns
df['str_feat'] = df.x + df.y
# Use where to adjust the feature 'feat' such that every time 'aaaaa' apears in df.x it should be the same in df.feat
df['str_feat_adj'] = df.func.where(df.x == 'aaaaa', 'aaaaa', df.str_feat)

# Do value counts of the new feature "str_feat" ---> takes <2 seconds
df.str_feat.value_counts(progress=True)

# Do value counts of the adjusted feature "str_feat_adj" ---> takes ~40 seconds.
df.str_feat_adj.value_counts(progress=True)

# Add a numerical virtual columns
df['num_feat'] = df.z + df.w
df['num_feat_adj'] = df.func.where(df.z == 0, 0, df.num_feat)


# Do value counts of the new feature "num_feat" ---> takes <1 second
df.num_feat.value_counts(progress=True)

# Do value counts of the adjusted feature "num_feat_adj" ---> takes <1 second
df.num_feat_adj.value_counts(progress=True)

```

JovanVeljanoski:
This has now been greatly improved!
What before took 40s now is only 2.5s ish.

If the data is stored as an hdf5 file, that decreases to 0.5s.

So i consider this fixed !
Thank you

PR #1186, opened 2021-01-25
String features for use with Apache Arrow v3.0

maartenbreddels:
We should support the string features that are ported to Apache Arrow v3, which should be released soon
 * [x] split https://github.com/apache/arrow/pull/8271
 * [x] work on arbitrary nested lists
 * [x] have custom join since https://github.com/apache/arrow/pull/8990 will not be in v3
 * ~~[ ] trim https://github.com/apache/arrow/pull/8621~~
 * [x] fillna for strings https://github.com/apache/arrow/pull/8628 which should make our fillna MUCH faster 🔥 
 * [x] Possibly implement `where` using fillna/take since that is also quite slow on strings now

### DIFF
--- a/ci/conda-env-nightlies.yml
+++ b/ci/conda-env-nightlies.yml
@@ -1,1 +1,1 @@
-# pyarrow
+pyarrow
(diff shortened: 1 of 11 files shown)


maartenbreddels:
Seems like trim has two bugs, it doesn't take the TrimOptions for ascii kernels, and the tests fail. I'll open a PR to expose them later, and we can file an issue at the Arrow JIRA.

maartenbreddels:
Closes https://github.com/vaexio/vaex/issues/537

PR #1193, opened 2021-01-27
fix: NumpyDispatch with ColumnStringArrow

kmcentush:
In the case of ColumnStringArrow, NumpyDispatch was converting to a PyArrow array but then not doing anything with it. Instead, it needed to updated the object's `_array` attribute.


kmcentush:
Oh wow. This actually forces the UnicodeDecodeError I was mentioning over Slack for a ton of unit cases...

maartenbreddels:
Thanks, this revealed an underlying issue that had me puzzled for a while, which #1194 should solve, if that is merged, could you rebase this?

maartenbreddels:
Thanks @kmcentush !

PR #1206, opened 2021-02-02
Feat: allow parsing of the FloatDtypes() from pandas

JovanVeljanoski:
This PR allows vaex to parse the `Float64Dtype()` from pandas, which allows for masked values (also works for 32bits).

- [x] Implementation
- [x] Test update
- [ ] Review


tests/from_pandas_test.py
```diff
@@ -22,6 +22,7 @@ def test_from_pandas():
 
     # Get pandas dataframe
     pandas_df = pd.DataFrame(dd_dict)
+    pandas_df['float_missing_masked'] = pandas_df['float_missing'].astype(pd.Float64Dtype())
```
maartenbreddels, line 25: You are relying on astype to go from nan to None, I propose we add a new series with an explicit dtype, similar to int_missing.

JovanVeljanoski: Done!

PR #1221, opened 2021-02-18
Allow use of widget progress bar in PCA

JovanVeljanoski:
Just makes things look prettier.


maartenbreddels:
You've checked the spec, so you need to run:
```
$ python -m vaex.ml.spec packages/vaex-ml/vaex/ml/spec.json
```
This change is compatible with older states files, so this is ok.

PR #1244, opened 2021-03-09
State transfer fix 

JovanVeljanoski:
This exposes unwanted behavior when filtered dataframe is also sliced. See comments in the code for mode details.


JovanVeljanoski:
This is amazing - thank you!!

PR #1256, opened 2021-03-14
River integration

JovanVeljanoski:
This PR provides a convenient wrapper around models and even pipelines of [River](https://github.com/online-ml/river).
The requirement is that the models implement the `learn_many` method, and currently it is assumed that the `predict_many` method returns a pandas Series. 

For the reason / motivation behind this, please see https://github.com/online-ml/river/discussions/515.

With the wrapper, the River models become part of the vaex computational graph, and one gets full benefit of the serialization and lazy evaluation. 

Here is a basic example:

```python
import vaex
import vaex.ml
from vaex.ml.incubator.river import RiverModel

from river.linear_model import LinearRegression
from river import optim

df = vaex.example()

# Train test split
df_train, df_test = df.ml.train_test_split(test_size=0.2, verbose=False)

# Features
features = ['x', 'y', 'z', 'vx', 'vy', 'vz']
target = 'FeH'

# Preprocessing
df_train = df_train.ml.standard_scaler(features=features, prefix='scaled_')

# New pre-processed features
features = df_train.get_column_names(regex='^scaled_')

river_model = RiverModel(model=LinearRegression(optimizer=optim.SGD(lr=0.1), intercept_lr=0.1), 
                         features=features, 
                         target=target, 
                         batch_size=10_000, 
                         num_epochs=1,
                         prediction_name='river')
river_model.fit(df=df_train, progress='widget')
df_train = river_model.transform(df_train)

# Serialize to disk
df_train.state_write('./state_river.json')
df_test.state_load('./state_river.json')
df_test.head(5)
```

One can even use the full River pipeline, provided each step in the pipeline implements the `learn_many` method:

```python
import vaex
import vaex.ml
from vaex.ml.incubator.river import RiverModel

from river.linear_model import LinearRegression
from river import optim
from river.compose import Pipeline
from river.preprocessing import StandardScaler

df = vaex.example()

# Train test split
df_train, df_test = df.ml.train_test_split(test_size=0.2, verbose=False)

# Features
features = ['x', 'y', 'z', 'vx', 'vy', 'vz']
target = 'FeH'


pipe = Pipeline(('scale', StandardScaler()),
                ('model', LinearRegression(optimizer=optim.SGD(lr=0.1), intercept_lr=0.1))
               )


river_pipe = RiverModel(model=pipe,
                        features=features, 
                        target=target, 
                        batch_size=10, 
                        num_epochs=1,
                        prediction_name='river')

river_pipe.fit(df=df_train, progress='widget')
df_train = river_pipe.transform(df_train)

# Serialize to disk
df_train.state_write('./state_river.json')
df_test.state_load('./state_river.json')
df_test.head(5)                        
```

Currently only the `LinearRegression` and `LogisticRegression` models are tested. We hope that River will implement the `learn_many` method to more of their models - they have a decent variety of models, including tree-based models and are doing excellent work. 

The Dense neural network implementation in principle should also work, but the `predict_many` method returns a pandas DataFrame instead of a series. 

Thoughts & Comments welcome.

### DIFF
--- a/packages/vaex-ml/vaex/ml/incubator/river.py
+++ b/packages/vaex-ml/vaex/ml/incubator/river.py
@@ -70,11 +70,19 @@ class RiverModel(state.HasState):
     num_epochs = traitlets.Int(default_value=1, allow_none=False, help='Number of times each batch is sent to the model.')
     shuffle = traitlets.Bool(default_value=False, allow_none=False, help='If True, shuffle the samples before sending them to the model.')
     prediction_name = traitlets.Unicode(default_value='prediction', help='The name of the virtual column housing the predictions.')
+    prediction_type = traitlets.Enum(values=['predict', 'predict_proba'], default_value='predict',
+                                     help='Which method to use to get the predictions. \
+                                     Can be "predict" or "predict_proba" which correspond to \
+                                     "predict_many" and "predict_proba_many in a River model respectively.')
 
     def __call__(self, *args):
         X = {feature: np.asarray(arg, np.float64) for arg, feature in zip(args, self.features)}
         X = pd.DataFrame(X)
-        return self.model.predict_many(X).values
+
+        if self.prediction_type == 'predict':
+            return self.model.predict_many(X).values
+        else:
+            return self.model.predict_proba_many(X).values
 
 
     def predict(self, df):
@@ -84,7 +92,7 @@ class RiverModel(state.HasState):
         :returns: A in-memory numpy array containing the Model predictions
         :rtype: numpy.array
         '''
-        return self.model.predict_many(df.to_pandas_df(column_names=self.features)).values
+        return self.transform(df)[self.prediction_name].values
 
 
     def transform(self, df):
@@ -111,7 +119,7 @@ class RiverModel(state.HasState):
         '''
 
         # Check whether the model is appropriate
-        assert hasattr(self.model, 'learn_many')
+        assert hasattr(self.model, 'learn_many'), 'The model must implement the `.learn_many` method.'
 
         n_samples = len(df)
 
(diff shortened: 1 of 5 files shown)


packages/vaex-ml/vaex/ml/incubator/river.py
```diff
@@ -0,0 +1,137 @@
...
+        # Portions of the DataFrame to evaluate
+        expressions = self.features + [self.target]
+
+        for epoch in range(self.num_epochs):
+            for i1, i2, df_tmp in df.to_pandas_df(column_names=expressions, chunk_size=self.batch_size):
+                progressbar((n_samples * epoch + i1) / (self.num_epochs * n_samples))
+                y_tmp = df_tmp.pop(self.target)
+
+                if self.shuffle:
+                    shuffle_index = np.arange(len(df_tmp))
```
maartenbreddels, line 129: this does not shuffle, do you want np.random?

MaxHalford:
Note that River's `LogisticRegression` returns a `DataFrame` when `predict_proba_many` is called.

packages/vaex-ml/vaex/ml/incubator/river.py
JovanVeljanoski: there was a line missing - fixed -thanks!

JovanVeljanoski:
> Note that River's `LogisticRegression` returns a `DataFrame` when `predict_proba_many` is called.

Thanks, I was not aware! 
I've added (rudimentary) support for this. The pandas series is just converted to numpy arrays, so you lose naming information. 

@maartenbreddels it would be nice if a model returns a pandas dataframe, the output to be json records (meaning `pd.Dataframe.to_dict(orient='records')`, and save it as a arrow `Struct`.  If we can do this, we can preserve the names of the columns. This would also be relevant for the tensorflow integration (if possible of course).

I could not figure out how to do this with `pyarrow` tho, could you give me some advice ?

JovanVeljanoski:
@maartenbreddels Please check out the failures, it is something to do with strings and arrow..

maartenbreddels:
@JovanVeljanoski i think we should import this in generate.py so we get the fast API as well, what do you think?

JovanVeljanoski:
> @JovanVeljanoski i think we should import this in generate.py so we get the fast API as well, what do you think?

Yup, just added this.

maartenbreddels:
Cool, really like to see this released now :)

PR #1264, opened 2021-03-16
Fix: Support for loading state files from GCS

JovanVeljanoski:
This PR irons out the support for loading state files from cloud storage. 

To do:
- [x] Implement a unit-test for this
- [ ] Add support to ignore or not to cache files (as model/state might be updated in the cloud, and we want to pick up this change)
- [ ] Review


maartenbreddels:
Thx!

PR #1305, opened 2021-04-05
RandomProjections implementation

JovanVeljanoski:
This PR implements the `RandomProjections` method for dimensional reduction. 

It uses the methods in scikit-learn for creating the random matrices, so scikit-learn is needed as a dependency of using this transformer.  Scikit-learn is however not needed to apply the transformations of an already fitted transformer. 

- [x] `RandomProjections` implementation
- [x] Unit tests
- [ ] Code review

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -8,6 +8,9 @@ from vaex.utils import _ensure_strings_from_expressions
 import numpy as np
 import warnings
 
+sklearn = vaex.utils.optional_import("sklearn.decomposition")
+sklearn = vaex.utils.optional_import("sklearn.random_projection")
+
 help_features = 'List of features to transform.'
 help_prefix = 'Prefix for the names of the transformed features.'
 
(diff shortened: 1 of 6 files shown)


xdssio:
Looks great

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -121,6 +121,117 @@ def transform(self, df, n_components=None):
...
+        else:
+            raise traitlets.TraitError('`density` must be 0 < density <= 1.')
+
+    def fit(self, df):
+        '''Fit the RandomProjections to the DataFrame.
+
+        :param df: A vaex DataFrame.
+        '''
+
+        sklearn = vaex.utils.optional_import("sklearn.random_projection")
```
maartenbreddels, line 191: The idea of this function is that we can put this on the top level instead.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -121,6 +121,117 @@ def transform(self, df, n_components=None):
...
+    prefix = traitlets.Unicode(default_value="random_projection_", help=help_prefix)
+    random_state = traitlets.Int(default_value=None, allow_none=True, help='Controls the pseudo random number generator used to generate the projection matrix at fit time. Used to get reproducible results.')
+    random_matrix_ = traitlets.List(traitlets.List(traitlets.CFloat()), help='The random matrix.').tag(output=True)
+
+    @traitlets.validate('eps')
+    def _valid_eps(self, proposal):
+        if (proposal['value'] > 0) & (proposal['value'] < 1):
+            return proposal['value']
+        else:
+            raise traitlets.TraitError('`eps` must be between 0 and 1.')
```
maartenbreddels, line 176: we can also rely on min/max of the trait, like traitlets.Int(..., min=0, max=1), however, those bounds are inclusive.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -196,7 +307,7 @@ class OneHotEncoder(Transformer):
     >>> import vaex
     >>> df = vaex.from_arrays(color=['red', 'green', 'green', 'blue', 'red'])
     >>> df
-     #  color
+     #  color®
```
maartenbreddels, line 310:
```suggestion
     #  color
```
Typo?

tests/ml/ml_test.py
```diff
@@ -37,6 +44,39 @@ def test_valid_sklearn_pca(df_iris):
...
+@pytest.mark.parametrize("n_components", [5, 15, 150])
+@pytest.mark.parametrize("matrix_type", ['gaussian', 'sparse'])
+def test_random_projections(n_components, matrix_type):
+    df = vaex.from_dict(data=data_maker(n_rows=100_000, n_cols=31))
+    features = df.get_column_names()
+
+    rand_proj = vaex.ml.RandomProjections(features=features, n_components=n_components, matrix_type=matrix_type)
+    df_trans = rand_proj.fit_transform(df)
+    assert np.array(rand_proj.random_matrix_).shape == (n_components, 31)
+    assert len(df_trans.get_column_names(regex='^rand')) == n_components
```
maartenbreddels, line 56: should we test the fast api? or since that is automated, we assume it works?

maartenbreddels (changes requested):
Great work, and thanks a lot for solving the fsspec issue!!
Just some minor comments

tests/ml/ml_test.py
JovanVeljanoski: Ok I can re-write it to use the fast api

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Yeah i tried that and found that out. Do you mind leaving it like this, or do you propose a chance of some kind?

maartenbreddels: no, if this was the reason, we keep it like this!

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -8,6 +8,9 @@
 import numpy as np
 import warnings
 
+sklearn = vaex.utils.optional_import("sklearn.decomposition")
+sklearn = vaex.utils.optional_import("sklearn.random_projection")
```
maartenbreddels, line 12: The 1st statement doesn't do anything, since the 2nd overwrite it, but I understand what you mean. But you do not refer to decomposition anyway, so we can just remove it right?
```suggestion
sklearn = vaex.utils.optional_import("sklearn.random_projection")
```

maartenbreddels:
Thanks!

PR #1310, opened 2021-04-07
Fix: evaluating very long expressions

JovanVeljanoski:
We get memory errors when trying to evaluate **very** long expressions. This happens for example when doing pca transformations on 100s of columns (which is a common use of a PCA transformation). 

- [x] Write tests to expose the issue
- [ ] Fix


kmcentush:
I understand that this isn't the highest priority, but is there any guess as to a timeline when this may be looked into more?

maartenbreddels:
I wonder if we should do the sum/product with a function instead, like dot product, that might also be more efficient.

JovanVeljanoski:
Will close this since https://github.com/vaexio/vaex/pull/1671 was just merged. 

If this problem re-appears in a different context, let's revisit this.

PR #1311, opened 2021-04-08
doc: extend the df.categorize example

rgommers:
Without the added `df =` the second line of the example did nothing.

Showing that the representation doesn't change but `is_category` will now reveal
that the dataframe knows a column has a categorical dtype also seems useful.


maartenbreddels:
Thanks Ralf!

PR #1323, opened 2021-04-21
Adds dtype arg to dt.func.where()

### LINKED ISSUE #1309: [FEATURE-REQUEST] OHE encoder to output columns with boolean dtype
OHE currently outputs int64 cols which is highly uneconomical. It is possible to cast it to int32 or bool of course but this adds another virtual column in the df state and it makes the state transfer slower.

This is probably nothing when dealing with small amount of data but when things get bigger, it helps a lot to save some space by downcasting integer and floats.

kmcentush:
Fixes #1309

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -2561,7 +2561,7 @@ def add_geo_json(ds, json_or_file, column_name, longitude_expression, latitude_e
 import vaex.arrow.numpy_dispatch
 
 @register_function()
-def where(condition, x, y):
+def where(condition, x, y, dtype=None):
     # special where support for strings
     # TODO: this should be replaced by an arrow compute function in the future
     if type(x) == str:
@@ -2613,4 +2613,9 @@ def where(condition, x, y):
         return values
     # default callback is on numpy
     ar = np.where(condition, x, y)
+
+    # Cast to dtype
+    if dtype is not None:
+        ar = ar.astype(dtype)
+
     return ar
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -2611,11 +2611,18 @@ def where(condition, x, y, dtype=None):
         choices = vaex.array_types.concat([x, y])
         values = choices.take(indices)
         return values
-    # default callback is on numpy
-    ar = np.where(condition, x, y)
 
-    # Cast to dtype
+    # cast x and y
     if dtype is not None:
-        ar = ar.astype(dtype)
+        if np.can_cast(x, dtype):
+            dtype_scalar = np.dtype(dtype)
+            x = dtype_scalar.type(x)
+        if np.can_cast(y, dtype):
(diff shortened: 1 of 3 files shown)


packages/vaex-core/vaex/functions.py
```diff
@@ -2613,4 +2613,9 @@ def where(condition, x, y):
         return values
     # default callback is on numpy
     ar = np.where(condition, x, y)
+
+    # Cast to dtype
+    if dtype is not None:
+        ar = ar.astype(dtype)
+
```
maartenbreddels, line 2620: We can be more efficient and do sth like this before the where:
```if dtype is not None:
    if isinstance(x, numbers.Number):
        dtype_scalar = np.dtype(dtype)
        x = dtype_scalar.type(x)
    if isinstance(y, numbers.Number):
        dtype_scalar = np.dtype(dtype)
        x = dtype_scalar.type(x)
```
We'd need to the final `astype` to catch cases for arrays though.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -423,6 +422,9 @@ def fit(self, df):
             uniques.append(unique_values)
         self.uniques_ = uniques
 
+        # detect ability to downcast to uint8
+        self.downcast_uint8 = np.can_cast(self.one, np.uint8) and np.can_cast(self.zero, np.uint8)
```
maartenbreddels, line 426: nice! I think we don't have to make it a member though
```suggestion
```

packages/vaex-core/vaex/functions.py
maartenbreddels: btw, good that you check for dtype is not None, because otherwise I think it will cast to float64!

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -431,17 +433,18 @@ def transform(self, df):
         :rtype: DataFrame
         '''
         copy = df.copy()
+        dtype = self.downcast_uint8 and 'uint8' or None
```
maartenbreddels, line 436: We can put the check here right? And I'm never a fan of using `and` this way, I prefer `x if cond else y`
```suggestion
        downcast_uint8 = np.can_cast(self.one, np.uint8) and np.can_cast(self.zero, np.uint8)
        dtype = 'uint8' if downcast_uint8 else None
```

maartenbreddels (changes requested):
Nice work Kyle, I have added some suggestions, let me know if you agree with them.

packages/vaex-core/vaex/functions.py
kmcentush: Good comments! I'll make the changes tonight. I should have taken this approach prior. Guess that's what 1wk out of town does to you. :)

kmcentush: Ok, I made some changes. Let me know what you think!

maartenbreddels:
Thanks Kyle!

PR #1326, opened 2021-04-24
🐛Fix isin() dtype downcasting and use_hashmap=False

### LINKED ISSUE #1320: [BUG-REPORT]
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Please provide a clear and concise description of the problem. This should contain all the steps needed to reproduce the problem. A minimal code example that exposes the problem is very appreciated.

Following code returns an incorrect result

```python
df=vaex.from_arrays(x=np.array([1.01,1.02,1.03]))

df.x.isin([1])
```

Expression = isin_set(x, var_isin_ordered_set)
Length: 3 dtype: bool (expression)
0 True
1 True
2 True

should be all false

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.1
 - Vaex was installed via: pip / conda-forge / from source  conda-forge
 - OS: fedora 33

**Additional information**
Please state any supplementary information or provide additional context for the problem (e.g. screenshots, data, etc..).

kmcentush:
- Fixes #1320 at the Python level; technically, the bug is caused at the C++ level when the `set.isin()` function implicitly downcasts its own values to the same type as the value its comparing against; ex: input column is a float and `isin([1])` is called (1 is an int) --> all floats are downcasted/rounded down to the nearest int and can return incorrect results
- One test case/assert isolates a potential C++ bug but I was not able to fix it at this time; more details are below

Currently, one assert case fails for a masked string column when `use_hashmap = False`. Specifically, I believe there is a bug at the C++ level when a masked string column's `<string_column>.string_sequence.isin()` function is called. There are no issues when `use_hashmap = True` as evident by an earlier assert in the test case. Hoping a vaex maintainer can take a closer look at it or guide me in the right direction.

### DIFF
--- a/packages/vaex-core/src/strings.cpp
+++ b/packages/vaex-core/src/strings.cpp
@@ -377,10 +377,10 @@ class StringSequenceBase : public StringSequence {
             if(has_null() || others->has_null()) {
                 for(size_t i = 0; i < length; i++) {
                     bool found = false;
-                    if(is_null(i)) {
+                    if(!is_null(i)) {
                         auto str = view(i);
                         for(size_t j = 0; j < others->length; j++) {
-                            if(others->is_null(j)) {
+                            if(!others->is_null(j)) {
                                 auto other = others->view(j);
                                 found = str == other;
                                 if(found)
--- a/packages/vaex-core/vaex/expression.py
+++ b/packages/vaex-core/vaex/expression.py
@@ -864,6 +864,7 @@ def f({0}):
         """
         if use_hashmap:
             # easiest way to create a set is using the vaex dataframe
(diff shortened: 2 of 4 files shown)


kmcentush:
Whoops, looks like I need to rebase onto master to clean up the git history. Let me fix that.

maartenbreddels:
Thanks, Kyle for finding the cause of this. I've parametrized the tests, and fixed the other bug you found.

PR #1344, opened 2021-05-05
🔨 categorical data should hide that it is dict/ordinal encode

maartenbreddels:
It should just work as if a dataframe after df.ordinal_encode is the
same (so unique gives the same results etc). The only thing that should
be different is the performance.


maartenbreddels:
TODO:
 [ ] - value_counts

maartenbreddels:
@JovanVeljanoski I think we can merge this, since you need to opt in using `df._auto_encode()` to get the new behaviour (which should be the default in v5). Maybe we should open an issue for the 2 TODO's?

JovanVeljanoski:
Do it!

PR #1346, opened 2021-05-05
✨ arrow/pandas cannot do all data conversion yet

maartenbreddels:
e.g. dict encoded uint8/uint16
TODO: test uint16/32/64


maartenbreddels:
This should be done similar to #1344

PR #1355, opened 2021-05-18
Corrected typo - missing % in formatting string

Barry1:
There was a typo - a print call with a string an immediately after that the arguments to be inserted without the percent sign


maartenbreddels:
Thanks Bastian!

Barry1:
> Thanks Bastian!

You're welcome - it's always a pleasure assisting good things - even with such small bricks. Best regards.

PR #1364, opened 2021-05-21
To records

xdssio:
This is a _to_records_ implementation for Vaex.

It follows the Vaex convention and parameters like _to_arrays_, and _to_dict_, but results with the equivalent results for the pandas _df.to_dict(orient='records')_.

This help with the integrations of Vaex with applications which relies on records.

example:
```
>>> df = vaex.from_dict({'A':[1,2,3],'B':['a','b','c']})
>>> for  _, _ , chunk in df.to_records(chunk_size=2): 
...    print(chunk)
[{'A': 1, 'B': 'a'}, {'A': 2, 'B': 'b'}]
[{'A': 3, 'B': 'c'}]


>>> df.to_records(0)
{'A': 1, 'B': 'a'}

```


tests/evaluate_test.py
```diff
@@ -48,6 +48,19 @@ def test_to_dict(df_local, chunk_size, parallel, array_type):
...
+@pytest.mark.parametrize("parallel", [True, False])
+@pytest.mark.parametrize("array_type", ['numpy', 'list', 'python'])
+def test_to_records(df_local, chunk_size, parallel, array_type):
+    df = df_local
+    for i1, i2, chunk in df.to_records(chunk_size=chunk_size, parallel=parallel, array_type=array_type):
+        assert isinstance(chunk, list)
+        assert len(chunk) <= chunk_size
+        assert isinstance(chunk[0], dict)
+    record = df.to_records(0)
+    assert isinstance(record, dict)
```
maartenbreddels, line 61: It is missing a test without a chunk_size, which should not return an iterator.

maartenbreddels (commented):
Yeah, I really think we should have this, not sure about the method name since https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_records.html is something completely different.

What we could do, is stay close to https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_dict.html and add an orient='dict' option (although not intuitive for new people, easier to pandas people).

JovanVeljanoski:
I like this PR. I think this is super useful to have. 
As for the API.. no strong preference. 

Perhaps simpler to adopt pandas's convention and add the `orient` argument in `to_dict`, but I am also happy for this to stay as is. If it stays, I can see us answering bunch of issues here about why it is different than the pandas behaviour (while we do for hdf5 and few other methods).

xdssio:
Although in general, I would agree with copy pandas API,  vaex has many differences with all of these methods to pandas; the iterators in chunks, the defaults return structures, and of course, the index which makes practically all outputs a bit different than the vaex version. 
If pandas made a it so that having orient='records' and to_records been to different things, doesn't mean we have to.

One of the most useful parts of it is actually the ability to get a single line, which will be lost if we use the pandas API.

maartenbreddels:
@JovanVeljanoski does it make sense to include this in the io docs? (you probably know the best place for this)

packages/vaex-core/vaex/dataframe.py
```diff
@@ -2686,6 +2686,44 @@ def f(i1, i2):
                 previous_chunk = previous.result()
                 yield previous_l1, previous_l2, previous_chunk
 
+    @docsubst
+    def to_records(self, index=None, selection=None, column_names=None, strings=True, virtual=True, parallel=True,
+                   chunk_size=None, array_type='python'):
+        """Return a list of [{column_name: value}, ...)] "records" where each dict is an evaluated row.
```
maartenbreddels, line 2692: to make the substitution not work here:
```suggestion
        """Return a list of [{{column_name: value}}, ...)] "records" where each dict is an evaluated row.
```

packages/vaex-core/vaex/dataframe.py
```diff
@@ -2686,6 +2686,44 @@ def f(i1, i2):
...
+
+                :param index: an index to use to get the record of a specific row when provided
+                :param column_names: list of column names, to export, when None DataFrame.get_column_names(strings=strings, virtual=virtual) is used
+                :param selection: {selection}
+                :param strings: argument passed to DataFrame.get_column_names when column_names is None
+                :param virtual: argument passed to DataFrame.get_column_names when column_names is None
+                :param parallel: {evaluate_parallel}
+                :param chunk_size: {chunk_size}
+                :param array_type: {array_type}
+                :return: list of [{column_name:value},...] records
```
maartenbreddels, line 2702: same here.
```suggestion
                :return: list of [{{column_name:value}}, ...] records
```

JovanVeljanoski:
Yeah, once this is merged I will update the I/O docs, unless you wait it as part of this PR?

maartenbreddels:
@JovanVeljanoski separate is good, wanna open an issue to keep track?

Thanks @xdssio !

PR #1365, opened 2021-05-22
Make development setup easier with one-line make command

Alon-Alexander:
Hi, this is my first PR to `vaex`.
I would really like to start contributing to the project, and starting out I had a rough time installing `vaex` for development.
I decided to try and make it easier to install for development as my first issue.

Eventually I also made `vaex-core`'s `setup.py` a bit more standard (using `Pybind11Extension`).

_Note_: I didn't fully understand the `vendor` folder. Not its purpose nor how it is supposed to be used.
Thus I might have changed things there not in the way it was designed to be, so I would like an explanation on how to use it in the designed way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,3 @@
-
-
-
 nose:
 	nosetests --with-coverage --cover-package=vaex --cover-html test/dataset.py
 # test/ui.py
@@ -13,13 +10,6 @@ coverage:
 	coverage html -d cover
 	open cover/index.html
 
-deps:
-	git -C packages/vaex-core/vendor/string-view-lite pull || \
-   		git clone https://github.com/martinmoene/string-view-lite.git packages/vaex-core/vendor/string-view-lite
-
-	git -C packages/vaex-core/vendor/hopscotch-map pull || \
-   		git clone https://github.com/Tessil/hopscotch-map.git packages/vaex-core/vendor/hopscotch-map
-
 test:
 	VAEX_DEV=1 python -m pytest tests/
 
@@ -27,8 +17,10 @@ clean-test:
 	rm -rf smæll2.parquet smæll2.yaml tests/data/parquet/ tests/data/smæll2.csv.hdf5 tests/data/unittest.parquet
 	find . -wholename "./tests/data/parquet_dataset_partitioned_*" -delete
 
+generate-docs:
+	make -C ./docs/ html
+
 init:
-	make deps
 	python -m pip install -e .[dev]
 
 uninstall:
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,32 @@ coverage:
 	coverage html -d cover
 	open cover/index.html
 
+deps:
+	git -C packages/vaex-core/vendor/string-view-lite pull || \
+   		git clone https://github.com/martinmoene/string-view-lite.git packages/vaex-core/vendor/string-view-lite
+
+	git -C packages/vaex-core/vendor/hopscotch-map pull || \
+   		git clone https://github.com/Tessil/hopscotch-map.git packages/vaex-core/vendor/hopscotch-map
+
 test:
-	python test/dataset.py
+	VAEX_DEV=1 python -m pytest tests/
+
+clean-test:
+	rm -rf smæll2.parquet smæll2.yaml tests/data/parquet/ tests/data/smæll2.csv.hdf5 tests/data/unittest.parquet
+	find . -wholename "./tests/data/parquet_dataset_partitioned_*" -delete
+
+init:
+	make deps
+	python -m pip install -e .[dev]
+
+uninstall:
+	 @pip uninstall -y `pip list | grep -o -E "^vaex([-a-zA-Z0-9]+)?"` 2> /dev/null ||\
+	 	echo "Can't uninstall: Vaex is not installed"
+
+clean:
+	make uninstall clean-test
+	rm -rf packages/vaex-core/build/
+	find . -wholename "*egg-info*" -delete
 
 release:
 	echo ""
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = ../venv/bin/sphinx-build
+SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
 
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = ../venv/bin/sphinx-build
 PAPER         =
 BUILDDIR      = build
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -174,7 +174,7 @@ html_theme_options = dict(
 extra_navbar = "<h1>dsadsa</h2>"
 
 html_sidebars = {
-	"**": ["sbt-sidebar-footer.html", "sbt-sidebar-nav.html", "searchbox.html"]
+	"**": ["sbt-sidebar-footer.html", "sbt-sidebar-nav.html", "sidebar-search-bs.html"]
 }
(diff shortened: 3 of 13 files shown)


Makefile
```diff
@@ -13,8 +13,32 @@ coverage:
...
+	rm -rf smæll2.parquet smæll2.yaml tests/data/parquet/ tests/data/smæll2.csv.hdf5 tests/data/unittest.parquet
+	find . -wholename "./tests/data/parquet_dataset_partitioned_*" -delete
+
+init:
+	make deps
+	python -m pip install -e .[dev]
+
+uninstall:
+	 @pip uninstall -y `pip list | grep -o -E "^vaex([-a-zA-Z0-9]+)?"` 2> /dev/null ||\
+	 	echo "Can't uninstall: Vaex is not installed"
```
maartenbreddels, line 36: love it!

maartenbreddels (changes requested):
Great start, love that you first start attacking issues you encounter.

The vendor directory is where the git submodules live, and as documented at `https://vaex.io/docs/installing.html#for-developers` you'll get them if you run:
`git clone --recursive https://github.com/vaexio/vaex`.

As everyone (including me) simply does a `git clone <URL>` you'll miss those, and get a very vague error msg when running setup.py. Maybe a good fix for that would be to check the directories (e.g. `packages/vaex-core/vendor/hopscotch-map`) in setup.py and if they are empty, suggest the user to run `git submodule update --init` which should fix that.

What do you think?

cheers,

Maarten

packages/vaex-core/src/vaexfast.cpp
```diff
@@ -2122,7 +2122,7 @@ static struct PyModuleDef moduledef = {
 #define INITERROR return
 #endif
 
-
+__attribute__((visibility("default")))
```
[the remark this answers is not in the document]
Alon-Alexander: When changing to `pybind11`'s extension, it automatically adds some default (recommended) configuration for the compilation. One of the options is (on non-windows) `-fvisibility=hidden`, which makes all symbols (like functions) local by default, which means they are not exported.
That helps to create a smaller binary size at the end and is recommended by pybind11, but with that option you need to explicitly make a function (like the module init function) global.

docs/source/installing.rst
```diff
@@ -49,7 +49,8 @@ If you want to work on vaex for a Pull Request from the source, use the followin
 * make sure the dev versions of pcre are installed (e.g. ``conda install -c conda-forge pcre``)
 * install using (note: if you're on Windows, make sure that your command line/terminal has administrator privileges):
 
- * ``pip install -e .``  (again, use (ana)conda or virtualenv/venv)
+ * Create and activate a virtualenv/venv
+ * ``make init``
```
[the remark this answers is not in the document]
Alon-Alexander: If make doesn't work on windows, they won't be able to use the makefile at all :)
Do you have another good alternative? Because I don't want to also inline the `make deps`, that would be messy.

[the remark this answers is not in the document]
Alon-Alexander: Someone without `make` won't be able to use that file at all, and that's fine because it's a utility file.
But if they use the makefile they probably have `make`.
Do you have another solution for that? Because I don't want to change `make clean` for example, as it would be very messy.

packages/vaex-core/src/vaexfast.cpp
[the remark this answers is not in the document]
maartenbreddels:
```suggestion
```
compiler doesn't seem to like it.

packages/vaex-core/setup.py
```diff
@@ -75,14 +60,12 @@ def __str__(self):
     extra_compile_args.append("-mmacosx-version-min=10.9")
 
 # on windows (Conda-forge builds), the dirname is an absolute path
-extension_vaexfast = Extension("vaex.vaexfast", [os.path.relpath(os.path.join(dirname, "src/vaexfast.cpp"))],
+extension_vaexfast = Pybind11Extension("vaex.vaexfast", [os.path.relpath(os.path.join(dirname, "src/vaexfast.cpp"))],
```
maartenbreddels, line 63: Not using Pybind11 here:
```suggestion
extension_vaexfast = Extension("vaex.vaexfast", [os.path.relpath(os.path.join(dirname, "src/vaexfast.cpp"))],
```

packages/vaex-core/setup.py
```diff
@@ -2,7 +2,7 @@
 import sys
 import os
 import imp
-from setuptools import Extension
+from pybind11.setup_helpers import Pybind11Extension
```
maartenbreddels, line 5: still need the import 
```suggestion
from setuptools import Extension
from pybind11.setup_helpers import Pybind11Extension
```

maartenbreddels:
I wonder if pybind11 adds some compilation flags, because now the vaexfast extension doesn't export a symbol anymore 🤷

Alon-Alexander:
I see your point @maartenbreddels 
I do know that pybind adds a few flags (this is why I added the attribute), but as it looks it changes something globally (as it doesn't expose the symbol in the changes you made).
Why don't we compile everything with pybind? Do we care?
Another solution would be to add the attribute only on non-windows (as I should have originally added it 😅 ).

maartenbreddels:
> Why don't we compile everything with pybind?

Yeah, I'm good with that, but CI failed with it, maybe we should also use `PYBIND11_MODULE` in vaexfast.cpp, do you want to try that?

maartenbreddels:
I don't know why the linux build has this visibility issue, lost there 🤷

Alon-Alexander:
I noticed that most changes we talked about in this PR are unrelated to the change I wanted to make and belong in another refactoring PR.
Since I really want to merge the changes that will help other people develop vaex, I reverted most changes - the ones that refactored old code instead of addressing the issue I wanted to address.

docs/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = sphinx-build
+SPHINXBUILD   = ../venv/bin/sphinx-build
```
maartenbreddels, line 6: I'll revert this since it's very specific to use of venv
```suggestion
SPHINXBUILD   = sphinx-build
```

docs/source/conf.py
```diff
@@ -174,7 +174,7 @@ def system(cmd):
 extra_navbar = "<h1>dsadsa</h2>"
 
 html_sidebars = {
-	"**": ["sbt-sidebar-footer.html", "sbt-sidebar-nav.html", "sidebar-search-bs.html"]
+	"**": ["sbt-sidebar-footer.html", "sbt-sidebar-nav.html", "searchbox.html"]
```
maartenbreddels, line 177: why was this needed?

docs/source/installing.rst
```diff
@@ -44,7 +44,7 @@ If you want to work on vaex for a Pull Request from the source, use the followin
 * make sure the dev versions of pcre are installed (e.g. ``conda install -c conda-forge pcre``)
 * install using (note: if you're on Windows, make sure that your command line/terminal has administrator privileges):
 
- * ``pip install -e .``  (again, use (ana)conda or virtualenv/venv)
+ * ``make init`` or ``pip install -e .[dev]``  (again, use (ana)conda or virtualenv/venv)
```
maartenbreddels, line 47: zsh friendly
```suggestion
 * ``make init`` or ``pip install -e ".[dev]"``  (again, use (ana)conda or virtualenv/venv)
```

maartenbreddels (commented):
yeah, good idea, really want to get this in

maartenbreddels:
old
image
vs new:
image
I think I prefer the old one

docs/source/conf.py
maartenbreddels:
```suggestion
	"**": ["sbt-sidebar-footer.html", "sbt-sidebar-nav.html", "sidebar-search-bs.html"]
```
lets revert (see my comment)

Alon-Alexander:
What are we waiting for?
Is there anything else which requires change?

maartenbreddels:
No, was waiting for CI, failures are unrelated (tensorflow version issues, if you want to take a look at that, feel free).
Will merge, thanks a lot!

PR #1375, opened 2021-05-26
Update imports and names in fits.py

meredith-durbin:
Hello, another quick fix to the FITS reader from me! Adds a few missing imports and updates leftover pre-v4.0 naming conventions.


maartenbreddels:
Thanks, I would have loved to see some tests for this for future stability, but I currently don't have the time for that. I'll merge it for now, so it's fixed in the next release.

PR #1376, opened 2021-05-27
💚 ml spec was changed

maartenbreddels:


maartenbreddels:
and somehow the `packages/vaex-ml/vaex/ml/spec_new.json` got into the repo, we don't need that

JovanVeljanoski:
I probably messed up at some point..

maartenbreddels:
np, we only need to take a look at py37 and then we should be able to get green ci again :)

PR #1377, opened 2021-05-27
🐛 column names equal to function names

maartenbreddels:
cc @antoinerg who got bitten by a column called 'float', this now works:
```python
import vaex
import os
with open("data.csv", "w") as fd:
    fd.write("id,float\n0,7.5\n1,8.5\n")
df = vaex.from_csv('data.csv')
print(df["float"].nunique())
df
```


maartenbreddels:
CI issue fixed in #1376

Issue #929, opened 2020-08-11
Groupby on multiple columns 


abaradia1978:
Hi,

I am trying to use groupby on a dataset with 200mn rows (13GB of data). It's a HDF5 file and has 20 columns. It works fine if I perform groupby on 3-4 columns. But if I try to perform groupby on 8-10 columns and aggregate (sum) on one column, it triggers a very high CPU and Memory usage and finally exists the process. I am using Jupyter notebook to do this activity on a server with 8 cores and 100GB of RAM. 

I don't consider this to be a very large dataset and server size should have ideally managed this load. Or am I missing something?

I even tried to split the vaex dataframe in 4 equal size and perform the same operation on one split dataframe. Still it would break because of high CPU/Memory requirement.

The command is as under:

`df.groupby(by=['product_code','product_category','product','sub_product','business_segment','region','state','city','c_year','c_quarter','c_month'],agg={ 'cnt' : vaex.agg.count() } )`

JovanVeljanoski:
Hi, 

Perhaps can you are hitting the case discussed here: https://github.com/vaexio/vaex/issues/919

Can you check that you can hold the output dataframe (the grouped dataframe) in memory?

Cheers, 
J.

abaradia1978:
Unique values in my dataset are not more than 50 in each column on which groupby is performed. And also data does not exists for all combinations of values across these columns. I ran below code to see uniqueness across all these columns.

```
def combine(a,b,c,d,e,f,g,h,i,j,k):
    return a + '_' + b + '_' + c + '_' + d + '_' + e + '_' + f + '_' + g + '_' + h + '_' + i + '_' + j + '_' + k

df['combined'] = combine(df.product_code,df.product_category,df.product,df.sub_product,df.business_segment,df.region,df.state,df.city,df.c_year,df.c_quarter,df.c_month)

df.combined.nunique()
```
The count was only 48896. The granular data is 13GB and RAM available is 100GB. Isn't this enough to group data?

I also noticed very high CPU usage which at times reaches 90% and then it crashes.

JovanVeljanoski:
Hi, 

I am not sure I am correct about this.. but say you are doing a groupby on  just two columns, where column "a" has 2 unique values and column "b" has 3 unique values. Say you do just one aggregation on top of that groupby. The output would be a dataframe with 6 rows (2 * 3), and 3 columns (one for the aggregation specified and 2 for the columns you are grouping by). 

[...]

Issue #602, opened 2020-02-18
Kernel died when using lots of column in groupby


ramadhiantoagung:
I.e. if i have below table

```
from itertools import product

listA =list(range(1, 14))
listB =list(range(1, 26))
listC =list(range(1, 23))
listD =list(range(1, 25))
listE =list(range(1, 11))
PDF = pd.DataFrame(list(product(listA, listB,listC ,listD,listE )), columns=['listA', 'listB','listC','listD','listE'])
```

above will create pandas dataframe with 2mio row

`VDF = vaex.from_pandas(PDF)`

convert above into vaex

`VDF.groupby(VDF.listA,agg='count')
`
above are fast

`VDF.groupby([VDF.listA,VDF.listB],agg='count')
`
above still fast but considerably slower

`VDF.groupby([VDF.listA,VDF.listB,VDF.listC],agg='count')
`
above are really slow

`VDF.groupby([VDF.listA,VDF.listB,VDF.listC,VDF.listD],agg='count')
`
above will kill the kernel

the funny thing is, dask and pandas easily do that. Do you have any lead about this behaviour?

right now im using a loop to limit the groupby into only 2 item, but its adding considerable time

Btw this is coming from real financial data, but since im not able to release said data, this is equal representative. the real data is~50times bigger since each combination will have several period and multiple lines

Thankss

JovanVeljanoski:
Hi, 

Thank you for the report, but unfortunately, I can not reproduce the issue. With your example, this is what I am getting (on a 2013 macbook pro (13"))

<img width="1094" alt="image" src="https://user-images.githubusercontent.com/18574951/74715647-240c1900-522d-11ea-8ecf-07ba3b9e6f4b.png">

JovanVeljanoski:
I also tried the same exercise on 100M rows, since you suggest that is the size of your real data. 
The groupby happen in 1 second for me..

![image](https://user-images.githubusercontent.com/18574951/74723522-0645b080-523b-11ea-9edf-cca0743b8a75.png)

ramadhiantoagung:
Hi, ill try to obfuscate my real data, will get back ya

ramadhiantoagung:
Hi, ill try to obfuscate my real data, will get back ya

jplauri:
I just gave this a try on Windows 10 and I was also unable to reproduce the issue.

pietrolesci:
Hi all, 

[...]

PR #1381, opened 2021-06-02
🐎 perform groupby in a sparse way for less memory usage/performance (~250 times faster)

maartenbreddels:
```python
import numpy as np
import vaex
n = 10_000
x = np.random.rand(n)
y = np.random.rand(n)
z = np.random.rand(n)
df = vaex.from_arrays(x=x, y=y, z=z)
```

before:
```python
%%time
dfg = df.groupby(['x', 'y'], agg=vaex.agg.sum('z'))
#CPU times: user 4.16 s, sys: 5.94 s, total: 10.1 s
#Wall time: 10.1 s
```

after:
```python
%%time
dfg = df.groupby(['x', 'y'], agg=vaex.agg.sum('z'))
# CPU times: user 44.7 ms, sys: 2.52 ms, total: 47.2 ms
# Wall time: 38.7 ms
```

which is about ~250 times faster. A lot of time was wasted on allocating memory as well.

There are situations where this is slower, so we may want to release this with `assume_sparse=False` first.

cc @nicolaskruchten

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -319,6 +319,7 @@ class DataFrame(object):
 
     def map_reduce(self, map, reduce, arguments, progress=False, delay=False, info=False, to_numpy=True, ignore_filter=False, pre_filter=False, name='map reduce (custom)', selection=None):
         # def map_wrapper(*blocks):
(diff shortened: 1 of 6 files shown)


maartenbreddels:
Fixes #602
Fixes #929

Issue #1366, opened 2021-05-22
Leave logging configuration to client code


ynouri:
**Description**

The `import vaex` statement enforces an opinionated logging configuration. This prevents the client code to set a different logging configuration. For example:

```
import logging
import vaex

logging.basicConfig(level=logging.INFO, format="%(levelname)s --> %(message)s")
logger = logging.getLogger(__name__)

if __name__ == "__main__":
    logger.info("Logging config is usually left to client code.")
```
does output:
```
INFO:MainThread:__main__:Logging config is usually left to client code.
```
instead of:
```
INFO --> Logging config is usually left to client code.
```

**Source code responsible for this behavior**

https://github.com/vaexio/vaex/blob/master/packages/vaex-core/vaex/__init__.py#L626

**Suggested solution**

Encapsulate the logging configuration *recommended* by Vaex in a function that is not executed at module import. This leaves the user freedom to choose to use the Vaex logging config, or their own logging config.

For example:

```
import vaex
vaex.set_logging_basic_config()
```

**Thanks**

Thank you for this awesome and exciting project!

JovanVeljanoski:
Hi, 

This is a very reasonable request and something that I agree that should be done. Would you be comfortable to taking this on and doing a PR?

Thanks!

maartenbreddels:
I agree, we should not do that. Might be convenient to have some env vars to control this. e.g.:
```
$ VAEX_LOG=debug python myvaex.py
```

PR #1384, opened 2021-06-03
🐛 Apply formatting to named logger

robocioaca:
I've done the minimal intervention necessary to comply with the issue, as raised. However, the [consistent logging issue in a library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) might be a more involved discussion, and possibly requiring architecture input from you... 

* The VAEX logging will happen through the named logger object.
* The formatter applies to the named logger instead of the root one.
* Switched the obsolete FATAL loglevel to ERROR

Fixes vaexio/vaex#1366

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -592,44 +592,54 @@ def zeldovich(dim=2, N=256, n=-2.5, t=None, scale=1, seed=None):
     return vaex.file.other.Zeldovich(dim=dim, N=N, n=n, t=t, scale=scale)
 
 
+# create named logger, for all loglevels
+logger = logging.getLogger('vaex')
+logger.setLevel(logging.DEBUG)
+
+# create console handler and accept all loglevels
+ch = logging.StreamHandler()
+ch.setLevel(logging.DEBUG)
+
+# create formatter
+formatter = logging.Formatter('%(levelname)s:%(threadName)s:%(name)s:%(message)s')
+
+# add formatter to console handler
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Great work! I've always been confused about the logging framework and how it should be used by developers, so your input is very much appreciated.

> However, the [consistent logging issue in a library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) might be a more involved discussion, and possibly requiring architecture input from you

Could you elaborate on that (possibly in a new issue, since I'll merge this for now when green)

robocioaca:
Thank you for the speedy response and kind words, @maartenbreddels !

In case you decide that sorting out the logging is a suitable refactoring step, I'll gladly do the work. As I see things now, These are the steps I'll go through:

1. I'll go first and see how do the **mature** libraries from the NumFOCUS projects implement logging. Maybe there is a golden standard to glean from them.
2. Write a proposal for you to consider - and publish it in issue form as advised.
3. It's highly likely that the refactoring will lightly touch many files in the vaex sub-modules; which might be a _hairy_ task for my current level of fluency with the project. I'd like to get better acquainted with the testing/CI infrastructure you have in place here (try and make as robust changes as possible).

However, in the meantime, there are many issues to work on, already vetted and prioritized by the support team. My first reflex would be to battle some of them, to get some experience in the whole development process setup in vaex.

maartenbreddels:
Note that logging is now easier to configure/disable with https://github.com/vaexio/vaex/pull/1811

Issue #1388, opened 2021-06-09
[BUG-REPORT] IPython version checks when using development version of IPython.


dalejung:
**Description**
vaex errors on `check_ipython` when `IPython` is installed with `develop`. When IPython is installed with `pip install -e` a `.dev` is appended to the version number. The check assumes that the `IPython.__version__` will always be numbers.

`8.0.0.dev` is the breaking version for me. 

```
      6 IPython = sys.modules.get('IPython')
      7 if IPython:
----> 8     IPython_version = tuple(map(int, IPython.__version__.split('.')))
      9     if IPython_version < (7, 0, 0):
     10         raise RuntimeError(f'You are using IPython {IPython.__version__} while we require 7.0.0, please update IPython')

ValueError: invalid literal for int() with base 10: 'dev'
```

**Software information**
 - Vaex version:
 {'vaex': '4.1.0',
 'vaex-core': '4.1.0',
 'vaex-viz': '0.5.0',
 'vaex-hdf5': '0.7.0',
 'vaex-server': '0.4.0',
 'vaex-astro': '0.8.0',
 'vaex-ui': '0.3.0',
 'vaex-ml': '0.11.1',
 'vaex-graphql': '0.2.0'}

 - Vaex was installed via: from source
 - OS: Linux

**Additional information**

maartenbreddels:
Fixed in https://github.com/vaexio/vaex/pull/1389

PR #1389, opened 2021-06-09
Fix version compare when running development IPython which has versions like 8.0.0.dev. #1388

dalejung:
Simple fix to handle dev versions like 8.0.0.dev. Assume we only want the first 3 version bits.


maartenbreddels:
Thanks Dale!

PR #1391, opened 2021-06-09
✨ limit rows in groupby using early stop.

maartenbreddels:
This can be useful if you only want to do a groupby when the result
is small and otherwise don't care about the result.

cc @nicolaskruchten

See usage:
image


nicolaskruchten:
Great! How early is "early" in this case? like 10% of the time to do the whole group-by or... ?

maartenbreddels:
As soon as it can, even before doing the aggregation, example:
image

It takes 500ms to do the whole groupby on 1.1 billion rows, but if you limit to 2 rows, it will finds it out in 25ms.

nicolaskruchten:
Awesome, this looks like exactly what the doctor ordered :)

PR #1401, opened 2021-06-15
Fixes for shift related methods

JovanVeljanoski:
This contains fixes for various shift related issues:

- [ ] fix issues when `diff` is used for certain hdf5 files


maartenbreddels:
It did not have to do with hdf5, but with longer dataframes, causing a chunking, which exposed a bug.

PR #1402, opened 2021-06-16
🐛 column names equal to function names in selections/filters

maartenbreddels:
Similar to #1377 cc @antoinerg  @nicolaskruchten


maartenbreddels:
CI issues is solved in another PR

Issue #1413, opened 2021-06-18
NaNs are not correctly exported to HDF5 format


humcasma:
A dataframe containing NaNs is correctly exported to, e.g., CVS format. However, when exporting to HDF5, the NaNs disappear.

Running
```
df1 = vaex.from_pandas(pd.DataFrame([1,2,3], columns=['idx']))
df2 = vaex.from_pandas(pd.DataFrame({'idx':[5,6,7], 't1': [0,0,0]}))
df3 = df1.join(df2, how='left', on='idx', rsuffix='_y')
```
results in
```
#	idx	idx_y	t1
0	1	--	--
1	2	--	--
2	3	--	--

```
However, after exporting to and reading from HDF5:
```
df3.export('test.hdf5')
vaex.open('test.hdf5')
```
we have
```
#	idx	idx_y	t1
0	1	999999	999999
1	2	999999	999999
2	3	999999	999999
```

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: {'vaex-core': '4.3.0.post1', 'vaex-hdf5': '0.8.0'}
 - Vaex was installed via: pip
 - OS: Linux

JovanVeljanoski:
Hey @humcasma 

Thank you for submitting this and for the reproducible example! This kind of issues help a lot, so thank you!

I will try to make unit-test this weekend, and hopefully we can get to fix this soon!

humcasma:
A bit more of info on when this problem happens...
For what I have seen, it affects integer and date columns. If there is an integer column that is only partially filled after a join, when saving to hdf5 the datatype of the column is kept as integer, but since hdf5 does not support integers with missing (NaNs) values (if I am not wrong), the NaNs are filled in. If converting to float before saving, the problem disappears for integers.

Ben-Epstein:
Can confirm on 4.8.0 this is still an issue

Ben-Epstein:
@humcasma For what it's worth, if you need a workaround and must keep the nulls, this works
```
import vaex
import pandas as pd
import pyarrow as pa

df1 = vaex.from_pandas(pd.DataFrame([1,2,3], columns=['idx']))
df2 = vaex.from_pandas(pd.DataFrame({'idx':[1,6,7], 't1': [0,0,0]}))
df3 = df1.join(df2, how='left', on='idx', rsuffix='_y')

df3["idx_y"] = df3.idx_y.astype(pa.int64())
df3["t1"] = df3.idx_y.astype(pa.int64())
[...]

PR #1418, opened 2021-06-18
Fix export (hdf5) of joined dataframe which contains missing values

JovanVeljanoski:
This PR address the issue raised in https://github.com/vaexio/vaex/issues/1413

Notes:
- The added test checks the export against hdf5, arrow, parquet and csv. Only hdf5 and csv report problems, but for different reasons;
- HDF5 test fails because the data seems to be saved incorrectly;
- The csv file is exported correctly. However when the data is read, the missing values are converted do `nan`. I think this should now automatically be turned into a missing (masked) values, not `np.nan`. What do you think @maartenbreddels @xdssio. This might be more tricky tho.. so I am happy to leave it as is and adjust the test. Just wondering what the ideal case should be - in my opinion missing values.

Checklist:
- [x] Add tests
- [ ] Make tests pass

### DIFF
--- a/tests/export_test.py
+++ b/tests/export_test.py
@@ -171,3 +171,15 @@ def test_export_csv(df_local, tmpdir):
     df.export_csv(path)
 
     assert '123456' in vaex.open(path)
+
+@pytest.mark.parametrize("filename", ["test.hdf5", "test.arrow", "test.parquet", "test.csv"])
+def test_export_joined_dataframe(tmpdir, filename):
+    df1 = vaex.from_dict({'x': [1, 2, 3]})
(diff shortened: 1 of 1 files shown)


Ben-Epstein:
fwiw i believe the issue with hdf5 stems from the fact that you cannot have "missing" (nan) values in int dtypes in numpy.

```
import numpy as np

np.array([1,2,3,4,None], dtype=np.int64)

>>> TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

Converting the column to something non-int would work, but that may not be the desired outcome.
```

PR #1421, opened 2021-06-19
Fix: filtering using a column of changed dtype

JovanVeljanoski:
This solves the issues raised in: https://github.com/vaexio/vaex/discussions/1420

Checklist:
- [x] Make a unit test
- [ ] Make the test pass


maartenbreddels:
Ok, the issue seems more like str->int/float conversion. Do you agree?

Issue #1368, opened 2021-05-24
[FEATURE-REQUEST]:Porting str.extract functionality of pandas


datascientistpur:
Hi Team,

There is a 'str.extract' function that exists in pandas.The main use is to extract relevant groups of text from string.

The same is not available in vaex.Is it feasible to port it?

PFB the functionality
![image](https://user-images.githubusercontent.com/62737832/119312412-f3b4f500-bc8f-11eb-8d16-2d0d2821a2b9.png)

maartenbreddels:
Hi,

did you see #1154 and the comments? Maybe that solves your issue as well.

cheers,

Maarten

datascientistpur:
Thanks,will give it a shot

Issue #1154, opened 2021-01-13
[FEATURE-REQUEST] Regex extract and bit shifting


mansenfranzen:
Hi Vaex-Community,

I'm fairly new to Vaex and I really like the progress you're making!

I've been using pandas and spark a lot in the past but have come across several performance issues with both recently. I'm tempted to checkout vaex for a current use case but I'm not sure of the required functionality is already present. I went through the documentation and also consulted closed/open issues without success.

### Regex extract
In pandas there is [`extract`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.extract.html) and in spark there is [`regexp_extract`](http://spark.apache.org/docs/latest/api/python/pyspark.sql.html?highlight=regex#pyspark.sql.functions.regexp_extract) to extract one or more values from a string via a regular expression. I couldn't find anything comparable to this in vaex. Is there anything related already?

### Bit shifting
In spark there is [`shiftLeft`](http://spark.apache.org/docs/latest/api/python/pyspark.sql.html?highlight=regex#pyspark.sql.functions.shiftLeft) to shift a value by a given number of bits. Is there anything comparable in vaex?

Thanks a lot!

maartenbreddels:
Hi,

thanks!
There are several functions using regex, but the base problem was that we did not have a data structure that would support multiple values, now that we support Apache Arrow, and together with the upstream work at https://github.com/apache/arrow/pull/8459 we will have this in the near future.

All standard Python operators are supported,e.g.:
![image](https://user-images.githubusercontent.com/1765949/105049546-0cf12f00-5a6d-11eb-8424-41882f52b25a.png)

Regards,

Maarten

maartenbreddels:
You can probably use https://vaex.io/docs/api.html#vaex.expression.StringOperations.replace for this (ctrl-f for regex should work)

crliu95:
> Hi,
> 
> thanks!
> There are several functions using regex, but the base problem was that we did not have a data structure that would support multiple values, now that we support Apache Arrow, and together with the upstream work at [apache/arrow#8459](https://github.com/apache/arrow/pull/8459) we will have this in the near future.
> 
> All standard Python operators are supported,e.g.:
> ![image](https://user-images.githubusercontent.com/1765949/105049546-0cf12f00-5a6d-11eb-8424-41882f52b25a.png)
> 
> Regards,
> 
> Maarten

Hi Maarten,

[...]

PR #1423, opened 2021-06-21
Add `str_extract_regex` to dataframe str function namespace.

mansenfranzen:
This PR addresses issue #1154 and #1368 and is work in progress since there are conceptual questions remaining.

Adding a wrapper for apache arrows [`extract_regex`](https://arrow.apache.org/docs/cpp/compute.html#string-extraction) under the vaex dataframe string namespace is straightforward. However, handling the returned result is somewhat more complicated because it is not a single column of a primitive type but rather a collection of primitive types (StructArray in Arrow). There may be different valid approaches here which I would like to discuss with you before going any further (creating tests):

1. **Plain Arrow StructArray**: Just keep the standard arrow struct array and do not modify it any further. Arrow's struct arrays are currently not properly converted to dask/numpy within vaex (calling `to_dask_array`/`to_numpy`). This could be solved however (using structured numpy arrays perhaps for numpy). Interestingly, conversion to pandas works fine already (via dicts). There is one big question remaining, however. How does a common vaex user access the fields of the underlying struct via the vaex expression object? For example, how to apply some further string modifications on extracted values being contained in a vaex expression? My first attempt would be to use the arrow struct array directly instead of using the vaex expression layer (this may be due to my lack of familiarity with vaex).

2. **Decompose/flatten**: Instead of returning a single expression, provide a python dictionary where names correspond to capture groups identifiers and values refer to vaex expressions of primitive types. This would solve the issue of having complex data types being contained in a single vaex expression. However, this would break the common usage patterns of most string functions which return a single vaex expression instead of a dictionary of vaex expressions.

3. **Support both** - plain struct array and flattening via function parameter or separate functions: In my opinion both approaches serve valid use cases and should be supported. The default behaviour could be the plain struct array being contained in a single vaex expression. Additionally, there could be parameter like `flatten` or a separate utility function to flatten/decompose a expression holding a complex data type into multiple expressions containing only primitive types.

What is your opinion?

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -1945,6 +1945,28 @@ def str_extract_regex(x, pattern):
                         the regular expression '(?P<letter>[ab])(?P<digit>\\d)'.
     :returns: an expression containing a struct with field names corresponding to capture group identifiers.
 
+    Example:
+
+    >>> import vaex
+    >>> email = ["foo@bar.org", "bar@foo.org", "open@source.org", "invalid@address.com"]
+    >>> df = vaex.from_arrays(email=email)
+    >>> df
+    #  email
+    0  foo@bar.org
+    1  bar@foo.org
+    2  open@source.org
+    3  invalid@address.com
+
+    >>> pattern = "(?P<name>.*)@(?P<address>.*)\.org"
+    >>> df.email.str.extract_regex(pattern=pattern)
+    Expression = str_extract_regex(email, pattern='(?P<name>.*)@(?P<addres...
+    Length: 4 dtype: struct<name: string, address: string> (expression)
+    -------------------------------------------------------------------
+    0      {'name': 'foo', 'address': 'bar'}
+    1      {'name': 'bar', 'address': 'foo'}
+    2  {'name': 'open', 'address': 'source'}
+    3                                     --
+
     """
 
     if not isinstance(x, vaex.array_types.supported_arrow_array_types):
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -1936,6 +1936,22 @@ def str_split(x, pattern=None, max_splits=-1):
         return pc.split_pattern(x, pattern=pattern, max_splits=max_splits)
 
 
+@register_function(scope='str')
+@auto_str_unwrap
+def str_extract_regex(x, pattern):
+    """Extract substrings defined by a regular expression using Apache Arrow (Google RE2 library).
+
+    :param str pattern: A regular expression which needs to contain named capture groups, e.g. ‘letter’ and ‘digit’ for
+                        the regular expression '(?P<letter>[ab])(?P<digit>\\d)'.
+    :returns: an expression containing a struct with field names corresponding to capture group identifiers.
+
+    """
+
+    if not isinstance(x, vaex.array_types.supported_arrow_array_types):
+        x = pa.array(x)
+
+    return pc.extract_regex(x, pattern=pattern)
+
 
 @register_function(scope='str')
 @auto_str_unwrap
--- a/tests/strings_test.py
+++ b/tests/strings_test.py
@@ -300,6 +300,17 @@ def test_string_replace_regex_unicode(dfs, pattern, replacement, flags):
            dfs.s.str_pandas.replace(pattern, replacement, flags=flags, regex=True).tolist()
 
 
+def test_string_extract_regex():
+    ds = vaex.from_arrays(email=["foo@bar.org", "bar@foo.org", "open@source.org", "invalid@address.com"])
+    pattern = "(?P<name>.*)@(?P<address>.*)\.org"
+
+    expr = ds.email.str.extract_regex(pattern=pattern)
+    assert expr.tolist() == [{"name": "foo", "address": "bar"},
+                             {"name": "bar", "address": "foo"},
+                             {"name": "open", "address": "source"},
+                             None]
+
+
 @pytest.mark.parametrize("sub", ["v", unicode_compat("æ")])
 @pytest.mark.parametrize("start", [0, 3, 5])
 @pytest.mark.parametrize("end", [-1, 3, 5, 10])


maartenbreddels:
> 3\. The default behaviour could be the plain struct array being contained in a single vaex expression.

I think this is the way to go, struct support is already trickling in: https://github.com/vaexio/vaex/blob/b1c79858852db94a66f986160add4f27593f929f/packages/vaex-core/vaex/datatype.py#L330

I think it would be nice to have some example in the docstring, and a test in tests/strings_test.py.

I would say that `Expression.__getitem__` would be the proper way to extract a field from a struct, e.g.:
```python
df.x[:, 'city'] -> gets a field from a struct
df.x[:, ['city', 'country']] -> 'projects' a struct to a different struct
```

@JovanVeljanoski: what do you think?

If we do this, we should include this in the tutorial I think.

cheers,

Maarten

mansenfranzen:
Hi Maarten,

thanks for the quick response :-).

> > 3. The default behaviour could be the plain struct array being contained in a single vaex expression.
> 
> I think this is the way to go

Agreed. Accordingly, this PR solely adds the `extract_regex` function which returns a struct array and leaves the more complex struct support to other PRs.

> I think it would be nice to have some example in the docstring, and a test in tests/strings_test.py.

Sure, I'm going to add them since the result type is now fixed.

> I would say that `Expression.__getitem__` would be the proper way to extract a field from a struct, e.g.:
> 
> ```python
> df.x[:, 'city'] -> gets a field from a struct
> df.x[:, ['city', 'country']] -> 'projects' a struct to a different struct
> ```

Just to provide another flavor of how pyspark handles field access of structs using dot notation which I would prefer for getting a single field from a struct (even though I dislike mixing up methods and value access under the same namespace...):

```python
df.x.city -> gets a field from a struct
```

maartenbreddels:
I like that, but that would limit the field names to valid Python identifiers. We have done this for the dataframe class, so I kinda like it, but we still need a way to get other fields, similar to `df.x` vs `df['x']`

JovanVeljanoski:
Actually I would prefer to not have dot notation in this case. 
While we have done in for the dataframe class to get column names, and it is often convenient.. we get too many cases in which column names (or keys in this case) have something like "word.anotherword", and I see this often with various project that I work on.

So my vote goes for what Maarten suggested originally.

mansenfranzen:
### Dot notation

@maartenbreddels and @JovanVeljanoski I fully agree - dot notation has its drawbacks because it requires valid python identifiers and it potentially shadows existing dataframe methods/attributes if the column name is `values` for example.

### Test and example
Anyway, I added an example and a simple test case which just confirms that the function wrapping works as expected. It makes no sense to test any regex specifics here since this is left to Arrow. 

### Modification of `numpy_dtype`
Additionally, I modified `packages/vaex-core/vaex/array_types.py` to provide a fallback in case an arrow type has no pandas type equivalence (when calling `arrow_type.to_pandas_dtype()`). This does not change the semantics but rather covers an edge case. If pyarrow will get an equivalence for `StructArray` for pandas in the future, the code should remain to work. This change was necessary to make `Expression._repr_plain_()` work. Otherwise, it fails with `NotImplementedError` and nothing is shown in jupyter notebook when printing the expression instance containing the regex result.

### Remaining limitations

As noted before, `to_numpy` and `to_dask_array` do not work due to missing struct array support. This should be dealt with in another PR.

### Site note regarding running tests

Running tests locally really required me to look into the CI. A simple `pytest` invocation does not work. Few things that needed to be done before testing succeeds:

- install test packages (pytest, gcsfs, s3fs, graphviz) which are missing by default [dev install](https://vaex.io/docs/installing.html#for-developers)
- explicitly provide targets (test directories and files) otherwise pytest starts tests in vendored packages which fail due to missing deps

I hope to provide a PR for this anytime soon.

maartenbreddels:
> As noted before, `to_numpy` and `to_dask_array` do not work due to missing struct array support. This should be dealt with in another PR.

Sounds good, i think there are many other code paths that probably cannot handle that. For now, also the hdf5 export cannot support it, but I want to support that in the future.



> Running tests locally really required me to look into the CI. A simple `pytest` invocation does not work. Few things that needed to be done before testing succeeds:

@Alon-Alexander did some great work on that, it just needs a little love, maybe you can also take a look at that PR https://github.com/vaexio/vaex/pull/1365/files

maartenbreddels (approved):
Looks good, I am ok with merging this, except that we cannot yet extract a single column from this, e.g. `df['x'][:,'name']` do you want to do this in this PR, or a different PR?

mansenfranzen:
I think it would be more sensible to create a new PR to extract single columns from a struct array.

### Separate namespace for struct operations
As an idea, vaex already has separate namespaces for string `str`, datetime `dt` or timedelta `td` operations. For clarity and explicitness, vaex could add a separate namespace for operations on struct like dtypes. It could include methods like `get_field_names` or `get_field_by_name` or similar along with the shorthand notation (or syntactic sugar) of `df['x'][:,'name']`.

### Expected result type of `df['x'][:,'name']` 
Concerning the implementation, I'm still struggling with the vaex internals. The code snippet `df['x'][:,'name']` should return an instance of type `Expression`, right? It should be an expression/virtual column in order to further process it (let's we want to split and extracted string value). It does not make sense to return the underlying arrow string array here because we want to keep the lazyness and characteristics of the expression object.

### Need help
Assuming we want to return an expression instead of an arrow string array, I tried to understand the `Expression` class.
As far as I understand, it contains the processing steps (like `2*x+1`) as a plain string representation (attribute `expression`) which in turn is converted to an AST (attribute `ast`). Finally, the AST is used to compute the actual result. Essentially, I can manually create an Expression by just providing the string representation:

```python
import vaex
from vaex.expression import Expression

ds = vaex.example()
Expression(ds, "(x+2)")
```

Result:

```
Expression = (x+2)
Length: 330,000 dtype: float32 (expression)
-------------------------------------------
     0    3.23187
     1     1.8363
     2  -0.120256
     3    6.71559
     4    9.21719
       ...       
329995    3.99387
329996    5.71809
329997    2.36885
329998    1.88741
329999    22.7962
```

However, I got lost somewhere following the path to the actual execution engine (Expression.values -> Expression.evaluate() -> DataFrame.evaluate() -> DataFrame._evaluate_implementation() -> DataFrame.map_reduce() -> Tasks, Executer, Scheduler etc...). Unfortunatley I hit the wall of complexity here and got lost along the way. The main question is: what is the correct syntax for field access of a struct that the execution/evaluation engine understands?

Please feel free to correct me if I've already came to some wrong assumptions here. Any documentation regarding the `Expression` class and how the actual execution works can really help me.

Also, this entire discussion should be actually moved to the separate PR or issue.

maartenbreddels:
> vaex could add a separate namespace for operations on struct like dtypes

Yes, I agree

> The code snippet `df['x'][:,'name']` should return an instance of type `Expression`, right?

Yes, I'd expect it to return sth like `Expression(f'struct_<nice_name>({self.expression}, 'name')` for this case. Then I expect a a function `struct_<nice_name>` in functions.py (or make functions_struct.py and import it at the end of functions). 

https://github.com/vaexio/vaex/pull/1185 could also be a good example.

Yeah, Vaex can become quite complicated due to the expressions, I don't have time to explain it now, but I hope with the example PR you can managed to get this working.

mansenfranzen:
The struct array support is dealt with in #1447. This PR is completed in my opinion. Only thing remaining may be an addition to the changelog.

mansenfranzen:
Just rebased to current master. Since #1447 has been merged, the current functionality can be safely used now.

maartenbreddels:
Really nice to see this PR this small and elegant!

PR #1433, opened 2021-06-24
🐎 tasks refactor (allows more caching)

maartenbreddels:


maartenbreddels:
@JovanVeljanoski there seems to be a vaex-ml failure due to precision (new releases of sklearn?) Maybe you can fix that using a less precise check?

maartenbreddels:
Some of the notebooks failed because the spec changed for vaex-server, but I upgrade the server on the fly, and 1 of the runs passed 🥳

Issue #1404, opened 2021-06-16
open() fails with misleading error message upon unknown file extension


humcasma:
The following code fails on the second line with error message `OSError: Cannot open dataframe.hdf nobody knows how to read it.`:

```
df.export_hdf5('dataframe.hdf')
vaex.open('dataframe.hdf')
```

Renaming `dataframe.hdf` to `dataframe.hdf5` and running `vaex.open('dataframe.hdf5')` works fine.
If `open()` does not know the type of file from its extension and it is not possible to determine it from the file's content, it could try all possible ways it knows of opening a file, and only raise that error message if all attempts fail. Alternatively, it could fail with a more descriptive error message: `Unknown file extension. I only recognise these extensions: ...`. `export_hdf5()` should also complain, in first place, about the unknown file extension.

JovanVeljanoski:
Hi, 

Yeah this is a good point! `vaex.open` relies heavily on the files having the correct extension. 

It might be too involved on the short term to have each file checked / tried against all the openers.. but for an unknown/unsupported extension, a more helpful message along the lines of what you suggest would very good indeed!

@humcasma are you willing to do a PR for this perhaps?

humcasma:
Hi @JovanVeljanoski  - Unfortunately I am quite busy with a project deliverable these days, so I have no time to propose a solution right now. When I get some time, if the solution is not yet implemented , I would not mind doing it.

PR #1437, opened 2021-06-28
Allow opening of valid files with non-standard extensions.

JovanVeljanoski:
See https://github.com/vaexio/vaex/issues/1404 for discussions on this.


maartenbreddels:
Cool, thanks!

PR #1438, opened 2021-06-28
Small improvements to reduce warnings.

JovanVeljanoski:
Small changes throughout the codebase aimed to reduce various types of warnings that can be safely avoided.

### DIFF
--- a/tests/ml/ml_test.py
+++ b/tests/ml/ml_test.py
@@ -536,7 +536,12 @@ def test_kbinsdiscretizer(tmpdir, strategy):
                                y=[1, 2, 5, 6, 10, 9])
 
     trans = vaex.ml.KBinsDiscretizer(features=['x', 'y'], n_bins=3, strategy=strategy)
-    df_train_trans = trans.fit_transform(df_train)
+
+    if strategy == 'quantile':
+        with pytest.warns(UserWarning):
+            df_train_trans = trans.fit_transform(df_train)
+    else:
+        df_train_trans = trans.fit_transform(df_train)
     df_test_trans = trans.transform(df_test)
 
     if strategy == 'quantile':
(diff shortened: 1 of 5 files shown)


tests/ml/ml_test.py
```diff
@@ -536,7 +536,12 @@ def test_kbinsdiscretizer(tmpdir, strategy):
                                y=[1, 2, 5, 6, 10, 9])
 
     trans = vaex.ml.KBinsDiscretizer(features=['x', 'y'], n_bins=3, strategy=strategy)
-    df_train_trans = trans.fit_transform(df_train)
+
+    if strategy == 'quantile':
+        with pytest.warns(UserWarning):
```
maartenbreddels, line 541: Please use match to make sure it's giving an expected warning.

JovanVeljanoski: good idea - done!

tests/ml/ml_test.py
```diff
@@ -536,7 +536,12 @@ def test_kbinsdiscretizer(tmpdir, strategy):
                                y=[1, 2, 5, 6, 10, 9])
 
     trans = vaex.ml.KBinsDiscretizer(features=['x', 'y'], n_bins=3, strategy=strategy)
-    df_train_trans = trans.fit_transform(df_train)
+
+    if strategy == 'quantile':
+        with pytest.warns(UserWarning, match='Bins whose width are too small'):
```
maartenbreddels, line 541: For some reason, this doesn't trigger on Linux, and causes CI to fail.

JovanVeljanoski: yeah i tested it also, and on a mac it also does not show a warning anymore.. maybe related to a numpy version.. Anyway now removed, but I found some other warnings to get rid off :)

maartenbreddels:
the ml spec changed :)

PR #1447, opened 2021-07-01
Add basic support for accessing struct arrays

mansenfranzen:
This PR continues work described in #1423. It adds support to access individual fields of arrow based struct arrays and wraps them in vaex expressions.

The following additions are made:

- Adds `StructOperations` namespace to `Expression`.
- Adds `Expression.struct.get_field` and `Expression.struct.project`.
- Adds `Expression.struct.field_names` and `Expression.struct.field_types`
- Adds shorthand notation to use  `struct_get_field` and `struct_project` via `Expression.__getitem__`.

Still to do and to be discussed:

- [ ] Add examples to documentation.
- [ ] Add changes to changelog.

Please feel free to modify the function names. I'm not 100% sure about the API.


maartenbreddels:
I gave it a quick review, and so far it looks really good! I think I agree with the API, it's close to the Arrow naming etc, so I like that.

What I do suggest is to move most of your code into vaex/struct.py, which will probably move to vaex/cpu/struct.py in the future. We'd need to import it at the bottom of functions.py I think, to force registration.

Regarding API, I'd like @JovanVeljanoski to also take a look.

maartenbreddels:
If you rebase, CI should get greener :)

JovanVeljanoski (commented):
Hi, 

This is great work! Excellent start, and definitely an area where we can improve! I am very happy that you are taking this on!

I really like the tests and the docstrings!

Here are some comments from me regarding the api. I also did some testing, etc..
 - I understand arrow calls these things "fields", but I wonder if calling them "keys" instead is better. Ideally, (personally) i do not want the end user to know they are dealing with arrow structures (even if they are), just to use vaex, as close to python (and perhaps pandas) as possible. And as far as I can see, struct arrays are very similar to dictionaries in Pythong, so why not adopt the nomenclature from there: `get_field` -> `get` (since there is nothing else to get?), `field_names` -> `keys` and `field_types` -> `dtypes` (this makes it consistent with other methods available elsewhere in vaex and other libraries). My point is I don't want the user to learn new things if it can be avoided, and something that can be intuitively reused perhaps it should be? Or perhaps I am missing something major here..
 - I am not a big fan of the shorthand notation, but maybe others would like it. I kind of like the "one way to do things", but if it does not add too much maintenance overhead going forward i don't mind it at all
 - currently displaying struct expressions in jupyter / ipython does not work
 - currently `df.dtypes` does not show anything (or crashes on `print`) when struct array is present in the dataframe
 - currently `df.x.dtype` does not show anything (or crashes on `print`) with `x` being struct array
 - I would prefer if `df.x.struct.field_types` returns a pandas series, like `df.dtypes` normally would. I know we don't want to depend on pandas here, and long term we want to improve here, but for now, it would be nice if things were consistent. (Also tied to the `df.x.struct.dtypes` comment)
 - Finally (at first I thought this is a bug, but i see it is a feature :D), i'd like to discuss the behaviour when a user wants to get a key that is not existent. Now this raises an exception (and is tested!), but I wonder if `df.x.struct.get_field('not_there')` should return a column of masked values? Or perhaps this does not make sense? Maybe (like a python dict), using the [] notation calling a key/field that does not exist should raise an exception, but using get should return something (or at least have an optional argument for a default value if key is not found)?

maartenbreddels:
> * I understand arrow calls these things "fields", but I wonder if calling them "keys"

fields can be non-unique, keys convey they are unique.

> * I am not a big fan of the shorthand notation

I am :)

> * I would prefer if `df.x.struct.field_types` returns a pandas series, like `df.dtypes` normally would. I know we don't want to depend on pandas here

I dunno, but also here.. it's non-unique, which is kinda of a pain, which means also dicts don't cut it.


What about `.field_names`, `field_types` that returns lists, and `.fields` return a dict, which should raise when there are duplicate field names?

JovanVeljanoski:
While pyarrow allows fields to indeed be non-unique, vaex (currently) does not. Also exporting the data to pandas or python lists gives incorrect results for repeating fields. 

Is there a reason why this is even allowed? ( I mean repeating fields)

mansenfranzen:
First of all, thanks for all the feedback :smiley:.

### Move to `struct.py` and register

> What I do suggest is to move most of your code into vaex/struct.py, which will probably move to vaex/cpu/struct.py in the future. We'd need to import it at the bottom of functions.py I think, to force registration.

I moved the struct functionality into a separate module (`vaex/struct.py`). Additionally, I needed to separate the `register_function` function from the `functions.py` to `vaex/registry.py` in order prevent circular imports when registering the newly added `vaex/struct.py`. I guess it could be beneficial to split the `functions.py` even more into smaller modules (e.g. one for each scope).  In my opinion it seems reasonable to have the register function at a separate place from those functions using it to be registered. 

The actual registration (importing the code that is registering functions) is now done via `vaex.__init__.py`. I guess it could be sensible to move the registration to a separate module, too.

### Rebase 

> If you rebase, CI should get greener :)

Done - I hope it's better now but I'm getting lots of errors when testing locally (`pyarrow.lib.ArrowInvalid: Needed to copy 1 chunks with 0 nulls, but zero_copy_only was True...`). Is this specific to my PR (which I think it shouldn't be) or is this a known issue at the moment?

### API and naming convention

> get_field -> get (since there is nothing else to get?), field_names -> keys and field_types -> dtypes

I agree. Vaex supports both numpy and arrow memory backends even though structs are currently only supported for arrow. However, numpy support could be added and perhaps at some point in time there will be even a new and better backend (not arguing that arrow is bad at all!). Hence, the API should be backend agnostic and python-like. That's why I already adjusted the naming convention here accordingly.

### Displaying and dtype issue
> currently displaying struct expressions in jupyter / ipython does not work
currently df.dtypes does not show anything (or crashes on print) when struct array is present in the dataframe
currently df.x.dtype does not show anything (or crashes on print) with x being struct array

It does work now ([fix](https://github.com/vaexio/vaex/pull/1447/files#diff-abb6583a69adcc7bcd412945dbef00a3cc3f9ef455a2ca17f61dac2b7c579a23R200)). It already worked in the previous, related PR #1423 and I just didn't want to copy the fix due to duplication. However, this fix is even more relevant in this PR as it directly addresses basic struct functionality. I added tests for all 3 cases to ensure they remain to work in the future.

### `pd.Series` instead of python dict
>  would prefer if df.x.struct.field_types returns a pandas series, like df.dtypes normally would.

Agreed - done.

### Discussion: raising an error if field does not exist
> Finally (at first I thought this is a bug, but i see it is a feature :D), i'd like to discuss the behaviour when a user wants to get a key that is not existent. Now this raises an exception (and is tested!), but I wonder if df.x.struct.get_field('not_there') should return a column of masked values?

In my opinion this should raise an error since the field simply does not exist. What would be the use case to return a column of masked values?

### Discussion: shorthand notation
> I am not a big fan of the shorthand notation, but maybe others would like it. I kind of like the "one way to do things", but if it does not add too much maintenance overhead going forward i don't mind it at all

I'm kind of used to it due to using pandas. However, in pandas it is used on dataframes, not on columns containing struct like data structures. However, the concept remains the same - there is a row and column/field identifier being contained in a single expression which I like. However, to be consistent, the shorthand notation should also exist on vaex dataframes to select rows and columns via bracket notation. I thought this would be the case but actually its not. 

### Pyarrow Copy vs. View

When using the `struct_project` method (e.g. selecting two from three fields of a struct), I assume the newly created struct does not create a copy but only view, right? I haven't tested it, yet. I'm sure you're more familiar with it. 

### Minor remarks

The diff for `functions.py` is a bit messed up because pycharm decided (without my confirmation...) to drop unused imports. Sorry for that.

maartenbreddels:
> While pyarrow allows fields to indeed be non-unique, vaex (currently) does not.

You are comparing arrow field names with column names I'm assuming? I don't think because column names are unique, that we should not support non-unique fields for arrow struct arrays. We don't want to end up in a situation where we cannot support a particular arrow/parquet/feather file because we didn't support it right.

What about inheriting `collections.abc.Mapping`, and if we have duplicates keys, we could raise an error, and provide an extra interface for that edge case.

maartenbreddels:
@mariobuikhuizen suggested the idea of renaming the field names, to make them unique, so if we have `a, b, a`, we would rename it to `a, b, a_1`.

JovanVeljanoski:
Seems reasonable to me. I guess with the similar strategy as the renaming of invalid column names we do now?

I mean.. if you do `get_field('a')` and there are two or more fields called `a`, which one would you get, and how to get the other one(s). Anyway, renaming or even raising exception saying that something is not supported seems ok to me.

maartenbreddels:
> I guess with the similar strategy as the renaming of invalid column names we do now?

we don't do that any more, but that is the idea.

Well, if we ingest a struct array with two fields named 'a', we would rename the second, so you'd always get the first.

mansenfranzen:
Renaming columns may get into trouble if the new, renamed name (e.g. `a -> a_1`) is also already present. Additional rules would be required to handle this case, too. Increasing the number suffix may not help if this naming scheme was already used to generate the original field names.

What about using the unique index identifiers when dealing with duplicated field names in struct arrays in arrow? In case a field name is ambiguous, force the user to use index positions instead of field labels. 

Interestingly, this is how arrow works at the moment. Accessing a field with duplicated label raises a key error while using the index position works fine:

```python
import pyarrow as pa

array = pa.StructArray.from_arrays(arrays=[[1, 2, 3], 
                                           [4, 5, 6], 
                                           [7, 8, 9]], 
                                   names=["col1", "col1", "col3"])

# get mapping between field idx and field label
index_label_mapping = {idx: field.name for idx, field in enumerate(array.type)}
index_label_mapping # {0: 'col1', 1: 'col1', 2: 'col3'}

# raises key error!
array.field("col1")

# works
array.field("col3")

# works -> integers map to field index positions
array.field(0)

# works
array.field(1)

# works
array.field(2)
```

However, this approach again is very specific to arrow and probably won't apply to other backends. @maartenbreddels @JovanVeljanoski Do you consider to add numpy support for structs or do you intend to solely rely on arrow for structs here? If arrow is the only way, then we could be build the solution around arrow specifics (e.g. using index positions).

By the way, the conversion from arrow to pandas succeeds silently in arrow without raising an explicit warning at the moment:

```python
array.to_pandas()

0    {'col1': 4, 'col3': 7}
1    {'col1': 5, 'col3': 8}
2    {'col1': 6, 'col3': 9}
dtype: object
```

Well, how often does it occur that you have duplicated field names anyway? If someone really requires them, force them to rename before wrapping them with arrow or request the user to provide an explicit index position to label mapping.

maartenbreddels:
If the fallback is using index, that sounds reasonable to me, and in line
with arrow. I think bumpy structs could be interested to support, bit that
would match this idea I think.

(from mobile phone)

mansenfranzen:
@maartenbreddels In agreement with your last comment, I added support to handle duplicated field names. In case a duplicated field is accessed via its label, a value error is thrown with relevant information to use index based access instead. To get the mapping between field labels and field indicies, I added the two more struct properties `indices` and `keys_indices`. The error message above has a hint to use `keys_indices`.

Both label and index position based access are now completely supported. When using `project`, you may even mix label and index position based access.

I also needed to intercept how a struct with duplicated field names is printed (jupyter) in `formatting.py` in order to prevent getting errors from pyarrow which currently does not support standard conversion methods (e.g. `as_py()`) for structs with duplicated field names. Even using native pyarrow does not work to get a print representation for a struct array with duplicated field names. 
Please note, I added the index position in parenthesis in the print output to make fields with duplicated field names distinguishable.

maartenbreddels:
See https://issues.apache.org/jira/browse/ARROW-13158 and https://github.com/apache/arrow/pull/10591 for the duplicate field names issue in pyarrow itself

tests/struct_test.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def test_struct_project_invalid_field(df):
+    with pytest.raises(ValueError):
+        df.array.struct.project(["doesNotExist"]).tolist()
+
+
+def test_struct_project_invalid_dtype(df):
+    """Ensure that struct function is only applied to correct dtype."""
+    with pytest.raises(TypeError):
```
maartenbreddels, line 121: Odd, I'd expect a KeyError, I guess this is Arrow's exception we're propagating here. Also, if used with an integer, it should be an IndexError, they seem to have a super class (https://docs.python.org/3/library/exceptions.html#LookupError) should we catch those issues ourselves, and throw an appropriate exception, or would a TypeError make sense (since the keys are part of the type).

packages/vaex-core/vaex/expression.py
```diff
@@ -196,6 +197,140 @@ def __init__(self, expression):
         self.expression = expression
 
 
+class StructOperations(object):
```
maartenbreddels, line 200: No need to inherit from object, what about inheriting from https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping ?

packages/vaex-core/vaex/expression.py
```diff
@@ -196,6 +197,140 @@ def __init__(self, expression):
...
+        >>> df.array.struct.keys_indices
+        0    a
+        1    b
+        dtype: object
+
+        """
+
+        struct = self.expression.values
+        self.assert_struct_dtype(struct)
+        return pd.Series(self.keys, index=self.indices)
```
maartenbreddels, line 286:
```suggestion
        return pd.Series(self.keys, index=self.indices, dtype='string')
```

maartenbreddels (commented):
Excellent work, also the test suite looks very thorough.

However, I am not sure the indices and keys_indices really add anything. I see myself using sth like
```
for index, key in enumerate(foo.keys()):
```
But I don't see a use for keys_indices, but please try to convince me otherwise.

Also, I think if we make keys a method, and inherit from abc.Mapping, we'll have a much more standard interface (which might fail for duplicate field names, but that's why we have a custom keys method).

cheers,

Maarten

tests/struct_test.py
mansenfranzen: In this specific test case I basically want to ensure that struct operations are only applied to expressions/columns holding a struct dtype. An appropriate exception message is given that the used method is not valid on non struct dtypes.

Concerning tests checking for non existing labels or index positions (e.g. `test_struct_get_invalid_field`), I agree that is more appropriate to use `LookupError` instead of `TypeError`. An appropriate exception message about valid lookup values is already given.

packages/vaex-core/vaex/expression.py
mansenfranzen: I inherited from object to be consistent with the other operation classes (e.g. `StringOperations`). I agree, there shouldn't be a need to inherit from object because python 2 support for new style classes is probably not required anymore.

I really like the idea of inheriting from `collections.abc.Mapping` however, as you've pointed out, we get in trouble with duplicated field names here. Implementing `__getitem__` to handle duplicated fields should work just fine, but `__iter__` should return index positions (since they are truly unique) instead of labels which might be confusing when doing things like `for field, dtype in struct.items()` because field will be an integer instead of a string label. Perhaps we make things more complicated when trying to force the duplicated keys data type in the mapping interface. What do you think?

mansenfranzen:
@maartenbreddels Thanks for your feedback :-)

> See https://issues.apache.org/jira/browse/ARROW-13158 and apache/arrow#10591 for the duplicate field names issue in pyarrow itself

Good to see that this issue is already resolved in pyarrow. However, we will need to keep custom type checking and formatting logic here because `as_py()`will keep raising an error for duplicated field names. Instead, we may use pyarrows' `_as_py_tuple()` in future versions which is essentially very similar to my hackish formatting approach. So I guess for now we can keep my current solution as it supports also older pyarrow versions.

> But I don't see a use for keys_indices, but please try to convince me otherwise.

I agree, I wouldn't use it myself but I'm thinking of users who are not very familiar with python and for whom it is not obvious on how to get the index positions for corresponding field labels. Moreover, we have an implicit assumption here that keys are already ordered correctly regarding their index position. Therefore I thought such a property would fit in nicely. But it's up to you to decide - you have a way better feeling for what API is appropriate in vaex. I'm happy with or without it ;-).

I'll wait for your comments until applying code changes.

packages/vaex-core/vaex/expression.py
maartenbreddels: I think we can support Mapping, but all methods should raise a KeyError when there are duplicate keys, and hint the user to use integers for keys. So in 90% of the cases, users can just use the Mapping interface.



> but `__iter__` should return index positions

I think it should return the keys (strings) but raise a KeyError in the case of duplicate keys.

The most robust code would then be (and used internally in Vaex):
```python
for i in range(len(df.x.struct)):
  name = df.x.struct.keys()[i]
  expression = df.x.struct[i]
  ...

# or
for i, name in enumerate(df.x.struct.keys()):
  expression = df.x.struct[i]
  ...

```

But you can often get away by doing:
```python
for name, expression in df.x.items():
  ...
```

@JovanVeljanoski and @mansenfranzen do you both agree with this?

We can take this a step further.. and also implement items to make it work with duplicate keys, the only thing that wouldn't work would be `__getitem__` with a string argument.

tests/struct_test.py
maartenbreddels: @JovanVeljanoski strong opinions on this?

maartenbreddels:
> So I guess for now we can keep my current solution as it supports also older pyarrow versions.

👍 



> Moreover, we have an implicit assumption here that keys are already ordered correctly regarding their index position.

Yes, I think we should make that assumption, or rather state this explicitly.

I'm in favour of removing the code, the less code the better.

tests/struct_test.py
JovanVeljanoski: No strong opinion, but my preference in line with the @mansenfranzen describes above. 

regarding the key, indices, key_indices and the like, also no strong preference, but I am more in line with @maartenbreddels's approach here. I'm in not bloating up the API too much, unless you think it is quite common to access and use these things. 

Finally, speaking of duplicate keys, when they exist, displaying the dataframe in Jupyter is a bit funny (dunno if that can be helped.. maybe you are already aware of it..).

image

mansenfranzen: @maartenbreddels  @JovanVeljanoski 

Agreed, let's drop `key_indices` and explicitly state the assumption of correctly ordered field-index positions in docs.

Regarding the print output: it looks indeed a bit funny. Those backslashes are not present under my ubuntu 20.04 environment but I'll take a closer look at it. 

Anyhow, we have to agree on how print output should be formatted when duplicated fields are present. In my first attempt, I added the index position in parenthesis after the field name to distinguish duplicated fields. However, it looks odd to me right now. I assume there are different possibilities:

a) both field name and index position: `{'col1(0)': 1, 'col1(1)': "a", 'col3(2)': 4}`
b) fields only: `{'col1': 1, 'col1': "a", 'col3': 4}`
c) index only: `{0: 1, 1: "a", 2: 4}`

Currently I favor fields only because it is consistent even though duplicated fields occur. However, since we assume ordered field-index positions, one can still distinguish those values. What do you think?

JovanVeljanoski: Hey, 

I am inclined to agree with you, regarding the fields only approach. I fear otherwise would be confusing, and we've had in the past people that had that of the type "column_name[x]" as much as it was surprising to me..

packages/vaex-core/vaex/expression.py
mansenfranzen: Sounds reasonable to me. Here is a little prototype implementation which might help us to discuss it in more detail:

```python
import collections
import pyarrow as pa

values = [[1, 2, 3],
          [4, 5, 6],
          [7, 8, 9]]
names = ["col1", "col1", "col3"]
array = pa.StructArray.from_arrays(arrays=values, names=names)


class StructOperations(collections.abc.Mapping):
    def __init__(self, array):
        self.array = array
        
        keys = list(self.keys())
        self._duplicates = set([x for x in keys if keys.count(x) > 1])
               
    def __iter__(self):
        for name in self.keys():
            yield name
    
    def __getitem__(self, key):
        if key in self._duplicates:
            raise LookupError("Duplicated label: please use index position.")
        return self.array.field(key)
    
    def __len__(self):
        return len(self.array.type)
    
    def keys(self):
        return (field.name for field in self.array.type)
    
    def values(self):
        return (field.type for field in self.array.type)
    
    def items(self):
        return zip(self.keys(), self.values())
    
    def get(self, key):
        return self[key]
    

mapping = StructOperations(array)

mapping.get(0) # <pyarrow.lib.Int64Array object at 0x0000024AE5719408> [1, 2, 3]
mapping.get("col1") # LookupError
mapping.get("col3") # <pyarrow.lib.Int64Array object at 0x0000024AE5719D08> [4, 5, 6]

list(mapping.keys()) # ['col1', 'col1', 'col3']
list(mapping.values()) # [DataType(int64), DataType(string), DataType(int64)]
list(mapping.items()) # [('col1', DataType(int64)), ('col1', DataType(string)), ('col3', DataType(int64))]
```

I allowed string lookup values for `__getitem__`, too. The only way to run into a key error is indeed when using a string for duplicated fields. Otherwise, all methods work fine with duplicated field names. 

@maartenbreddels @JovanVeljanoski 
What do you think?

tests/struct_test.py
mansenfranzen: @JovanVeljanoski What would be your favorite way to represent structs with duplicated field names?

JovanVeljanoski: Fields only (option "b" from your previous message). If those are the names, we show them as they are.. 
In general I'd like to show/see the data as it is.

packages/vaex-core/vaex/expression.py
maartenbreddels: Yes, I think this would be great to use (sorry for the late reply, vacation period :))

mansenfranzen:
@maartenbreddels @JovanVeljanoski I modified the PR according to our latest discussions. I hope the string formatting looks normal now but I assume it is still not correct yet. I may need to take second look at it.

tests/struct_test.py
mansenfranzen: @JovanVeljanoski Can you please check if you still have the strange formatting output for structs with duplicated fields?

mansenfranzen:
Just rebased on current master. Somehow tests in `tests/strings_test.py` are still failing under ubuntu 20.04.

JovanVeljanoski:
Hi, 

Great stuff! For me in the jupyter notebook the string formatting looks alright. I don't see anything weird with it anymore! Is there anything in particular you'd like me to look at or test?

I just noticed this (my sincere apologies if this was part of the earlier discussion that I missed or misunderstood..): now things like `struct.keys()` and `struct.items()` return generators. Is this what we want? I (would) find this quite impractical myself, since most of the time i just want to get a quick peak at what things are. Do we often expect that the number of keys to be *so*  large so that we should have a generator by default? I was naively just expecting a list out of those methods...

Also:
- `df.x.struct.dtypes` returns a pandas series with the keys as index and their dtypes as values;
- `df.x.struct.items()` returns a generator of tuples where the 1st element is the key and the 2nd element is the dtype
- `df.x.struct.values()` returns a generator of the dtypes of each key

So to me they all return the same information really, in a slightly different way.  I would say only the `struct.dtypes` is needed? Perhaps naively, but I find the latter two methods confusing since their behavior  is different to that of python dict. Especially for `values` since such a method is used elsewhere in vaex, and it was always related to the data values contained within an expression.  Although in all fairness the docstrings do say what these methods do, but I would naively expect different behavour as these things were preprogrammed into me for years :) I hope you understand where I am coming from..

mansenfranzen:
I initially used lists for keys and values, too. However, since collection.abc.Mapping uses generators as return types, I decided to keep it consistent. But you're right, it is more intuitive to have lists right away, especially when working with it interactively in jupyter. On the other side, we could emphasize the use of struct.dtypes in the docs because this is what users are already familiar with and contains all relevant information about the struct. Anyway, I can live with it either way.

Struct.items will need to return tuples otherwise we might get in trouble with duplicated field names when using dicts.

mansenfranzen:
@JovanVeljanoski regarding the print formatting, if the previous weird behavior with backslashes is gone for structs with duplicated field names, it's all good and no further testing is required.

mansenfranzen:
I replaced the generators with lists and also changed `TypeError` to `LookupError` in case of invalid lookups (which we've agreed on earlier but I forgot to modify). 

@maartenbreddels @JovanVeljanoski Please check if there are any changes/improvements required from your perspective. If not, I might continue with adding a small section in the docs.

JovanVeljanoski:
Hey! 

Great stuff! I have one final thing from my side, this is the existence of `struct.values()` and `struct.items()`. I would argue that these are redundant, give the existence of `struct.dtypes`.  The former can be obtained as `struct.dtypes.tolist()` and the later as `list(df.x.struct.dtypes.items())`. 

I tested everything else (albeit with small data), and it works great!

mansenfranzen:
@JovanVeljanoski I agree that `struct.values()` and `struct.items()` are redundant. 

Since we're inheriting from `collections.abc.Mapping` for `StructOperations`, those methods along with `struct.keys()` are provided by this base class anyway. However, since we need to take care of duplicated keys, we need to overwrite them. 

Currently, the implementation of `struct.dtypes` actually relies on using  `struct.keys()` and  `struct.values()` wrapping both of them in a pandas series. In this sense, `struct.dtypes` is kind of the main redundant part here. But we want to keep `struct.dtypes` because users are familiar with it. In my opinion, we should emphasize the usage of `struct.dtypes` in the documentation.

maartenbreddels:
> Since we're inheriting from `collections.abc.Mapping` for `StructOperations`, those methods along with `struct.keys()` are provided by this base class anyway. However, since we need to take care of duplicated keys, we need to overwrite them.

I agree.

Please remove the WIP, I think this is (nearly?) good to go!

maartenbreddels:
Hmm, I think I misunderstood something. I don't see why the .values() return dtypes, I expected them to return the expressions that evaluate to that column, like `df.x[:,i]`.

packages/vaex-core/vaex/expression.py
```diff
@@ -196,6 +197,148 @@ def __init__(self, expression):
...
+        >>> df.array.struct.values()
+        [int64, string]
+
+        """
+
+        self._assert_struct_dtype()
+        dtypes = (field.type for field in self._array.type)
+        vaex_dtypes = [DataType(x) for x in dtypes]
+
+        return vaex_dtypes
```
maartenbreddels, line 282: I expected (since it's a mapping)
```suggestion
        return [self[i] for i in range(len(self))]
```

mansenfranzen: Good point - I guess we've misunderstood us here. From my perspective, the main purpose of the `StructOperations` object was to represent meta data (field names & types) and to provide a namespace for all struct relevant functions. The mapping I was concerned about was between field names and field types.

If I want to get the actual values of the struct, I would rather use `df.x` (returning the vaex expression) or `df.x.values` (returning the pyarrow struct) instead of `df.x.struct.values()` (which would return a python list of pyarrow arrays in your version). However, I totally agree that the name `values` is misleading when in fact dtypes are returned. It depends on the kind of mapping we're thinking of (*field names -> types* vs. *field names -> arrays*). Both implementations are somewhat not satisfying really. Your proposal adds yet another way to access the data adding a 3rd return type while my approach is somewhat misleading regarding the naming convention and what a user might expect.

What about having `df.x.struct.values()` return `df.x.values`? This would be consistent in my opinion even though it makes `df.x.struct.values()` completely redundant. Hence, there is one way to return the actual vaex expression via `df.x` and two consistent ways to return the underlying data via `df.x.values` or `df.x.struct.values()`.

JovanVeljanoski: The is exactly what I was referring to with my earlier comment about redundancy. 
I would tend to agree that `df.x.struct.values()` being the same as `df.x.values` is redundant, and does not really add anything in my opinion. 

@mansenfranzen if you like/need to keep the functionality of the curret `.struct.values()` that's fine, but perhaps  with a different name than `values` as that seems to be "reserved" for functionality of the type `df.x.values`. 

As for returning the data, i would cast my vote to just having `df.x.values` (which would return arrow thing). Dont know if the output of that would support something like `.tolist()` that would convert the arrow structure to list.. if there is, i say we are done. Otherwise maybe we can add that on our end?

maartenbreddels: > I would tend to agree that `df.x.struct.values()` being the same as `df.x.values` is redundant, and does not really add anything in my opinion.

This is not the case right.

df.x.struct has a mapping interface, which should behave like a dict, which means that if `df.x.struct['col1']` gives an expression, `df.x.struct.values()` should give a list of expressions right?

JovanVeljanoski: Yeah I agree with @maartenbreddels regarding `df.x.struct.values()`

But, i also do think that `df.x.values` should always exist, as there should be a uniform way to extract the raw data for any kind of column..

mansenfranzen: Yeah, you've convinced me ;-). Honestly, my initial perspective of having a mapping for field names to field types for `StructOperations` seems rather inappropriate to me now. 

> if you like/need to keep the functionality of the curret .struct.values() that's fine, but perhaps with a different name than values as that seems to be "reserved" for functionality of the type df.x.values.

@JovanVeljanoski No need to keep that functionality. 

> df.x.struct has a mapping interface, which should behave like a dict, which means that if df.x.struct['col1'] gives an expression, df.x.struct.values() should give a list of expressions right?

@maartenbreddels You're right. I'm going to change it accordingly (probably tomorrow).
> Also, please don't confuse df.x.values with this, which should never be there :) (my personal opinion.. :))

The `df.x.values` seems to have some controversial opinions here ;-). Is there an issue or PR discussing it in more detail to follow along?

maartenbreddels: > but perhaps with a different name than values as that seems to be "reserved" for functionality of the type df.x.values.

@JovanVeljanoski 
Just be be sure, `df.x.struct.values()` should implement what the mapping interface dictates (a list/sequence of Expressions), it has zero relation to `df.x.values` right

JovanVeljanoski: Yeah, absolutely, we agree there. 
I was just saying that `df.x.values` has its place.

maartenbreddels:
@mansenfranzen many many thanks on this excellent work!

@JovanVeljanoski I just realized the Struct accessor is not in the docs, maybe you can add this to api.rst?

JovanVeljanoski:
Yeah, I will do it unless @mansenfranzen beats me to it. I was planning a "docs" pr to document the few of the more recent changes/additions, and this was gonna be part of it.

mansenfranzen:
@maartenbreddels @JovanVeljanoski I'm glad I could contribute back a litte ;-). Thanks for your patience with this PR. Go ahead with the docs, too. 

I think #1423 can be now safely merged, too.

PR #1461, opened 2021-07-10
Remove old example folder from root & add new plotting examples in docs as a replacement.

JovanVeljanoski:
This PR removes the old "examples" folder from the root of the project. Those notebooks are very outdated and not maintained. 

Instead, the some of that material has been refactored and now exists as an "advanced plotting examples" in the documentation pages.


maartenbreddels:
Yeah, makes sense, thanks for cleaning up!

Issue #1361, opened 2021-05-20
[BUG-REPORT] DataFrame concatenation error with virtual columns


tgdcat:
**Description**
Running the following code will result in an error.
It will succeed if you do not use the virtual column.
```python
import numpy as np
import vaex

col_1 = np.array([1,2])
col_2 = np.array(['test1', 'test2'])
df1 = vaex.from_arrays(a=col_1, b=col_2)
df1['new_col'] = df1.a * 3

col_3 = np.array([3,4])
col_4 = np.array(['test3', 'test4'])
df2 = vaex.from_arrays(a=col_3, c=col_4)

res_df = vaex.concat([df1, df2])

# Exception has occurred: KeyError
# "Virtual column not found: 'new_col'"
#   File "C:\work\test\verify\main.py", line 28, in <module>
#     res_df = vaex.concat([df1, df2])
```
**Software information**
 - Vaex version {
      'vaex-core': '4.1.0',
      'vaex-hdf5': '0.7.0'
  }
 - Vaex was installed via: pip 
 - OS: Windows (Version 2004, build 19041.985)
 - Python 3.8.7

**Additional information**
Adding'new_col' to df2 succeeds, but I'd like to find another idea if possible.

maartenbreddels:
Hi, what is the expected behavior?

tgdcat:
Hi, thank you for your reply.

The expected behavior is to 'concat' a DataFrame with virtual columns.
Currently, an error will occur if 'new_col' (virtual column) exists in df1 and 'new_col' does not exist in df2.

Specifically, I want the following result when I run the code in the first post.
```
  #    a  b        new_col    c
  0    1  'test1'  888        --
  1    2  'test2'  999        --
  2    3  --       --         'test3'
  3    4  --       --         'test4'
```

tgdcat:
I installed Vaex 4.4.0 and tried it, and the problem was fixed.
  
Thank you for helping me.
@maartenbreddels 
@SyureNyanko

PR #1462, opened 2021-07-10
concat with the virtual column when not exist in the other dataframe

SyureNyanko:
Hi, this is my first PR to vaex.
This solves the issues raised in:  #1361
I want to fix vaex so that concat will succeed even if the name of the virtual column exists on one side and the column with that name does not exist on the other side. 
For example,
```python
df1 = vaex.from_arrays(x=np.arange(3))
df2 = vaex.from_arrays(x=np.arange(3))
df1['z'] = df1.x ** 2

df = df1.concat(df2)
#  -> KeyError: "Virtual column not found: 'z'"
```
In above example, df1 is
| # | x |  z
| ---- | ---- | ---- | 
| 0 | 0  |  0  |
| 1 |  1  |  1  | 
| 2 |  2  | 4  |

df2 is 

| # | x |  
| ---- | ---- | 
| 0| 0  |  
| 1|  1  |  
| 2|  2  |  

so, I expected df will be

| # | x |  z
| ---- | ---- | ---- | 
| 0  | 0  |  0  |
| 1 |  1  |  1  | 
| 2 |  2  | 4  |
| 3 | 0  |  --  |   
| 4 |  1  | -- |   
| 5 |  2  | -- |  

but, it ends in  KeyError: "Virtual column not found: 'z'".  

As I read the source of concat method in dataframe.py, I found that df._lazy_materialize(name) was called with an argument that does not exist in the df. So, I modified to check whether the column name exists in the Dataframe, in concat processing of virtual column.

Please let me know if I'm wrong, because I may not have the right way to fix it.

Cheers.

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -5853,7 +5853,8 @@ class DataFrameLocal(DataFrame):
                     # we have a mismatching virtual column, materialize it
                     for df in dfs:
                         # upgrade to a column, so Dataset's concat can concat
-                        dfs[dfs.index(df)] = df._lazy_materialize(name)
+                        if name in df.get_column_names(virtual=True, hidden=True):
+                            dfs[dfs.index(df)] = df._lazy_materialize(name)
 
         first, *tail = dfs
         # concatenate all datasets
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Sorry for the late reply (vacation period), this is really good work, and this indeed the right approach, thanks a lot! And thanks @JovanVeljanoski for pinging me.

PR #1469, opened 2021-07-17
🐛 xpression.minmax() not 

kmcentush:


tests/agg_test.py
```diff
@@ -509,3 +509,11 @@ def test_agg_count_with_custom_name():
     df_grouped = df.groupby(df.x, sort=True).agg({'mycounts': vaex.agg.count(), 'mycounts2': 'count'})
     assert df_grouped['mycounts'].tolist() == [3, 2, 1]
     assert df_grouped['mycounts2'].tolist() == [3, 2, 1]
+
+
+def test_minmax_ts(df):
```
JovanVeljanoski, line 514: I'd prefer that in addition to this, the test asserts for the min and max against the actual min and max values. It could happen that `minmax` is consistent with `min`  and `min`  but still objectively wrong.

kmcentush: Good point. Not sure why I didn't have that check in the first commit.

kmcentush:
Somehow missed all of those tests failing. Will check again on this PR later tonight.

maartenbreddels:
yeah, seems odd, you could try a rebase.. although I don't remember seeing these issues before.

kmcentush:
The latest commit should fix the issue. I was returning a list instead of the np array like before. Also, my previous implementation didn't respect delayed execution, so I changed the code to use the previous structure and just swapped out the `self.limits()` call for `min()` and `max()`, respectively.

This may suggest that the root case bug in `limits()` (or more likely, `StatOpMinMax`?) is still present for my added test case. But as stated above in the original PR comment, this now makes `StatOpMinMax` dead code.

kmcentush:
Not sure why it's failing on non-Windows builds... The errors in the Linux and Mac builds are notebook timeout errors.

JovanVeljanoski:
Ah don't worry about it. Those errors are unrelated. I hope we can clean them up soon. 
I don't think that the notebooks run on the windows CI which is why that part does not fail :). 

Looks good to me! But let's wait for @maartenbreddels to also give his blessing :)

Thanks!

maartenbreddels:
Thanks for this PR!
However, this reminds me that we actually need to have a minmax aggregator, now we lose a bit of performance by doing a min and a max separately.

Issue #1478, opened 2021-07-28
[BUG-REPORT] Unable to export dataframe to hdf5 when it includes multibyte named column (on Windows OS)


SyureNyanko:
Hi, everyone

**Description**

Runtime Error at h5py when saving dataframe includes many columns, and one of them contains Multibyte string.
It might only happen on Windows. (It didn't happen when I tried it on MacOS and Ubuntu )

```python
import vaex
source_dict = {"東京": [1, 2, 3], "a": [1, 2, 3], "b": [1,2,3], "c": [1, 2, 3], "d": [1, 2, 3], "a1": [1, 2, 3], "b2": [1,2,3], "c3": [1, 2, 3], "d4": [1, 2, 3]}
df = vaex.from_dict(source_dict)
df.export_hdf5("testfile.hdf5")
```
```
Traceback (most recent call last):
  File ".\main.py", line 8, in <module>
    df.export_hdf5("testfile.hdf5")
  File "C:\Users\naohiro.heya\.pyenv\pyenv-win\versions\3.8.7\lib\site-packages\vaex\dataframe.py", line 6486, in export_hdf5
    writer.write(self, chunk_size=chunk_size, progress=progress, column_count=column_count)
  File "C:\Users\naohiro.heya\.pyenv\pyenv-win\versions\3.8.7\lib\site-packages\vaex\hdf5\writer.py", line 40, in __exit__   
    self.close()
  File "C:\Users\naohiro.heya\.pyenv\pyenv-win\versions\3.8.7\lib\site-packages\vaex\hdf5\writer.py", line 31, in close      
    self.h5.close()
  File "C:\Users\naohiro.heya\.pyenv\pyenv-win\versions\3.8.7\lib\site-packages\h5py\_hl\files.py", line 457, in close       
    h5i.dec_ref(id_)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5i.pyx", line 153, in h5py.h5i.dec_ref
RuntimeError: Can't decrement id ref count (unable to extend file properly)
```

**Software information**
[...]

PR #1479, opened 2021-07-28
fix: export dataframe to hdf5 when it includes multibyte named column (on Windows OS) 

SyureNyanko:
Hi👋
This solves the issues raised in: #1478 
In export_hdf5 method, the procedure **was** as follows. 
1. Create/Open hdf5 file (in init function at Writer class)  
2. Open mmap file (in layout method)
3. Close hdf5 file (in close method)
4. Close mmap (in close method)

I made a hypothesis that this procedure might be not a good in windows environment and I read vaex/h5py/hdf5 source code. I didn't found the bug logic on the h5py/hdf5 source, but I was able to reproduce this bug in small scripts without vaex. 

```python
import h5py
import mmap

path = "mytestfile.hdf5"
f = h5py.File(path, "w") # Open HDF5 File =================
file = open(path, "rb+")  # Open mmap file ------------------------
fileno = file.fileno()
cg = f.require_group("column")

kwargs = {}
kwargs["access"] = mmap.ACCESS_READ 
mmap = mmap.mmap(fileno, 0, **kwargs)

for i in range(100):
    g = cg.require_group("東京"+str(i)) # create group hdf5 using multibyte name
    dset = g.create_dataset("mydataset", (100,), dtype='i')

file.close() #  Close HDF5 File =================
f.close()
mmap.close() # mmap close ------------------------
```
```
(on windows execution)
Traceback (most recent call last):
  File "C:\Users\heya\vaex-test\hf.py", line 21, in <module>
    f.close()
  File "C:\Users\heya\miniconda3\lib\site-packages\h5py\_hl\files.py", line 466, in close
    self.id._close_open_objects(h5f.OBJ_LOCAL | h5f.OBJ_FILE)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\h5f.pyx", line 352, in h5py.h5f.FileID._close_open_objects
RuntimeError: Can't decrement id ref count (unable to extend file properly)
```
H5py or hdf5 driver might be responsible for this bug, but this way we can avoid it .
1. Create/Open hdf5 file (in init function at Writer class)  
2. Open mmap file (in layout method)
3. Close mmap (in close method)
4. Close hdf5 file (in close method)

Please give me any advice for furthering the investigation or better solution.

Cheers.

### DIFF
--- a/packages/vaex-hdf5/vaex/hdf5/writer.py
+++ b/packages/vaex-hdf5/vaex/hdf5/writer.py
@@ -29,11 +29,11 @@ class Writer:
         self._layout_called = False
 
     def close(self):
-        self.h5.close()
         if self.mmap is not None:
             self.mmap.close()
             self.file.close()
-
+        self.h5.close()
+        
     def __enter__(self):
         return self
 
--- a/tests/export_test.py
+++ b/tests/export_test.py
@@ -116,6 +116,16 @@ def test_export_string_mask(tmpdir):
     df_arrow = vaex.open(path)
     assert df.s.tolist() == df_arrow.s.tolist()
 
+def test_export_unicode_column_name_hdf5(tmpdir):
+    # prepare many columns for multithreaded export
+    src_dict = {"あ": [1, 2, 3], "a": [1, 2, 3], "b": [1, 2, 3], 
+            "c": [1, 2, 3], "d": [1, 2, 3], "a1": [1, 2, 3], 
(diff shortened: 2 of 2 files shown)


JovanVeljanoski:
Thanks! Looks good to me! 
I'd still like @maartenbreddels to have a look!

maartenbreddels:
Great work Naohiro, this looks good to me!
CI failures are unrelated (and solved)

PR #1490, opened 2021-08-04
fix: astropy doesn't like left aligned data

maartenbreddels:
Caused CI to fail with the latest astropy (4.3), triggering:
```
OSError: No SIMPLE card found, this file does not appear to be a valid FITS file
```
Introduced in https://github.com/astropy/astropy/pull/10895


maartenbreddels:
failures are due to the dataframe server

PR #1502, opened 2021-08-11
remove print nan_value from map

fsiola:
A print statement was present in the map function.


maartenbreddels:
Nice catch thanks you!

PR #1509, opened 2021-08-12
Draft for the Dataframe interchange protocol

AlenkaF:
This is the first draft of the Dataframe protocol implementation for Vaex.
It works for

- numerical and boolean dtypes
- categorical columns constructed with `categorize()`

It is erroring with Arrow Dictionary when trying to convert dtype `arrow_type.to_pandas_dtype()`, line 305 in `_dtype_from_vaexdtype`. Do you have any suggestions?

What is still on my todo list:

- [ ] implement `expression.codes` see https://github.com/vaexio/vaex/pull/1503
- [x] materializing virtual columns
- [x] chunk handling
- [x] support of missing values

cc @maartenbreddels @JovanVeljanoski

### DIFF
--- a/tests/dataframe_protocol_test.py
+++ b/tests/dataframe_protocol_test.py
@@ -314,7 +314,7 @@ def assert_buffer_equal(buffer_dtype: Tuple[_VaexBuffer, Any], vaexcol: vaex.exp
     buf, dtype = buffer_dtype
     pytest.raises(NotImplementedError, buf.__dlpack__)
     assert buf.__dlpack_device__() == (1, None)
-    assert dtype[1] == vaexcol.dtype.numpy.itemsize * 8
+    assert dtype[1] == vaexcol.dtype.index_type.numpy.itemsize * 8
     if not isinstance(vaexcol.values, np.ndarray) and isinstance(vaexcol.values.type, pa.DictionaryType):
         assert dtype[2] == vaexcol.index_values().dtype.numpy.str
     else:


packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,489 @@
...
+        # Define how _dtype_from_vaexdtype calculates kind
+        # If it is internal, categorical must stay categorical (info in metadata)
+        # If it is external (call from_dataframe) must give data dtype
+        
+        bool_c = False
+        if self.is_cat:
+            bool_c = True # internal, categorical must stay categorical
+
+        dtype = self._col.dtype
+
```
maartenbreddels, line 270:
```suggestion
            
if dtype.is_encoded:^M
            dtype = dtype.index_type
```
Something like this? If the array is an arrow array, it's the type of the indices you are after right?

maartenbreddels (commented):
Excellent work, I gave it a first pass, hope that helps.

tests/dataframe_protocol_test.py
```diff
@@ -0,0 +1,67 @@
+import vaex
+import numpy as np
+import pyarrow as pa
+from vaex.dataframe_protocol import from_dataframe_to_vaex, _DtypeKind
+
+def test_float_only():
+	df = vaex.from_arrays(x=np.array([1.5, 2.5, 3.5]), y=np.array([9.2, 10.5, 11.8]))
```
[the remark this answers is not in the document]
AlenkaF: This is great. It errors for chunked arrow array, which makes sense. It should pass once I implement the chunking.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,489 @@
...
+            - 2 : sentinel value
+            - 3 : bit mask
+            - 4 : byte mask
+        Value : if kind is "sentinel value", the actual value. None otherwise.
+        """
+        _k = _DtypeKind
+        kind = self.dtype[0]
+        value = None
+        if kind == _k.FLOAT:
+            null = 1  # np.nan
```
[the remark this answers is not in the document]
AlenkaF: That clears up a lot for me, thanks!

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,489 @@
...
+            buffer = _VaexBuffer(self._col.to_numpy())
+            dtype = self.dtype
+        elif self.dtype[0] == _k.CATEGORICAL:
+            codes = self._col.values
+            # If values are subset of labels and are not codes, codes mst be generated
+            if self._col.values[0] in self.labels:
+                for i in self._col.values:
+                    codes[np.where(codes==i)] = np.where(self.labels == i) # if values are same as labels
+            else: 
+                codes = self._col.values # values are already codes for the labels
```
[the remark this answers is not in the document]
AlenkaF: I also think this will not be needed when I will be able to use `expressions.codes.`

But for now, if I take this example:
```python
df = vaex.from_arrays(year=[2012, 2015, 2019], weekday=[0, 4, 6])
df = df.categorize('year', min_value=2012, max_value=2019)
df = df.categorize('weekday', labels=['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'])
```
and use `.values` function on `year` and `weekday` column I get two different types of output. In the first case it is a list of labels and in the other case I get the codes.
```
array([2012, 2015, 2019])
array([0, 4, 6])
```
To get the codes for the `year` column I calculate them as in line 402-404.

For now I can't seem to find another way.

packages/vaex-core/vaex/dataframe_protocol.py
AlenkaF: This works, thank you! Added in new commit.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,725 @@
...
+            - Data types not included: complex, Arrow-style null, binary, decimal,
+              and nested (list, struct, map, union) dtypes.
+        """
+        dtype = self._col.dtype
+        
+        # Categorical
+        # If it is internal, kind must be categorical (23)
+        # If it is external (call from_dataframe), dtype must give type of the data
+        if self._col.df.is_category(self._col):
+            return (_DtypeKind.CATEGORICAL, 64, 'u', '=') # what should be the default??
```
AlenkaF, line 377: This will have to be revised. See first two items in https://github.com/data-apis/dataframe-api/issues/46#issue-930471826

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,725 @@
...
+        dataframe.append(vaex.from_dict(columns))
+        # chunk buffers are added to list of all buffers
+        _buffers.append(_buffers_chunks)
+        
+    if df.num_chunks() == 1:
+        _buffers = _buffers[0]
+    
+    # Join the chunks into tuple for now
+    df_new = vaex.concat(dataframe)
+    df_new._buffers = _buffers
```
AlenkaF, line 92: A nested list may be hard to read. Could be changed to a list of dictionaries with column names as keys if that would be better.

AlenkaF:
@maartenbreddels @JovanVeljanoski don't know if it will stay green =) nevertheless I think this is ready for a review.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,695 @@
...
+    DATETIME = 22
+    CATEGORICAL = 23
+
+    
+def convert_column_to_ndarray(col : ColumnObject) -> pa.Array:
+    """
+    Convert an int, uint, float or bool column to an arrow array
+    """
+    if col.offset != 0:
+        raise NotImplementedError("column.offset > 0 not handled yet")
```
AlenkaF, line 110: Is offset used in Vaex?

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
+"""
+Implementation of the dataframe exchange protocol.
+
+Public API
+----------
+
+from_dataframe : construct a vaex.dataframe.DataFrame from an input data frame which
+                 implements the exchange protocol
+
```
maartenbreddels, line 9:
```suggestion
                 
For a background and spec, see:
   * https://data-apis.org/blog/dataframe_protocol_rfc/
   * https://data-apis.org/dataframe-protocol/latest/index.html
```

maartenbreddels: Maybe these links are good to put in.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+        Return the buffer containing the data and the buffer's associated dtype.
+        """
+        _k = _DtypeKind
+        if self.dtype[0] in (_k.INT, _k.UINT, _k.FLOAT, _k.BOOL):
+            # If arrow array is boolean .to_numpy changes values for some reason
+            # For that reason data is transferred to numpy through .tolist
+            if self.dtype[0] == _k.BOOL and isinstance(self._col.values, (pa.Array, pa.ChunkedArray)):
+                buffer = _VaexBuffer(np.array(self._col.tolist(), dtype=bool))
+            else:
+                buffer = _VaexBuffer(self._col.to_numpy())
```
maartenbreddels, line 537: This could make a copy. We can use `.values/.evaluate()` but then we'd have to support arrow and numpy arrays, which should be fine, maybe the following works:
```suggestion
                buffer = _VaexBuffer(np.asarray(self._col.evaluate(), dtype=bool))
```

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+    def _get_data_buffer(self) -> Tuple[_VaexBuffer, Any]:  # Any is for self.dtype tuple
+        """
+        Return the buffer containing the data and the buffer's associated dtype.
+        """
+        _k = _DtypeKind
+        if self.dtype[0] in (_k.INT, _k.UINT, _k.FLOAT, _k.BOOL):
+            # If arrow array is boolean .to_numpy changes values for some reason
+            # For that reason data is transferred to numpy through .tolist
+            if self.dtype[0] == _k.BOOL and isinstance(self._col.values, (pa.Array, pa.ChunkedArray)):
+                buffer = _VaexBuffer(np.array(self._col.tolist(), dtype=bool))
```
maartenbreddels, line 535: Not sure why that happens, I'll see if a test fails because of that.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+    def __init__(self, df: vaex.dataframe.DataFrame, nan_as_null: bool = False, allow_copy: bool = True) -> None:
+        """
+        Constructor - an instance of this (private) class is returned from
+        `vaex.dataframe.DataFrame.__dataframe__`.
+        """
+        self._df = df
+        # ``nan_as_null`` is a keyword intended for the consumer to tell the
+        # producer to overwrite null values in the data with ``NaN`` (or ``NaT``).
+        # This currently has no effect; once support for nullable extension
+        # dtypes is added, this value should be propagated to columns.
```
maartenbreddels, line 619: Also, this would cast the column to a float (since that only supports nans)

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+        TODO: details on ``n_chunks``
+        """
+        if n_chunks == None:
+            size = self.num_rows()
+            n_chunks = self.num_chunks()
+            i = self._df.evaluate_iterator(self.get_column(0)._col, chunk_size=size // n_chunks)
+            iterator = []
+            for i1, i2, chunk in i:
+                iterator.append(_VaexDataFrame(self._df[i1:i2]))
+            return iterator
```
maartenbreddels, line 679: What about sth like:
```suggestion
            for i1, i2, chunk in i:
                yield iterator.append(_VaexDataFrame(vaex.from_dict(chunks))
```
This way consumers can consume the chunks, without us putting everything in memory.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+        elif self.num_chunks == 1:
+            size = self.num_rows()
+            i = self._df.evaluate_iterator(self.get_column(0)._col, chunk_size=size // n_chunks)
+            iterator = []
+            for i1, i2, chunk in i:
+                iterator.append(_VaexColumn(self._df[i1:i2]))
+            return iterator
```
maartenbreddels, line 686: Maybe I misunderstood, but I think we can simply do
```suggestion
            yield self
```
Similar to https://github.com/data-apis/dataframe-api/blob/27b8e1cb676bf10704d1dfc3dca0d0d806e2e802/protocol/pandas_implementation.py#L766

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -0,0 +1,688 @@
...
+    DATETIME = 22
+    CATEGORICAL = 23
+
+
+def convert_column_to_ndarray(col: ColumnObject) -> pa.Array:
+    """
+    Convert an int, uint, float or bool column to an arrow array
+    """
+    if col.offset != 0:
+        raise NotImplementedError("column.offset > 0 not handled yet")
```
maartenbreddels, line 109: Before I force pushed, you asked if we used offset. I think we need them for string arrays, and for slicing bitmask arrays, see https://github.com/data-apis/dataframe-api/issues/67

maartenbreddels (changes requested):
Nice work! my biggest comment/change was the change of the describe_null, which can be 3 options (no null, np.ma or arrow). I hope you agree with that.
Would love to see string support!

PS: Please keep the line endings as line feeds.

packages/vaex-core/vaex/dataframe_protocol.py
AlenkaF: Maybe I should restate: the values changed in a strange way when transferring data with `buffer_to_ndarray` in the case of a bool arrow array if `.to_numpy` was used.

But I will try to use previous suggestion  (`.values/.evaluate()`) for arrow and numpy and I will delete the comment if it works.

AlenkaF: I get the same strange change of values for `evaluate()` also.  It happens in the case of arrow arrays (int and bool) with missing values.

```python
df = vaex.from_arrays(
    arrow_int_m = pa.array([0, 1, 2, None, 0], mask=np.array([0, 0, 0, 1, 1], dtype=bool)),
    arrow_float_m = pa.array([0.5, 1.5, 2.5, None, 0.5], mask=np.array([0, 0, 0, 1, 0], dtype=bool)),
    arrow_bool_m = pa.array([True, False, True, None, True], mask=np.array([0, 0, 1, 1, 0], dtype=bool))
)

col = df.__dataframe__().get_column_by_name('arrow_int_m')
b, d = col.get_buffers()["data"]
buffer_to_ndarray(b, d)
```
output:
```
array([                  0, 4607182418800017408, 4611686018427387904,
         -2251799813685248,   -2251799813685248], dtype=int64)
```

I can commit with the error so you can have a look?

packages/vaex-core/vaex/dataframe_protocol.py
AlenkaF: I added this part thinking one could read a dataframe (that is not chunked) in chunks specifying the `n_chunks` (item 12 from [protocol-design-requirements](https://github.com/AlenkaF/dataframe-api/blob/main/protocol/design_requirements.md#protocol-design-requirements)).

But there are some errors left, sorry about that. It should be:

```python
    def get_chunks(self, n_chunks: Optional[int] = None) -> Iterable["_VaexDataFrame"]:
        """
        Return an iterator yielding the chunks.
        TODO: details on ``n_chunks``
        """
        if n_chunks == None:
            size = self.num_rows()
            n_chunks = self.num_chunks()
            i = self._df.evaluate_iterator(self.get_column(0)._col, chunk_size=size // n_chunks)
            iterator = []
            for i1, i2, chunk in i:
                iterator.append(_VaexDataFrame(self._df[i1:i2]))
            return iterator
        elif self.num_chunks() == 1:
            size = self.num_rows()
            i = self._df.evaluate_iterator(self.get_column(0)._col, chunk_size=size // n_chunks)
            iterator = []
            for i1, i2, chunk in i:
                iterator.append(_VaexDataFrame(self._df[i1:i2]))
            return iterator
        else:
            raise ValueError("Dataframe is already chunked.")
```

maartenbreddels:
Awesome to see the strings coming in, getting close!
I pushed some a change in how to consume the buffers for strings, which is more efficient (see vaex/arrow/convert.py for some more code examples), and I think we should also use that method to produce the buffers (so that my new mem copy test does not fail).

rgommers:
This is really great to see - thanks a lot @AlenkaF and @maartenbreddels!

PR #1510, opened 2021-08-12
Additional features for the keras transformer

JovanVeljanoski:
Few additional features for the tensorflow/keras addition to vaex-ml :

- [x] better testing
- [ ] all tests passing
- [x] `df.ml.tensorflow.to_keras_generator(..)` provides a nice way to stream vaex data to a keras model for training
- [ ] better docstrings
- [ ] update of the documentation: ml examples

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -55,7 +55,7 @@ jobs:
       if: matrix.os != 'windows-latest'
       run: |
         conda activate vaex-dev
-        mamba install -c conda-forge py-xgboost
+        mamba install -c conda-forge py-xgboost tensorflow
         # xgboost is not available on windows on conda-forge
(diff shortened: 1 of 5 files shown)


maartenbreddels:
As found in https://stackoverflow.com/questions/64881851/tensorflow-keras-model-load-error-valueerror-the-last-dimension-of-the-inputs we now use a directory to store the model.

maartenbreddels:
`  - package tensorflow-1.13.1-h21ff451_7 requires tensorflow-base 1.13.1 py37_7, but none of the providers can be installed` 😢

maartenbreddels:
we should skip windows https://github.com/conda-forge/tensorflow-feedstock/issues/124

maartenbreddels:
I think we can leave the last two points for a separate docs update.

PR #1516, opened 2021-08-13
🐛 vaex-server was broken (missing import and untracked file)

maartenbreddels:
This does not get triggered in CI, we might want to improve that.


maartenbreddels:
We now also install, and try out the command line, and server

Issue #1495, opened 2021-08-07
[BUG-REPORT] writer_threads is ignored when exporting to hdf5


dmvieira:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Please provide a clear and concise description of the problem. This should contain all the steps needed to reproduce the problem. A minimal code example that exposes the problem is very appreciated.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.3.0
 - Vaex was installed via: pip 
 - OS: ubuntu 20.04

**Additional information**
We're offering this option in API, but writer_threads is not in use

maartenbreddels:
Indeed, we need to pass that argument to `writer.write(..)`, do you want to open a PR for this?

PR #1520, opened 2021-08-14
Adding writer params to hdf5 writer

dmvieira:
Params was lost as described here: https://github.com/vaexio/vaex/issues/1495


maartenbreddels (approved):
Thanks a lot!

PR #1492, opened 2021-08-04
DOCS: fix typo in io example

JovanVeljanoski:
Addresses #1483

PR #1522, opened 2021-08-14
Docs: Keras + Struct docs

JovanVeljanoski:
Small update to the documentation pages:

- Added the `struct` methods to the API documentation
- Added a `Keras` example to the documentation
- Added a simplified example in the docstrings of `ml.tensorflow.KerasModel`
- Fixed typo found by #1492 

To do: probably later on in a separate PR, we should update the main tutorial, or gave an additional page explaining the usage of the arrow `struct` types.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -139,6 +139,13 @@ Timedelta operations
      :members:
      :special-members:
 
+Struct (arrow) operations
+~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


maartenbreddels (changes requested):
Please don't commit such big changes in the notebook 're-run'. Try avoid committing them, otherwise the repo will grow to big. Search for 'Kears' might find some typos :P

PR #1535, opened 2021-08-24
Parameter to decide if drop columns or not while running state_set

fsiola:
When using states to perform prediction on a single datapoint, a lot of time was spent on deleting columns. Allowing it to be optional reduced the prediction time by 10% in my case.


fsiola:
Build failed for some S3 reason. Not sure how to fix it :)

JovanVeljanoski:
Hey! Thanks for this. 
Ignore that S3 thing, it is caused by an external dependency. I will let @maartenbreddels review this when he has a chance.

maartenbreddels:
Yeah, I think this makes sense, thanks!

PR #1537, opened 2021-08-25
Feat(vaex-ml): MultiHotEncoder

JovanVeljanoski:
This PR introduces the `MultiHotEncoder` in vaex-ml.
This encoder (sometimes called Binary encoder) encodes categorical variables according to a multi-hot scheme.

What it does: first it ordinal-encodes a categorical column, and those ordinal encodings are converted to a binary number. 
Each digit of the binary number is a dimension of the multi-hot vector. 

This can be considered an improvement over the One-hot scheme, as it guards against too many dimensions for features with high cardinality, while removing the ordinality of the encoding. 

Checklist:
- [x] `vaex.ml.MultiHotEncoder` implementation
- [x] unit test
- [x] Docs updated (api docs and the ml tutorial)
- [x] Review


JovanVeljanoski:
Failure is unrelated.

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -448,6 +448,91 @@ def transform(self, df):
...
+            self.binary_labels_[feature] = {key: f'{int(bin(value)[2:]):0{self.n_dims_[feature]}d}' for key, value in labels_dict.items()}
+
+    def _get_n_dims(self, n_labels):
+        '''Get the number of dimensions for the multi-hot vector, based on the number of unique labels.'''
+        if np.mod(n_labels, 2) == 0:
+            binary = bin(n_labels)[2:]
+            return len(binary)
+        else:
+            binary = bin(n_labels)[2:]
+            return len(binary) + 1
```
maartenbreddels, line 517:
```suggestion
        '''Get the number of dimensions for the multi-hot vector, based on the number of unique labels.'''
        return int(math.ceil(math.log2(n_labels)))
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -448,6 +448,91 @@ def transform(self, df):
...
+                labels.remove(None)
+                labels.sort()
+                labels.insert(0, None)  # This is done in place
+            else:
+                labels.sort()
+
+            labels_dict = dict(zip(labels, np.arange(1, n_labels+1)))
+            self.labels_[feature] = labels_dict
+            self.n_dims_[feature] = self._get_n_dims(n_labels)
+            self.binary_labels_[feature] = {key: f'{int(bin(value)[2:]):0{self.n_dims_[feature]}d}' for key, value in labels_dict.items()}
```
maartenbreddels, line 508:
```suggestion
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -448,6 +448,91 @@ def transform(self, df):
...
+    1  green            0          1          0
+    2  green            0          1          0
+    3  blue             0          0          1
+    4  red              0          1          1
+    '''
+
+    prefix = traitlets.Unicode(default_value='', help=help_prefix).tag(ui='Text')
+    labels_ = traitlets.Dict(default_value={}, allow_none=True, help='The ordinal-encoded labels of each feature.').tag(output=True)
+    n_dims_ = traitlets.Dict(default_value={}, allow_none=True, help='Number dimensions of the multi-hot vector.').tag(output=True)
+    binary_labels_ = traitlets.Dict(default_value={}, allow_none=True, help='The binary-encoded labels of each feature.').tag(output=True)
```
maartenbreddels, line 487:
```suggestion
```

packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -448,6 +448,91 @@ def transform(self, df):
...
+        :return: A shallow copy of the DataFrame that includes the encodings.
+        :rtype: DataFrame
+        '''
+        copy = df.copy()
+        for feature in self.features:
+            unseen_value = f'{0:0{self.n_dims_[feature]}d}'
+            tmp = copy[feature].map(self.binary_labels_[feature], default_value=unseen_value)
+            for i in range(self.n_dims_[feature]):
+                name = f'{self.prefix}{feature}_{i}'
+                copy[name] = tmp.str.get(i).astype('int8')
```
maartenbreddels, line 532: No need to store things like n_dims etc, this can all be done at runtime, the only thing needed is the labels. And based on that you can test a bit, like this:
```suggestion
            labels_dict = dict(zip(labels, np.arange(1, n_labels+1)))
            tmp = copy[feature].map(labels_dict, default_value=unseen_value)
            for i in range(self.n_dims_[feature]):
                name = f'{self.prefix}{feature}_{i}'
                copy[name] = (tmp >> i) & 1
```
I think sth like this is much simpler and faster (not 100% sure if correct), but we should not go through a int->str->int conversion, this should all be done using integer math.

maartenbreddels (changes requested):
Nice encoder, see my comments on how we can do it faster :)

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Ok I will update this to make it more.. mathematical :)

packages/vaex-ml/vaex/ml/transformations.py
JovanVeljanoski: Ok cool, i will take this approach! It needs some more adjustments to do the ordering right

JovanVeljanoski:
I think I addressed all the points that you raised in the first pass of the review. I things are better this time around!

PR #1541, opened 2021-08-25
🐛 protect file creation parts with lock files

maartenbreddels:
This shows up when multiple threads or processes try to do a conversion
or create a cache file.


maartenbreddels:
@nicolaskruchten this should help with dash applications starting up and creating the cache file multiple time, and in dash apps using `vaex.open` with `convert=True`.

PR #1546, opened 2021-08-27
✨ avoid using nest_asyncio

maartenbreddels:
This avoids the use of nest_asyncio, where we have instead of 'shared-nested' ioloops, we make 'new-nested' ioloops. This means that if we do a async->sync->async call, the deepest async code will pause the ioloop of the parent. Note that this not work with vaex-jupyter, because that requires the jupyter ioloop to continue running (hence the WIP).

This idea was discussed in private with @SylvainCorlay, putting the PR out as a potential fix for https://github.com/spyder-ide/spyder/issues/16183


maartenbreddels:
Instead of making this the default (rendering vaex-jupyter useless), this can be configured using the env var
`VAEX_ASYNC=awaitio python ...` or settings `df.executor.async_method = "awaitio"`

PR #1563, opened 2021-09-02
🐛 fix writing higher d arrays to hdf5

maartenbreddels:


JovanVeljanoski:
This is excellent, thank you so much!

PR #1566, opened 2021-09-10
🐛 too much noise when exceptions occurred in the executor.

maartenbreddels:
If we had N threads, all throwing the same exception, N-1
warnings of un-awaited exceptions. We also always put the
exception in the log.


maartenbreddels:
Failure seems to be around convert=True with multithreaded, I think we may have a bug in there.

Issue #1019, opened 2020-10-22
[Feature request] Insert a column with constant value 


ttpro1995:
I would like a function to add new columns with constant value.
Preferable similar to `pandas.DataFrame.insert`. However, adding a new column is enough. 

![image](https://user-images.githubusercontent.com/7196876/96822228-c717d100-1453-11eb-9a50-d530854a67ee.png)

As my current workaround in Vaex. 

```
[df_vaex = df_vaex.extract() # require if there are filter apply on df_vaex 
df_vaex["e"] = np.repeat("five", df_vaex.shape[0])
```

![image](https://user-images.githubusercontent.com/7196876/96822594-a2702900-1454-11eb-9db9-63e40614caf1.png)

The problem here is that **.extract()** give a lot of memory footprint. It is seem that whole dataframe is save on RAM. 

`df_vaex["e"] = "five"` does not work, show weird error (is it bug ? More detail #1020)

maartenbreddels:
Yeah, I like this idea.

JovanVeljanoski:
Yeah, we've had a discussion on it before: https://github.com/vaexio/vaex/issues/455

There are also the variables, which is perhaps why we haven't thought about a constant column much before. But I agree it would be a nice addition to have.

PR #1570, opened 2021-09-13
feat(core): Method to add virtual columns having constant values.

JovanVeljanoski:
This PR adds a vaex method to add virtual columns having constant values. 
The method is called `vaex.vconstant`
This was requested a few of times in the past, example #1019, #1557

This is how for instance one can create a fully virtual dataframe:
```python

import vaex

N = 100_000_000
df = vaex.from_arrays(x=vaex.vconstant(value=10, length=N),
                      y=vaex.vconstant(value='Vaex', length=N),
                      z=vaex.vrange(0, N)
                     )
```

Checklist:
- [x] Implementation
- [x] Unit-tests
- [x] API docs update
- [ ] Code review

### DIFF
--- a/packages/vaex-core/vaex/column.py
+++ b/packages/vaex-core/vaex/column.py
@@ -79,10 +79,9 @@ class ColumnVirtualConstant(Column):
         return self.length
 
     def __getitem__(self, slice):
-        template = pa.array([self.value] * self.chunk_size, type=self.dtype)
+        template = pa.array([self.value] * self.chunk_size, type=self.dtype.arrow)
         n_chunks = (self.length + self.chunk_size - 1) // self.chunk_size
-        ar = pa.chunked_array([template] * n_chunks)
-        ar = ar.slice(length=self.length)
+        ar = pa.chunked_array([template] * n_chunks)[slice]
         return ar
 
     def _fingerprint(self):
@@ -91,9 +90,9 @@ class ColumnVirtualConstant(Column):
 
     def _get_dtype(self, dtype):
         if dtype is None:
-            return pa.array([self.value]).type
+            return vaex.datatype.DataType(pa.array([self.value]).type)
         else:
-            vaex.datatype.DataType(dtype).arrow
+            return vaex.datatype.DataType(dtype)
 
     def trim(self, i1, i2):
         length = i2 - i1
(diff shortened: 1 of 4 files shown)


packages/vaex-core/vaex/column.py
```diff
@@ -68,6 +68,38 @@ def trim(self, i1, i2):
...
+
+    def _fingerprint(self):
+        fp = vaex.cache.fingerprint(self.value, self.length, self.dtype, self.chunk_size)
+        return f'column-vconstant-{fp}'
+
+    def _get_dtype(self, dtype):
+        if dtype is None:
+            return pa.array([self.value]).type
+        else:
+            vaex.datatype.DataType(dtype).arrow
```
maartenbreddels, line 96: missing return, and the 2nd case wraps it in DataType, while the first does not.

packages/vaex-core/vaex/column.py
```diff
@@ -68,6 +68,38 @@ def trim(self, i1, i2):
...
+        self.chunk_size = chunk_size
+
+    def __len__(self):
+        return self.length
+
+    def __getitem__(self, slice):
+        template = pa.array([self.value] * self.chunk_size, type=self.dtype)
+        n_chunks = (self.length + self.chunk_size - 1) // self.chunk_size
+        ar = pa.chunked_array([template] * n_chunks)
+        ar = ar.slice(length=self.length)
```
maartenbreddels, line 85: this cannot be self.length, it should be related to `slice` (example are probably in the same .py file)

maartenbreddels (changes requested):
Nice! I spotted a few small issues.

packages/vaex-core/vaex/column.py
JovanVeljanoski: Ah indeed, sorry. 

The initial idea was for both cases dtype to be arrow-style, but now I see I can just use the arrow version if it when calling `pa.array`

packages/vaex-core/vaex/column.py
JovanVeljanoski: i adjusted things, hope it is better now.

maartenbreddels:
Love it, thanks!

Issue #1582, opened 2021-09-22
[BUG-REPORT] - Using virtual column after conversion to datetime64[ns] returns KeyError


harshasridhar:
**Description**
Converting the column datatype from string to 'datetime64[ns]' and accessing the date functions returns a KeyError.
The attached code snippet shows the conversion and the usage.

**Software information**
 - Vaex version :
 ```python
{
'vaex': '4.5.0',
 'vaex-core': '4.5.1',
 'vaex-viz': '0.5.0',
 'vaex-hdf5': '0.10.0',
 'vaex-server': '0.6.1',
 'vaex-astro': '0.9.0',
 'vaex-jupyter': '0.6.0',
 'vaex-ml': '0.14.0'
}
```
 - Vaex was installed via: pip on Google Colab
 - OS: Linux

**Additional information**
Using virtual column after conversion to datetime returns KeyError
```python
import vaex
import matplotlib.pyplot as plt
from matplotlib.colors import LogNorm
data=vaex.open('2017_yellow_combined.csv.hdf5')
data['hour_of_day'] = data.tpep_pickup_datetime.astype('datetime64[ns]').dt.hour
->plt.hist2d(data.hour_of_day.values,data.DOLocationID.values,norm=LogNorm())
```
```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/usr/local/lib/python3.7/dist-packages/vaex/scopes.py in evaluate(self, expression, out)
    105             # logger.debug("try avoid evaluating: %s", expression)
--> 106             result = self[expression]
    107         except KeyError:

27 frames
KeyError: 'Unknown variables or column: "dt_hour(astype(tpep_pickup_datetime, \'datetime64[ns]\'))"'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<string> in <module>

AttributeError: 'pyarrow.lib.TimestampArray' object has no attribute 'flags'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
KeyError: 'Unknown variables or column: "dt_hour(astype(tpep_pickup_datetime, \'datetime64[ns]\'))"'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
<string> in <module>

/usr/local/lib/python3.7/dist-packages/vaex/functions.py in _pandas_dt_fix(x)
    275     # see https://github.com/pandas-dev/pandas/issues/23276
[...]

PR #1583, opened 2021-09-22
Fix: datetime operations on virtual columns created via astype

JovanVeljanoski:
Exposed in #1582 

- [x] Unit-test
- [x] Make unit-test pass
- [ ] Review

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -274,6 +274,8 @@ def notna(x):
 def _pandas_dt_fix(x):
     # see https://github.com/pandas-dev/pandas/issues/23276
     # not sure which version this is fixed in
+    if isinstance(x, pa.lib.TimestampArray):
+        return x.to_pandas()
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I don't get how this tests the issue in #1582, since we do not (or should not) convert to an arrow array. Did you test if the test fails without your code?

JovanVeljanoski:
Yes of course it fails without the fix :) Hence the bug report of #1582 . You can try it!

So what happens is you start from a string, and that is handled by arrow. You cast that to datetime, but the string is then seen as an arrow Timestamp, and that breaks the `dt` stuff. So I cast it to pandas to make it work. 

Perhaps there is a better way of solving it, or maybe something with the `astype` is out of order, but I didn't want to touch that..

maartenbreddels:
Yeah, I checked. I think it's becase we use pa.array(..) as default

maartenbreddels:
Thanks !

PR #1586, opened 2021-09-24
Update installing.rst

yohplala:
@maartenbreddels 
As you propose in #1582, I have updated the install doc with following changes:
- removal of `vaex-arrow`
- addition of `vaex-ml`
- I saw in `meta.yaml` that `vaex-ui` is not mentionned: I took the initiative to remove it, and the sub-sequent paragraph that was related to it. Please, beware of this and confirm if it should be removed indeed, or if it should be kept.

Thanks for your feedbacks,
Bests

### DIFF
--- a/docs/source/installing.rst
+++ b/docs/source/installing.rst
@@ -30,14 +30,9 @@ Longer version
 
 If you don't want all packages installed, do not install the vaex package. The vaex package is a meta packages that depends on all other vaex packages so it will instal them all, but if you don't need astronomy related parts (``vaex-astro``), or don't care about graphql (``vaex-graphql``), you can leave out those packages. Copy paste the following lines and remove what you do not need:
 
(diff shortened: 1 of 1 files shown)


yohplala:
ah, I see maybe something else to change, please confirm. 
In `Longer version`paragraph, it is mentionned something about `vaex-graphql`.
But I don't see this package either in `meta.yaml`. Is it deprecated as well?
Should I remove this part as well from install doc?

" or don’t care about graphql (`vaex-graphql`)"

Thanks for your feedbacks,
Bests

maartenbreddels:
I think this is fine, thanks!

yohplala:
> I think this is fine, thanks!

@maartenbreddels 
Sorry, please what is `vaex-graphql` (my previous comment? Does it still exist?

maartenbreddels:
It exposes a data frame via a graphql interface, via http most of the time, the test server is down now, ask me again next week! :)

Issue #1545, opened 2021-08-26
[BUG-REPORT] Inconsistent behavior of NaN's and dropna() after chunked export to HDF5


chriguBERTO:
### Description
After loading a dataframe from a .csv file, chunking it, exporting it to hdf5 and reloading it, some of the missing values are changed into string `'nan'`. This has two major consequences:

1) `vaex.dataframe.DataFrame.dropna()` does not consider them
2) After mapping the `'nan'` onto `None` the function `vaex.dataframe.DataFrame.dropna()` only considers them if the parameter `column_names` is specified

Please note that there is the bug report [#1413](https://github.com/vaexio/vaex/issues/1413) that reported a NaN issue together with an hdf5 export. This issue is addressed in [#1418](https://github.com/vaexio/vaex/pull/1418).

#### Loading a .csv

Having a file `test.csv` 

```
x|y
0|2188
1|
2|BA5
3|
```

and running

```python
import vaex
vaex.from_csv('test.csv', sep='|', encoding='ANSI')
```

results in 

```python
  #    x  y
  0    0  2188
  1    1  --
  2    2  BA5
  3    3  --
```

Dropping the rows that contain at least one empty field works fine, so that 

```
df = vaex.from_csv('test.csv', sep='|', encoding='ANSI')
df.dropna()
```

results in

```python
  #    x  y
  0    0  2188
  1    2  BA5
```

#### Chunking, exporting and loading

After chunking, exporting to hdf5 an re-opening the files

```python
from pathlib import Path

chunks = vaex.from_csv('test.csv', chunk_size=2, sep='|', low_memory=False, encoding='ANSI')
for i, chunk in enumerate(chunks):
    filename = f'test_{i:02}.hdf5'
    chunk.export(filename)
   
hdf5_files = list(Path.cwd().glob('test_*.hdf5'))
df = vaex.open(hdf5_files)
```

we have a dataframe `df` such as

```python
  #    x  y
  0    0  2188
  1    1  nan
  2    2  BA5
  3    3  --
```

while `df.dropna()` finds the row 3 to be removed but misses the row 1

```python
  #    x  y
  0    0  2188
  1    1  nan
  2    2  BA5
```

### Preprocessing

Now we can choose to 'preprocess' the `'nan'` into `None` by

```python
df['y'] = df['y'].map(mapper={'2188': '2188', 'nan': None, 'BA5': 'BA5'})
```

what results in a `df` such as

```python
  #    x  y
  0    0  2188
[...]

PR #1620, opened 2021-10-07
Added test for dropna to consider hidden original columns

chriguBERTO:
The Bug-Report [1545](https://github.com/vaexio/vaex/issues/1545) described a broader issue that mainly broke down to the fact that `vaex.dataframe.DataFrame.dropna()` did not consider hidden original columns when the parameter `column_names` is unspecified. This is an attempt to add a corresponding test and fix the bug.

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4735,7 +4735,7 @@ class DataFrame(object):
     def dropmissing(self, column_names=None):
         """Create a shallow copy of a DataFrame, with filtering set using ismissing.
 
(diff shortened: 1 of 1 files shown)


JovanVeljanoski:
@maartenbreddels (if understand this correctly..)

What would happen in this case.
Say you have a original column without N/A values. You do some computation using that column, and overwrite it so you have a hidden column. 
Now you take that output and it is a basis for yet another computation, but for some reason (say division by zero) the output of this 2nd computation may have some N/As. And the chain can continue (and such chains are common in ML pipelines for example during preprocessing + feature engineering)..

Issue #1624, opened 2021-10-11
List of Numpy functions that can be used in expressions


dnessett:
I am trying to create a virtual column that is the one shifted value of another column. I know there is a shift() method on the dataframe class, but it appears that creates a whole new copy of the column. I tried to create an expression using a Numpy function, i.e., np.roll, that could then be used to form a virtual column, but it appears it isn't compatible with vaex expressions. Specifically, I did the following:

import vaex
osc_data = vaex.open('.../data.hdf5') # '...' is the path to data.hdf5
import numpy as np
osc_data['sx'] = np.roll(osc_data.x, -1)

This threw an error:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/var/folders/l5/wjz73b69427bh_9x0954j0dr0000gn/T/ipykernel_6758/539048344.py in <module>
      1 import numpy as np
----> 2 osc_data['sx'] = np.roll(osc_data.x, -1)
      3 # print(type(osc_data['sx']))
      4 # print(type(osc_data['x']))
      5 # print(osc_data[['x', 'sx']])

/usr/local/lib/python3.9/site-packages/vaex/dataframe.py in __setitem__(self, name, value)
   4969         if isinstance(name, six.string_types):
   4970             if isinstance(value, supported_column_types):
-> 4971                 self.add_column(name, value)
   4972             else:
   4973                 self.add_virtual_column(name, value)

/usr/local/lib/python3.9/site-packages/vaex/dataframe.py in add_column(self, name, f_or_array, dtype)
   3261                 self._length_original = _len(data)
   3262                 self._index_end = self._length_unfiltered
-> 3263             if _len(ar) != self.length_original():
   3264                 if self.filtered:
   3265                     # give a better warning to avoid confusion

/usr/local/lib/python3.9/site-packages/vaex/dataframe.py in _len(o)
     65 
     66 def _len(o):
---> 67     return o.__len__()
     68 
     69 

TypeError: len() of unsized object

My guess is np.roll is not one of the Numpy functions supported in vaex expressions. However, I could find not list of Numpy functions that are valid in vaex expressions. This means those of us coding with vaex have to use hit or miss guessing which functions are compatible and which are not. This is extremely frustrating.

As a side question, is there any other way of creating an expression that shifts a column without copying the column's data?

JovanVeljanoski:
@yohplala is correct on all accounts.

in Vaex `shift`, `diff` all make virtual columns. 

[...]

PR #1636, opened 2021-10-14
docs: improve the docstrings a bit

JovanVeljanoski:
the docstrings of diff and shift where not showing correctly.


JovanVeljanoski:
This issue was as noticed in #1624

Issue #1592, opened 2021-09-25
[BUG-REPORT] Accessing column values gives: AttributeError: 'pyarrow.lib.ChunkedArray' object has no attribute 'dtype'


abf7d:
**Description**
I have a 14 million row dataset in which I loaded from a feather file. I'm trying to detect which points are in a polygon, but am running into an error. Here is sample code which I used:
```
 px = np.array([1.5, 2.5, 2.5, 1.5])
 py = np.array([2.5, 2.5, 3.5, 3.5])
 df = vaex.from_arrays(x=[1, 2, 3], y=[2, 3, 4])
 df['inside'] = df.geo.inside_polygon(df['x'], df['y'], px, py)
 print(str(df))
```
This code is successful and prints out the dataframe with the `inside` virtual column. However, when I try to initialize the dataframe by loading my file instead using from_arrays, I get the error ` AttributeError: 'pyarrow.lib.ChunkedArray' object has no attribute 'dtype'`. Here is my code
```
 df = vaex.open(path)
 df['mask'] = df.geo.inside_polygon(df['filecol1'], df['filecol2'], px, py)
 print(str(df))
```
The last line crashes (I'm guessing because it is trying to load the values from the virtual column into memory). I do not know why this is happening. I am able to bin the data from the file with the following code:
```
hist = df.count(
    binby=axes,
    limits=limits,
    shape=(binnum,)*len(axes) 
)
```
This this line successfully gives binned data. Even just trying to access the values from the virtual column throws the same error: 
```
 z = df['inside'].values
```
How can I access which points are within the given polygon from my file loaded data?

**Software information**
 - Vaex version 4.4.0
 - Vaex was installed via: pip
 - OS: Windows

**Additional information**
Please state any supplementary information or provide additional context for the problem (e.g. screenshots, data, etc..).

Nicholas-Schaub:
I also ran into this issue. I did a little debugging and discovered that there appears to be an issue with how the expressions are being processed in the `geo` module.

I added some debug logs to track down what the issue is, and it looks like it is trying to search for column names in an empty dataframe. It seems like the original dataframe is maybe getting letting somewhere. The first error that get's thrown is a key error, and it cannot find the expression in the dataframe. However, the debug logs actually show this error occurs many times, even for the original columns that were passed in as `x` and `y`.

Here is the full traceback I was receiving when trying to reproduce:

```
Traceback (most recent call last):
[...]

PR #1644, opened 2021-10-14
test(geo): inside_polygon fails when loading from file

abf7d:
Created a test for inside_polygon which uses df_factory insead of vaex.from_arrays to simulate loading data from a file. Currently this test fails due to a bug when trying to access dataframe values. 

https://github.com/vaexio/vaex/issues/1592


abf7d:
Thanks for the tip. I removed the new unit test and replaced all `vaex.from_arrays` with `df_factory`.

maartenbreddels:
Thanks @abf7d !

PR #1671, opened 2021-10-28
Better dot_product

JovanVeljanoski:
This PR improves the dot_product. Before there was a limitation of the size of the vectors one can use to calculate a dot product. Now that limitation is lifted. So things like PCA, and Incremental PCA can be used with hundreds of dimensions. 

The dot product is made available more generally in `df.func.dot_product`.

### DIFF
--- a/packages/vaex-ml/vaex/ml/transformations.py
+++ b/packages/vaex-ml/vaex/ml/transformations.py
@@ -19,11 +19,6 @@ help_features = 'List of features to transform.'
 help_prefix = 'Prefix for the names of the transformed features.'
 
 
-def dot_product(a, b):
-    products = ['%s * %s' % (ai, bi) for ai, bi in zip(a, b)]
-    return ' + '.join(products)
-
-
 @register
 class StateTransfer(HasState):
     state = traitlets.Dict()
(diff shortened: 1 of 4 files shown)


packages/vaex-ml/vaex/ml/transformations.py
```diff
@@ -127,20 +122,14 @@ def transform(self, df, n_components=None):
         for i in range(n_components):
             vector = eigen_vectors[:, i]
             if self.whiten:
```
Ben-Epstein, line 124: a slightly cleaner way to do this could be 
```
expr = copy.func.dot_product(expressions, vector)
if self.whiten:
    expr = f'({expr}) / {np.sqrt(self.explained_variance_[i])}'
```

maartenbreddels: I agree, do you want to change that Jovan, ok are you ok with this?

JovanVeljanoski: Yeah I agree

JovanVeljanoski:
The failure is due to updates to `aibotocore`. More info can be found here: https://github.com/dask/dask/pull/8336

PR #1676, opened 2021-11-03
My Binder link hopefully works now

dasdristanta13:
Please check this out. Hopefully, this is working now.
It's better to put the link in the readme. Because when in the Github environment the notebooks don't allow the newer pop-up windows to open the binder environment. But when we put it into readme it opens up there without any problem.


maartenbreddels:
Hi,

thanks for you PR. I was wondering why you closed this. I'm a bit unsure though us providing my binder, since we are kind of compute heavy. What do you think @JovanVeljanoski ?

cheers,

Maarten

JovanVeljanoski:
I am fine with staying with mybinder if there are no technical or ethical problems (since as @maartenbreddels says we are compute heavy). Longer term I was thinking of moving to google colab or even kaggle. Will give people more room to experiment. But I am happy either way for now.

PR #1680, opened 2021-11-04
test: enable test disabled in #1594

maartenbreddels:


maartenbreddels:
Will need to look into appveyor later on

Issue #1697, opened 2021-11-12
[FEATURE-REQUEST] Could we have `str.rsplit()`?


yohplala:
**Description**
Hi,
`str.split()` is supported by vaex, but trying to use `str.rsplit()` returns an error.

```python
import vaex as vx
li = ['ah.1.arrow', 'ah.199.arrow', 'ah.79.arrow']
vdf = vx.from_arrays(fname = li)
vdf['ts'] = vdf['fname'].str.rsplit('.',1).apply(lambda x: x[-2])
```
Error is triggered when trying to display `vdf`.
```python
vdf
ERROR:MainThread:vaex:error evaluating: ts at rows 0-3
KeyError: 'Unknown variables or column: "lambda_function(str_rsplit(fname, \'.\', 1))"'
Out[54]: 
  #  fname         ts
  0  ah.1.arrow    error
  1  ah.199.arrow  error
  2  ah.79.arrow   error
```
But with `split()`, it works.

**Additional context**
Strings I want to process are file path (basepath + basename), but information that is of interest is at the end. Using `rsplit('.',1)` ought to be faster than splitting the full string with `split()`.

JovanVeljanoski:
This is a bug in arrow i think - which was fixed. Please update your `pyarrow`. I just tried with pyarrow==6.0.0 and rsplit works fine.

yohplala:
> This is a bug in arrow i think - which was fixed. Please update your `pyarrow`. I just tried with pyarrow==6.0.0 and rsplit works fine.

Hi,
May it be that you also solved something into last vaex version?
I upgraded to pyarrow 6.0.0 and I still have an error message (but not the same).

Here are the versions I am using for vaex:
{'vaex': '4.5.0',
 'vaex-core': '4.5.1',
 'vaex-viz': '0.5.0',
 'vaex-hdf5': '0.10.0',
 'vaex-server': '0.6.1',
 'vaex-astro': '0.9.0',
 'vaex-jupyter': '0.6.0',
 'vaex-ml': '0.14.0'}

Here is the error message.
```python
import vaex as vx
li = ['ah.1.arrow', 'ah.199.arrow', 'ah.79.arrow']
vdf = vx.from_arrays(fname = li)
vdf['fname'].str.rsplit('.',2)

Traceback (most recent call last):

  File "/home/yoh/anaconda3/lib/python3.8/site-packages/vaex/scopes.py", line 106, in evaluate
    result = self[expression]

[...]

PR #1700, opened 2021-11-13
fix(core): fix str.rsplit() when an explicit split pattern is specified

JovanVeljanoski:
Addresses the issue raised in #1697

Checklist:
- [x] Make a test
- [x] Make the test pass


maartenbreddels (approved):
Nice work!

Issue #1631, opened 2021-10-12
[BUG-REPORT] sort by 2 columns with different ordering yields unexpected behavior


Ben-Epstein:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Vaex dataframes can currently be sorted by 2 columns, but when specifying a value for `ascending` if you pass in a list of booleans (like in Pandas), it will `bool` the entire expression and return all columns ordered by that single bool.

The expected behavior should either be to follow the users specification, or throw an error saying that this functionality isn't currently supported.

Example:
```
import vaex
from random import random
import numpy as np
n = 100
col1 = np.random.randint(0, 20, size=n)
col2 = np.random.randint(0, 20, size=n)
arrays = {
    'col1': col1,
    'col2': col2
}
M = vaex.from_arrays(**arrays)
M.sort(by=['col1','col2'], ascending=[True, False])
```

For some more context, see this result:
![image](https://user-images.githubusercontent.com/22605641/136994979-ba2c55cf-5d85-41a7-91c0-460c5dd67f5d.png)
When 2 rows have the same value for `col1` you should expect col2 to be sorted descending, but it's not. 

By comparison, the same code done with pandas 
```
M.to_pandas_df().sort_values(by=['col1','col2'], ascending=[True, False])
```
yields
![image](https://user-images.githubusercontent.com/22605641/136995215-bf06f818-9daa-4055-bce0-674a10e5295d.png)

Not supporting this yet seems reasonable to me, but in that case the function should fail given the value passed to `ascending`

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`:
```
{'vaex': '4.5.0',
 'vaex-core': '4.5.1',
 'vaex-viz': '0.5.0',
 'vaex-hdf5': '0.10.0',
 'vaex-server': '0.6.1',
 'vaex-astro': '0.9.0',
 'vaex-jupyter': '0.6.0',
 'vaex-ml': '0.14.0'}
```
 - Vaex was installed via: pip / conda-forge / from source : pip
 - OS: macOS big sur

**Additional information**
Please state any supplementary information or provide additional context for the problem (e.g. screenshots, data, etc..).

JovanVeljanoski:
The behavior that you describe is simply not supported. I think the docstrings make that clear:

[...]

Issue #815, opened 2020-06-08
Sort by multiple columns of different types


ermalguni:
Is it possible to sort by multiple columns where the column data types are ints and strings?

JovanVeljanoski:
Yes. did you try it?

ermalguni:
Yes, and I got this: `TypeError: '<' not supported between instances of 'str' and 'float'`.

JovanVeljanoski:
You asked for ints right:) The error suggests that you are attempting to sort floats and strings. 

Joking aside, the sort of vaex should follow numpy convention. I just tried it out with ints and strings, and it works. I actually don't know why it would.. i was thought it would not. Maybe @maartenbreddels would have a better insight into this. 

In general, when talking about sorting strings, you generally refer to sorting them alphabetically right?

ermalguni:
Yes, I asked about ints because that is what the data types are, but the exception message talks about float32. That got me confused as to where did the float32 come from.

Regarding sorting string, yes I mean alphabetical sorting. My data contains geographical data.

maartenbreddels:
Could you give us a piece of code to reproduce the issue? (e.g. using vaex.from_arrays(...)`

abaradia1978:
<img width="537" alt="vaex_sorting_str_error" src="https://user-images.githubusercontent.com/33593180/87909646-42996900-ca86-11ea-8e73-a0c4a8e3a2b2.png">

Attached screenshot shows the error that appears while sorting string fields - the same works find with pandas. let me know if you need any clarification

abaradia1978:
<img width="346" alt="vaex_sorting_ascdes_error" src="https://user-images.githubusercontent.com/33593180/87910051-fd296b80-ca86-11ea-888b-18f208b9d1d4.png">

Also, while sorting multiple fields, if we give different sort order, it does not work - refer above screen shot - it works as expected in pandas

JovanVeljanoski:
Hi @abaradia1978 

Specifying distinct ordering per sorting key is not yet supported. We will discuss this if we can implement it efficiently. 
Thanks for bringing this to our attentino.

abaradia1978:
Thanks @JovanVeljanoski - we currently stuck because of this issue in one of our projects. Look forward to having it resolved. Also will you consider allowing sort on string fields as per my first screen shot above?

JovanVeljanoski:
Hi @abaradia1978  et al, 

[...]

PR #1703, opened 2021-11-14
Explicit `ascending` check in `sort` function

Ben-Epstein:
Explicitly only allow single boolean value sorting in `sort` function.

In numpy you can pass in a list to sort and the sort function will sort by each column independently based on the list. Since vaex does not support this, it should be explicit, to avoid unexpected results by the user. 

https://github.com/vaexio/vaex/issues/1631
https://github.com/vaexio/vaex/issues/815


maartenbreddels:
Yeah, good idea, thanks Ben!

PR #1721, opened 2021-11-19
Docs: add & improve docstrings

JovanVeljanoski:
Small PR to add and improve docstrings
- add docstrings for the `get_group` method of a grouped dataframe
- small modification of the docstrings in the `df.sort` method.


maartenbreddels:
Thanks Jovan!

PR #1735, opened 2021-11-23
Docs: add FAQ item

JovanVeljanoski:
Adding a new FAQ item.


docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+   "source": [
+    "#### What is the optimal file format to use with `vaex`?\n",
+    "\n",
+    "What is \"optimal\" may dependant on what one is trying to achieve. A quick summary would be:\n",
+    "\n",
+    "- For performance: [HDF5](https://en.wikipedia.org/wiki/Hierarchical_Data_Format);\n",
+    "- For interoperability: [Apache Arrow](https://arrow.apache.org/);\n",
+    "- For optimizing disk space: [Apache Parquet](https://parquet.apache.org/).\n",
+    "\n",
+    "`vaex` shines when the data is in a memory-mappable file format, namely HDF5, Apache Arrow, or [FITS](https://en.wikipedia.org/wiki/FITS). We say a file is memory mappable if it has the same structure in memory, as it has on disk. \n",
```
maartenbreddels, line 81:
```suggestion
    "`vaex` shines when the data is in a memory-mappable file format, namely HDF5, Apache Arrow, or [FITS](https://en.wikipedia.org/wiki/FITS). We say a file can be memory mapped if it has the same structure in memory, as it has on disk. Although any file can be memory mapped, if it requires deserialisation there is no advantage to memory mapping.\n",
```

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+    "\n",
+    "In principle, HDF5 and Arrow should give the same performance. For files that would fit into memry the performance between the two is the same. For single files that are larger available RAM, our tests show that HDF5 gives faster performance. What \"faster\" means will likely depend on your system, quantity and type of data.\n",
```
maartenbreddels, line 83:
```suggestion
    "In principle, HDF5 and Arrow should give the same performance. For files that would fit into memory the performance between the two is the same. For single files that are larger available RAM, our tests show that HDF5 gives faster performance. What \"faster\" means will likely depend on your system, quantity and type of data. Performance difference may be caused by converting bit masks to byte masks or by flattening chunked arrow arrays, this performance penalty will likely disappear in the future.\n",
```

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+    "\n",
+    "If your data is spread amongst multiple files that are concatenated on the fly, the performance between HDF5 and Arrow is expected to be the same. Our test show better performance when all of the data is contained in a single file, compared to multipe file.\n",
+    "\n",
+    "The Arrow file format allows seemless interoperability with other eco-systems. If your use-case requires sharing data with other eco-systems, e.g. Java, the Arrow file format is the way to.\n",
+    "\n",
+    "`vaex` also supports Parquet. Parquet is compressed, and thus not memory mappable. There is always a performance penalty when using Parquet, since the data needs to be decompressed before it is used. Parquet however allows lazy reading of the data, which can be decompressed on the fly. Thus `vaex` canvan easily work with Parquet files that are larger than RAM. We recommend using Parquet when one wants to save disk space. It can be also convenient when reading from i/o sources, like Cloud storage for example.\n",
```
maartenbreddels, line 89:
```suggestion
    "`vaex` also supports Parquet. Parquet is compressed, and memory mappable brings no advantage. There is always a performance penalty when using Parquet, since the data needs to be decompressed before it is used. Parquet however allows lazy reading of the data, which can be decompressed on the fly. Thus `vaex` canvan easily work with Parquet files that are larger than RAM. We recommend using Parquet when one wants to save disk space. It can be also convenient when reading from slow i/o sources, like spinner harddrives or Cloud storage for example. Note that using [df.materialize](api.html#vaex.dataframe.DataFrame.materialize) one can get the same performance as hdf5 or arrow files at the cost of memory (or disk space)\n",
```

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
     "```"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### What is the optimal file format to use with `vaex`?\n",
+    "\n",
+    "What is \"optimal\" may dependant on what one is trying to achieve. A quick summary would be:\n",
```
maartenbreddels, line 75:
```suggestion
    "What is \"optimal\" may dependent on what one is trying to achieve. A quick summary would be:\n",
```

maartenbreddels (commented):
Let me know if you agree with my suggestions

maartenbreddels (approved):

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### What is the optimal file format to use with `vaex`?\n",
+    "\n",
+    "What is \"optimal\" may dependent on what one is trying to achieve. A quick summary would be:\n",
+    "\n",
+    "- For performance: [HDF5](https://en.wikipedia.org/wiki/Hierarchical_Data_Format);\n",
+    "- For interoperability: [Apache Arrow](https://arrow.apache.org/);\n",
+    "- For optimizing disk space: [Apache Parquet](https://parquet.apache.org/).\n",
```
mariobuikhuizen, line 79: Maybe also network access?

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+    "\n",
+    "`vaex` shines when the data is in a memory-mappable file format, namely HDF5, Apache Arrow, or [FITS](https://en.wikipedia.org/wiki/FITS). We say a file can be memory mapped if it has the same structure in memory, as it has on disk. Although any file can be memory mapped, if it requires deserialisation there is no advantage to memory mapping.\n",
+    "\n",
+    "In principle, HDF5 and Arrow should give the same performance. For files that would fit into memory the performance between the two is the same. For single files that are larger than available RAM, our tests show that HDF5 gives faster performance. What \"faster\" means will likely depend on your system, quantity and type of data. This performance difference may be caused by converting bit masks to byte masks, or by flattening chunked Arrow arrays. We expect that this performance difference will disappear in the future.\n",
+    "\n",
+    "If your data is spread amongst multiple files that are concatenated on the fly, the performance between HDF5 and Arrow is expected to be the same. Our test show better performance when all the data is contained in a single file, compared to multipe file.\n",
```
mariobuikhuizen, line 85: multipe -> multiple

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+    "\n",
+    "The Arrow file format allows seamless interoperability with other ecosystems. If your use-case requires sharing data with other eco-systems, e.g. Java, the Arrow file format is the way to.\n",
```
mariobuikhuizen, line 87: eco-systems -> ecosystems

docs/source/faq.ipynb
```diff
@@ -66,6 +66,31 @@
...
+    "\n",
+    "`vaex` also supports Parquet. Parquet is compressed, therefore memory mapping brings no advantage. There is always a performance penalty when using Parquet, since the data needs to be decompressed before it is used. Parquet however allows lazy reading of the data, which can be decompressed on the fly. Thus `vaex` canvan easily work with Parquet files that are larger than RAM. We recommend using Parquet when one wants to save disk space. It can be also convenient when reading from slow i/o sources, like spinner hard-drives or Cloud storage for example. Note that by using [df.materialize](api.html#vaex.dataframe.DataFrame.materialize) one can get the same performance as HDF5 or Arrow files at the cost of memory or disk space.\n",
```
mariobuikhuizen, line 89: canvan -> can
spinner  -> spinning

mariobuikhuizen (commented):
Good description 👍, some small issues:

mariobuikhuizen (approved):

JovanVeljanoski:
Thank you both!

PR #1738, opened 2021-11-24
✨ improved progress support

maartenbreddels:
* progress is now an argument to vaex.open to be used in combination with convert
 * more titles for progress values
 * rich progress support


.github/workflows/pythonpackage.yml
```diff
@@ -205,6 +205,12 @@ jobs:
       run: |
         python -c "import vaex; df = vaex.example()"
 
+    - name: Test comand line
+      run: |
+        vaex convert ~/.vaex/data/helmi-dezeeuw-2000-FeH-v2-10percent.hdf5 test.parquet
+        pip install rich
```
JovanVeljanoski, line 211: Should this be added as a proper dependency now? So we can track in case something changes with versions etc.?

JovanVeljanoski (commented):
Wow this is amazing! Very exciting stuff!

.github/workflows/pythonpackage.yml
maartenbreddels: Not for now, the import is also done non-top level, if it mature, and if we are going to use rich at more places, I'm happy to make it a dependency.

maartenbreddels:
screencaps
vaex-progress-unique-jupyter

vaex-progress-export

PR #1741, opened 2021-11-25
Allow casting integers to timedelta64 type

JovanVeljanoski:
Assuming `x` is integer, this now will work: `df.x.astype('timedelta64[s]')`

Requested by: https://github.com/vaexio/vaex/discussions/1740

Checklist:
- [x] Apply fix
- [x] Unit test
- [ ] Review

### DIFF
--- a/tests/astype_test.py
+++ b/tests/astype_test.py
@@ -45,3 +45,11 @@ def test_astype_empty(df_factory):
     df = df_factory(x=[1, 2, 3])
     df = df[df.x<0]
     assert len(df.x.as_numpy().values) == 0
+
+
+def test_astype_timedelta(df_factory):
+    x = [23423, -242, 34656]
+    x_result = np.array([23423,  -242, 34656], dtype='timedelta64[s]')
+    df = df_factory(x=x)
+    df['x_expected'] = df.x.astype('timedelta64[s]')
+    assert x_result.tolist() == df.x_expected.tolist()
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -2512,13 +2512,25 @@ def _astype(x, dtype):
             # arrow does not support timestamp to int/float, so we use numpy
             y = x.to_numpy().astype(dtype)
             return y
+        if pa.types.is_duration(x.type):
+            # for consistency with timedelta, use numpy for this for now
+            y = x.to_numpy().astype(dtype)
+            return y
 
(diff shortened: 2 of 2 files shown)


tests/astype_test.py
```diff
@@ -45,3 +45,11 @@ def test_astype_empty(df_factory):
     df = df_factory(x=[1, 2, 3])
     df = df[df.x<0]
     assert len(df.x.as_numpy().values) == 0
+
+
+def test_astype_timedelta(df_factory):
+    x = [23423, -242, 34656]
+    x_result = np.array([23423,  -242, 34656], dtype='timedelta64[s]')
+    df = df_factory(x=x)
+    df['x_expected'] = df.x.astype('timedelta64[s]')
```
maartenbreddels, line 54: this does not test what you want I think, because x_result already has the dtype you want.

maartenbreddels (commented):
Code looks good, but are you sure the test failed?

tests/astype_test.py
```diff
@@ -45,3 +45,11 @@ def test_astype_empty(df_factory):
     df = df_factory(x=[1, 2, 3])
     df = df[df.x<0]
     assert len(df.x.as_numpy().values) == 0
+
+
+def test_astype_timedelta(df_factory):
+    x = [23423, -242, 34656]
+    x_result = np.array([23423,  -242, 34656], dtype='timedelta64[s]')
```
maartenbreddels, line 52:
```suggestion
    x_result = np.array(x, dtype='timedelta64[s]')
```

tests/astype_test.py
maartenbreddels: ok, I was confused, if you reuse x I think it is more clear.

JovanVeljanoski: It works, but the variable names are the wrong way around :S. I changed them to be what one would expect.

DanielvdC:
Any updates? I am quite dependent of this commit

JovanVeljanoski:
Ping @maartenbreddels

maartenbreddels (approved):
Great, thanks Jovan!

DanielvdC:
I am checking if I can use the function to create a column with datetime, like this:

`df['date'] = df.date_old.astype('datetime64') + df.time.astype('timedelta64[s]')`

where df.date_old is of dtype date32[day] and df.time is of dtype string.

But this generates the following error:

`ValueError: No type alias for timedelta64[s]`

Any thoughts?

PR #1744, opened 2021-11-26
✨ delete hdf5 file when export/convert fails

maartenbreddels:
Otherwise we leave an inconsistent file that may lead other
processes to think a file was property generated.

This can happen when someone ctrl-c.


maartenbreddels:
This will be a breaking change between Vaex-core and vaex-hdf5

PR #1753, opened 2021-12-01
Fix: Evaluate with selections

JovanVeljanoski:
Currently `df.x.evaluate(selection='...')` fails the underlying data of `x` is in arrow. 

This PR fixes that and adds a more stable test.

- [x] Expose issue via a test
- [ ] Make test pass


maartenbreddels (approved):
Nice, thanks!

xdssio:
I got that too and worked around it.

PR #1757, opened 2021-12-03
fix: histogram method on expression to propagate kwargs

JovanVeljanoski:
`df.x.viz.histogram(..)` was not propagating the `**kwargs` to the "parent" `df.viz.histogram(..)` method. Now it does.


maartenbreddels (approved):
Thanks!

PR #1766, opened 2021-12-13
Select by dtypes

xdssio:
Added functionality for selecting columns by dtypes:
```
df.get_columns_by_dtypes(float) -> [columns of types float, float32, float64]
df.get_columns_by_dtypes(np.float32) -> [columns of float32]
df.get_columns_by_dtypes([int, str]) -> [columns of types int and str]
```

Also added a getitem version:
```
df[str] == df[df.get_columns_by_dtypes(str)]
df[int, str] == df[df.get_columns_by_dtypes([str, int])]
```


packages/vaex-core/vaex/dataframe.py
```diff
@@ -4132,14 +4132,22 @@ def get_column_names(self, virtual=True, strings=True, hidden=False, regex=None)
...
         :param virtual: If False, skip virtual columns
         :param hidden: If False, skip hidden columns
         :param strings: If False, skip string columns
         :param regex: Only return column names matching the (optional) regular expression
+        :param dtypes: A list of dtypes to consider
         :param alias: Return the alias (True) or internal name (False).
         :rtype: list of str
         """
+        if isinstance(dtypes, type):
+            dtypes = [dtypes]
```
maartenbreddels, line 4149: In almost all places in vaex we use singular form, and a list is optional. in this case 'dtype' would be more consistent. Agreed @JovanVeljanoski ?

maartenbreddels (changes requested):

maartenbreddels:
I like it, could you remove the VW parts? (we can squash the PR later, so just got git rm and commit/push)

packages/vaex-core/vaex/dataframe.py
JovanVeljanoski: Agreed

JovanVeljanoski (approved):
I like this as well. Very useful. I'll let @maartenbreddels comment on implementation details if any.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -4132,14 +4132,22 @@ def get_column_names(self, virtual=True, strings=True, hidden=False, regex=None)
...
+        >>> df.get_column_names(regex='x.*')
+        ['x', 'x2']
+        >>> df.get_column_names(dtype=[str])
+        ['s']
 
         :param virtual: If False, skip virtual columns
         :param hidden: If False, skip hidden columns
         :param strings: If False, skip string columns
         :param regex: Only return column names matching the (optional) regular expression
+        :param dtype: A list of dtypes to consider
```
maartenbreddels, line 4144:
```suggestion
        :param dtype: Only return column names that have this dtype (pass a list to accept multiple dtypes)
```

packages/vaex-core/vaex/dataframe.py
```diff
@@ -4132,14 +4132,22 @@ def get_column_names(self, virtual=True, strings=True, hidden=False, regex=None)
...
         :param alias: Return the alias (True) or internal name (False).
         :rtype: list of str
         """
+        if isinstance(dtype, type):
```
maartenbreddels, line 4148:
```suggestion
        if not isinstance(dtype, (list, tuple)):
```

packages/vaex-core/vaex/dataframe.py
```diff
@@ -5204,7 +5214,17 @@ def __getitem__(self, item):
         elif isinstance(item, Expression):
             expression = item.expression
             return self.filter(expression)
+        elif isinstance(item, type):
```
maartenbreddels, line 5217:
```suggestion
        elif isinstance(item, (type, vaex.datatype.DataType)):
```

tests/select_dtypes_test.py
```diff
@@ -0,0 +1,33 @@
...
+    assert df.get_column_names(dtype=float) == ['b']
+    assert df.get_column_names(dtype=np.float64) == ['b']
+    assert df.get_column_names(dtype=np.float32) == []
+    assert df.get_column_names(dtype=[int, float]) == ['a', 'b']
+    assert df.get_column_names(dtype=str) == ['d', 'e']
+    assert df.get_column_names(dtype=bool) == ['c']
+
+
+def test_getitem_by_dtypes(df):
+    assert df[int].get_column_names() == ['a']
```
maartenbreddels, line 26:
```suggestion
    assert df[int].get_column_names() == ['a']
    assert df[df.a.dtype].get_column_names() == ['a']
```

maartenbreddels (changes requested):
Yeah, i like it, some small corrections which I hope you can accept,

packages/vaex-core/vaex/dataframe.py
xdssio: done

maartenbreddels (approved):
I really like this PR, I'm a bit worried that if we don't document this (say in the tutorial) that this feature will do unnoticed

JovanVeljanoski:
Closed via #2160

PR #1768, opened 2021-12-15
handle chunked arrays in _pandas_dt_fix

chiaolun:
When loading feather files with pyarrow 6.0.1, dt methods fail because the underlying arrays are chunked. This change modifies the test to handle chunked arrays.

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -283,7 +283,9 @@ def scalar_timedelta(amount, unit):
 def _pandas_dt_fix(x):
     # see https://github.com/pandas-dev/pandas/issues/23276
     # not sure which version this is fixed in
-    if isinstance(x, pa.lib.TimestampArray):
+    if isinstance(x, pa.lib.TimestampArray) or (
+        isinstance(x, pa.lib.ChunkedArray) and isinstance(x.type, pa.lib.TimestampType)
+    ):
         return x.to_pandas()
     if not x.flags['WRITEABLE']:
         x = x.copy()
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -283,9 +283,7 @@ def scalar_timedelta(amount, unit):
 def _pandas_dt_fix(x):
     # see https://github.com/pandas-dev/pandas/issues/23276
     # not sure which version this is fixed in
-    if isinstance(x, pa.lib.TimestampArray) or (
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/functions.py
```diff
@@ -283,7 +283,9 @@ def scalar_timedelta(amount, unit):
 def _pandas_dt_fix(x):
     # see https://github.com/pandas-dev/pandas/issues/23276
     # not sure which version this is fixed in
-    if isinstance(x, pa.lib.TimestampArray):
+    if isinstance(x, pa.lib.TimestampArray) or (
+        isinstance(x, pa.lib.ChunkedArray) and isinstance(x.type, pa.lib.TimestampType)
+    ):
```
maartenbreddels, line 288:
```suggestion
    if vaex.array_types.is_arrow_array(x) and isinstance(x.type, pa.lib.TimestampType):
```

maartenbreddels (changes requested):
Cool, thanks for opening this PR. would be nice to test this as well. @JovanVeljanoski wrote a test on this in https://github.com/vaexio/vaex/commit/28cb5ed2d67ae25f3aae3ec990a0078d0ac5fe3b I think if we use the df_factory or the df_factory_arrowfixture, it would have been picked up.
(plenty of examples in the other tests I think(

chiaolun:
It looks like `df_factory_arrow_chunked` was not in fact producing chunked dfs, chasing that down now.

chiaolun:
Okay test working, catches the bug.

chiaolun:
If it is an arrow array but not a timestamp, i.e. if you call `dt` on a non-datetime column, the error would be quite inexplicable right now, what's the best way to throw an error for that?

maartenbreddels:
I was wondering about that myself as well, how we can improve this. I wonder if we should check at the accessor level what the allowed types are. E.g., just accessing `df.x.dt` where x is a float, should throw a `TypeError`.
@JovanVeljanoski @xdssio what do you think about that?
Similar for `df.str` (although here we support list of strings as well, so it's a bit more complicated).

Another solution is to do a typecheck at the decorator we use (register_function), and give a list of dtypes, or a callable that checks the type). Cons: may introduce runtime performance penalty, Pro: if you execute the function via a different path (e.g. loading a state/pipeline or calling df.func.df_...) you also get a proper error message.

JovanVeljanoski:
Yeah indeed good catch @chiaolun . 
I would go for.. the option with the smallest performance penalty :). But I am happy to hear other opinions!

maartenbreddels:
I think we can merge this first, this is already an improvement.

PR #1771, opened 2021-12-15
🔨 cleanup of progress bar and tree

maartenbreddels:
We now also report if it comes from cache or when an error occured.

Also moved out of utils.py to progress.py

### DIFF
--- a/ci/conda-env.yml
+++ b/ci/conda-env.yml
@@ -36,6 +36,7 @@ dependencies:
 - pytest-timeout
 - python-graphviz
 - py-xgboost
+- rich
 - s3fs>=2021.8.0
 - scikit-learn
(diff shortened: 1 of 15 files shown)


maartenbreddels:
should be more stable now with rich, and we now also see with [1] and [2] which pass over the data it did for which task, here it takes.3 passes.
image

maartenbreddels:
Also, using the context manager, everything will be added to the progress tree:
image

We also/still support via progress=True/'rich'

maartenbreddels:
note that in this last example, we didn't use delay=True, causing it to do 4 passes, instead of 2. @JovanVeljanoski I think this can be a good addition to the docs, explaining how vaex works.

PR #1772, opened 2021-12-15
Make hdf5 files always generate same file

fsiola:
We add in all occurances of `h5py.require_dataset` the parameter `track_times=False`. With this change, the file generated by exporting to hdf5 will always have the same hash.


maartenbreddels (approved):
Hi,

I made some changes, because there is some old code that is kept for backward compatibility (which should be removed in the future, but not changed).
For the rest this looks good, nice work!

maartenbreddels:
Great, many thanks for looking into this, and making such a good PR.

PR #1773, opened 2021-12-16
Handle empty arrow files

chiaolun:
Pass along schema explicitly so that empty files can be opened


maartenbreddels (approved):
Nice work, clear description, clear test.

Issue #1701, opened 2021-11-14
[BUG-REPORT] Vaex dataframe access crashes asyncio event loop tasks


schwingkopf:
**Description**
I'm experiencing issues when using vaex in the context of an asyncio event loop. The repeated interaction with the data frame seems to kill auxiliary tasks in the same event loop. The following code reproduces my issue:
```
import asyncio
import vaex
import numpy as np

async def vaex_task(df):
    while True:
        await asyncio.sleep(0.1)
        print(df.count())

async def aux_task():
    while True:
        await asyncio.sleep(0.3)
        print("Aux task alive")

df = vaex.from_dict(dict(x=np.linspace(0, 1., 100)))

loop = asyncio.get_event_loop()
loop.create_task(vaex_task(df))
loop.create_task(aux_task())
loop.run_forever()
```

Example output is:
```
100
100
Aux task alive
100
100
100
Aux task alive
100
100
Aux task alive
100
100
100
Aux task alive
100
100
Exception in callback <TaskWakeupMethWrapper object at 0x00000156BC40C4F8>(<Future finished result=None>)
handle: <Handle <TaskWakeupMethWrapper object at 0x00000156BC40C4F8>(<Future finished result=None>)>
Traceback (most recent call last):
  File "C:\Python37\lib\asyncio\events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
RuntimeError: Cannot enter into task <Task pending coro=<aux_task() running at D:\git_repos\genericosci\src\osci_backend\minimal_vaex_problem.py:20> wait_for=<Future finished result=None>> while another task <Task pending coro=<vaex_task() running at D:\git_repos\genericosci\src\osci_backend\minimal_vaex_problem.py:16>> is being executed.
Task was destroyed but it is pending!
task: <Task pending coro=<aux_task() running at D:\git_repos\genericosci\src\osci_backend\minimal_vaex_problem.py:20> wait_for=<Future finished result=None>>
100
100
100
100
100
100
100
100
100
100
```

The time after which the exception is thrown varies from run to run, but usually happens within the first 10 iterations of `aux_task` 

Does that look like a bug in vaex or am I doing something wrong here?

Thanks in advance for looking into this!

**Software information**
[...]

PR #1783, opened 2021-12-20
✨ do not use asyncio for the sync execute path

maartenbreddels:
We use generators now, so the 'top level'
execute_async can still do task switching while sharing the same
code with the sync version.

The remote executor will prefer to use the async version, and will
use nest asyncio when a sync call is done.


maartenbreddels:
Fixes 
 * #1701

Should fix
 * https://github.com/spyder-ide/spyder/issues/16183


Improvement of https://github.com/vaexio/vaex/pull/1546

PR #1788, opened 2021-12-21
🐛 keep virtual column order when renaming/dropping to not break state transfer

JovanVeljanoski:


maartenbreddels:
Ok, the underlying issue was the change of order of virtual columns

fsiola:
Any ideas on when a version with this fix will be out?

maartenbreddels:
Soon,I hope next week :)

PR #1793, opened 2021-12-21
Iterators for ML

xdssio:
This PR makes iterating on the dataframe simpler with focus for ML.
* Iteration in chunks.
* Minimal user code.
* Progressbar
* Multiple epochs

### Core.
*df.to_numpy()* and *df.to_chunks()* does exactly what you would expect.
* *df.to_numpy()* is a wrapper to evaluate_iterator but maintain the shape you would expect as values in chunks.
* *df.to_chunks()* wraps the iterations, num_epochs, shuffle and progressbar to simplify the common use cases.

Examplex:
```
df.to_numpy() == df.as_numpy().values

for i1, i2, chunk in df.to_numpy(chunks_size=100):
   <chunk is a numpy array with (chunk_size, len(columns))>

for i1, i2, chunk in df.to_chunks():
   <chunk is a vaex dataframe with the chunk size)>

```
### ML
```
df.ml.to_x_y(self, features=None, target=None, num_epochs=1, chunk_size=None, shuffle=False,
               selection=None, strings=True, virtual=True, parallel=True, progress=None, array_type='pandas')
```

Returns X, y as numpy or pandas, and allow simple iterations for online learning, wrapping the common use cases of num_epochs, chunks, shuffle, progressbar.

Examples:
```
for X, y in ds.ml.to_x_y(features=features, target=target, chunk_size=100, num_epochs=2, progressbar=True):
    model.fit(X,y)
 
```

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -18,6 +18,9 @@ defaults:
 
 jobs:
   main:
+    env:
+      CACHE_NUMBER_COMPILE: 5
+      MAMBA_ROOT_PREFIX: /dev/shm/mambaroot
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -27,12 +30,16 @@ jobs:
         python-version: [3.7, 3.8, 3.9]
 
     steps:
+    - name: disk space
+      if: ${{ (github.event.pull_request.head.repo.full_name == 'vaexio/vaex') && (matrix.os == 'ubuntu-latest') }}
+      run: |
+        df -h
     - uses: maxim-lobanov/setup-xcode@v1
       # alternative would be to upgrade tapi to 1100.0.11, we can possibly remove this in the future
       if: matrix.os == 'macOS-latest'
       with:
         xcode-version: "11.7.0"
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
       with:
         submodules: true
     - uses: actions/checkout@v2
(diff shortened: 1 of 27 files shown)


JovanVeljanoski:
Hi, 

Interesting PR. Here are my thoughts:

1. Regarding `df.to_numpy()` proposal:
Something very similar already exists actually, it is called `df.to_arrays()`, where by default you get the data in the raw original format. You can also pass an `array_type` kwarg `df.to_arrays(array_type='numpy')` to force the output to be numpy. There are examples of this in the [documentation](https://vaex.io/docs/example_io.html#DataFrame-representations). I think we could potentially have `df.to_numpy()`, but I think that is kinda reserved for any potential work on https://github.com/vaexio/vaex/pull/415. @maartenbreddels would know more. Besides, point 3 below would kind of handle this use-case in the short term anyway right?

2. Regarding `df.to_chunks(..)`. The `epoch` kwarg really tells me that this is aimed at ML applications, so i don't think it should live in "dataframe.py".  Besides there are already several similar methods to this, so I don't think it add much value. I think point 3 below satisfies this usecase completely.

3. This is potentially interesting. I guess it can abstract some of the code that is in `sklearn.IncrementalPredictor`, `river.RiverModel` and `catboostCatBoostModel`, and make the data iteration process available as an api. When would you use this however? Perhaps to train "online" models but .. "offline" :). The power of vaex is the expression system, the state etc.. so just training a model without then building an effective `transform` method seems line an under-utilization as you don't get the benefit of the overall "auto-pipeline" feature. That is why we have the model wrappers in `vaex-ml`.  In any case, I would be happy for this to go through, but I think the implementation needs to be improved a bit. For instance there is already `df.to_pandas()` that supports chunks, and is tested and efficient (and you have used it in your earlier PR :D). I think this can be refactored to be much more concise by reusing existing code. In turn that would require a small refactor of `sklearn.IncrementalPredictor`, `river.RiverModel` and `catboostCatBoostModel` I think, to make use fo this feature. 

These are my thoughts, I'd wait for @maartenbreddels next.

maartenbreddels:
> 1. I think we could potentially have `df.to_numpy()`, but I think that is kinda reserved for any potential work on [feat(core): implement nep13 and nep18 to allow sklearn integration #415](https://github.com/vaexio/vaex/pull/415). @maartenbreddels would know more. Besides, point 3 below would kind of handle this use-case in the short term anyway right?

I think the name is good, that PR is for the .numpy accessor which gives a numpy interface.

Looking at the implementation, I think `to_numpy` can use `to_arrays`, which would make refactoring code easier. I think also the docstring of both should make it's difference in output more clean, maybe we should explicitly say '2d numpy array'. I also wonder what happens if you call it with mixed types (string, int, float).

With @JovanVeljanoski 's blessing, I think it would be easier to move forward by opening a separate PR per feature. I think `to_numpy` is easiest to accept.

xdssio:
I think that `np.array(df.to_arrays(arrays_type='numpy')).T` is a lot of non-standard code to teach and explain while `to_numpy` is simple and self-explanatory and works within the Vaex system (and also apply to chunks). 

The `to_chunks`, is just a good general wrapper, around the common things. Of course we can let them add the progressbar and use the whatever to_pandas/to_arrays/to_dict etc, but if we can save the use code and make it simple, why not? save them the time to figure out our to combine the iterator with chunks and epochs, to use vaex progressbar instead of tqdm. It just help create cleaner code on the user side.

I think after this one, we can go and refactor all online algorithms and see best practices. 
There is now a way to do it with River, XGBoost, CatBoost, LightGBM, SKlearn, River, VowpalWabbit. the `to_x_y()` cover many if this nicely.

PR #1795, opened 2021-12-22
Rename map

xdssio:
Rename multiple columns

Simply iterate and rename columns if possible.
returns the new columns which were added.

I have created it on another function to allow backward compatibility.
We can also rename this function to `rename_columns` for example, or something else, or allow *name* param to be a dict in `rename()`.

Example:
```
df.renames({'x': 'x1', 'y': 'y1'})

```

What do you think?

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -18,6 +18,9 @@ defaults:
 
 jobs:
   main:
+    env:
+      CACHE_NUMBER_COMPILE: 5
+      MAMBA_ROOT_PREFIX: /dev/shm/mambaroot
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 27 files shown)


JovanVeljanoski:
Is this the same idea as https://github.com/vaexio/vaex/pull/1135?

I think we got stuck there on the API name, and whether the current `rename` should be refactored to support this but not break existing functionality (since it is used heavily internally), or exist as a separate method. 

My preference would be for a single `rename` method that can handle this as well (if that is possible). So less methods for users to need to remember. 

What do you think?

xdssio:
I have

> Is this the same idea as #1135?
> 
> I think we got stuck there on the API name, and whether the current `rename` should be refactored to support this but not break existing functionality (since it is used heavily internally), or exist as a separate method.
> 
> My preference would be for a single `rename` method that can handle this as well (if that is possible). So less methods for users to need to remember.
> 
> What do you think?

I initially had the same idea. I can check if name is a dict, and then apply it.
@maartenbreddels What do you think?

maartenbreddels:
Yes, let's put it in a single method.

PR #1800, opened 2021-12-24
Expression iterator

xdssio:
I found myself needing to zip columns all the time.
This makes expressions look and feel a bit more like a list.

1. Implement *\_\_len\_\_*
```
len(df.x) == len(df)
```

2. Implement *\_\_iterator\_\_*
* I used chunks of 1000

Example
```

# this holds 1000 values in memory at a time
for x in df.x:
  <x> as values

# This bring all values into memory
for x in df.x.values.tolist(): 
   pass

```

This makes zip works efficiently :
```
for x,y in zip(df.x,df.y):
   pass
 
```

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -18,6 +18,9 @@ defaults:
 
 jobs:
   main:
+    env:
+      CACHE_NUMBER_COMPILE: 5
+      MAMBA_ROOT_PREFIX: /dev/shm/mambaroot
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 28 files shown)


maartenbreddels:
> This makes zip works efficiently :
> 
> ```
> for x,y in zip(df.x,df.y):
>    pass
>  
> ```

This will be really efficient actually, since it you have two expressions that depends on a shared calculation, it will execute that twice. It will be more efficient to use:
```
for i, row in df['x', 'y'].iterrrows():
  print(row['x'], row['y'])
```
However, I really like the idea of using chunks! iterrows doesn't use that. Is it an idea to merge these ideas, and document them in the tutorial (@JovanVeljanoski ?)

PR #1802, opened 2021-12-28
ci: restore python 3.6

maartenbreddels:
Given CI runs much better, and we still distributed py36 wheels, lets see if we can get CI restored again.


maartenbreddels:
@JovanVeljanoski gave his blessing by call

JovanVeljanoski:
Blessed :) 

But let's keep all the Python versions for now, and see how flaky things get before we remove things

PR #1811, opened 2022-01-04
feat: configure logging using settings system

maartenbreddels:
We also use Rich to have beautiful default logging to console. This can be disabled using the settings.
image


maartenbreddels:
yeah, did some extra cleanups :)

Issue #1378, opened 2021-05-28
[FEATURE-REQUEST] Create OHLC from tick data


Penacillin:
**Description**
Trying to create OHLC  (Open High Low Close) data from price tick data. Can sort of create Open, High, Low, but Open is wrong and times have some offset. Not sure about close.

```python
import pandas as pd
import vaex
import numpy as np
import time
dates = pd.date_range("01-01-2019", "14-04-2020", freq="60s")
num = len(dates)
vdf = vaex.from_arrays(ts=pd.to_datetime(dates), x=np.random.randint(1, 1000, num))
print(vdf.head(17))
print()

# Desired output
print(vdf.to_pandas_df().resample('15Min', on='ts')['x'].ohlc())
print()

# Create Open High Low (Bit off)
vdf2 = vdf.groupby(by=vaex.BinnerTime(vdf.ts, resolution='m', every=15), agg={'O': vaex.agg.first('x', 'ts'), 'H': vaex.agg.max('x'), 'L': vaex.agg.min('x')})
print(vdf2)
print()

# Create Close?
vdf3 = vdf.groupby(by=vaex.BinnerTime(vdf.ts, resolution='m', every=15), agg={'C': vaex.agg.first('x', '-ts')})
print(vdf3)
```

Output:
```bash
#    ts                             x
0    2019-01-01 00:00:00.000000000  5
1    2019-01-01 00:01:00.000000000  20
2    2019-01-01 00:02:00.000000000  690
3    2019-01-01 00:03:00.000000000  434
4    2019-01-01 00:04:00.000000000  686
...  ...                            ...
12   2019-01-01 00:12:00.000000000  182
13   2019-01-01 00:13:00.000000000  530
14   2019-01-01 00:14:00.000000000  659
15   2019-01-01 00:15:00.000000000  929
16   2019-01-01 00:16:00.000000000  734

                     open  high  low  close
ts                                         
2019-01-01 00:00:00     5   894    5    659
2019-01-01 00:15:00   929   929  217    611
2019-01-01 00:30:00   424   966   41    228
2019-01-01 00:45:00    19   977   19     42
2019-01-01 01:00:00   137   989   96    686
...                   ...   ...  ...    ...
2020-04-13 23:00:00   756   994   99    204
2020-04-13 23:15:00   510   847    3      3
2020-04-13 23:30:00   128   898   62    501
2020-04-13 23:45:00   920   937   54    626
2020-04-14 00:00:00   694   694  694    694

[...]

PR #1834, opened 2022-01-15
[NOT TO BE MERGED] Agg first / cast or raise error in case expression & order_expression are not the same dtypes

yohplala:
Hi,

Following discussion in #1378 , I open this PR, not ready to be merged, far from here.
As a reminder of the initial proposal from @maartenbreddels 

--------------
_The issue is that first assumes that both arguments are of the same type, which is not true, and we don't check that. So we have to_
  - _Check that and raise an error, or cast_
  - _Implement mixed types for first_
 
_Would you like to take a look at 1, and see if you can open a PR for that? I think this is all in agg.py_

--------------

I feel very very stupid with this 1st attempt, I beg your pardon, this may be more a loss of time for you, but if you feel you can afford it, it seems I have tons of questions. I ask clemency for my low level of understanding / coding skills.

--------------
**Installation**
As you pointed out @maartenbreddels , it seems installing _compilers_ was the missing trick in my installation.
Also, 1st question,

  - 1. Instead of using `pip install -e ".[dev]"`, I used `pip install -e .`. Do you know what can be the difference?

--------------
**Implementation**
Following your comment _I think this is all in agg.py_, I implemented checks on dtype in `AggregatorDescriptorBasic._prepare_types()`. File modifications can be seen in this PR.
I also started a test case `test_first` in `tests.agg_test`.

  - 2. Is this where you see the changes to be made?
  - 3. But already, these changes do not work.
     - 3.1 The 1st assert (that I commented out with last commit), which do not present any dtype trouble: the order of agg result -column `x_first`, albeit containing the right values, is not ordered as expected. The assert fails on `[7, 11, 9] == [7, 9, 11]`. Is it expected that the `groupby` with `agg.first` does not maintain order between groups?
     - 3.2 In `AggregatorDescriptorBasic._prepare_types()`, I could see that `dtype` is retrieved with `df['...'].data_type().index_type`. I could check on a dummy test that both `data_type().index_type` and `data_type()` returns actually the same type of object `vaex.datatype.DataType`. Please, what does `index_type` over `data_type()`?
     - 3.3 Should have been the 1st question: With changes made, I am modifying the dtype of some columns. Is it what you have in mind?

--------------
**AggregatorDescriptorMean**

I also had a look to how some other `Aggregator` class are implemented, and I found lines I just don't know / understand the syntax, and more specifically in `AggregatorDescriptorMean.add_tasks()`.
  - 4. Please what does mean a line like:
        `expression = expression_sum = expression = df[str(self.expression)]`
         ?
         Is there a kind of short-circuit logic to assign `expression` variable?

I am sorry, I feel very stupid with these questions.
Please, feel free to disregard these, I understand this is close to ask a free course on vaex and python programming here :)
Bests,

### DIFF
--- a/packages/vaex-core/vaex/agg.py
+++ b/packages/vaex-core/vaex/agg.py
@@ -259,15 +259,12 @@ class AggregatorDescriptorBasic(AggregatorDescriptor):
             if self.short_name in ['sum', 'summoment']:
                 self.dtype_out = self.dtype_in.upcast()
             if self.short_name == "first":
-                oe_dtype = df[str(self.expressions[1])].data_type().index_type
-                if self.dtype_in != oe_dtype:
-                    print(f'dtype_in: {self.dtype_in}')
-                    print(type(self.dtype_in))
-                    print(f'oe_dtype: {oe_dtype}')
-                    print(type(oe_dtype))
-                    if np.can_cast(oe_dtype, self.dtype_in):
-                        df[str(self.expressions[1])] = df[str(self.expressions[1])].astype(self.dtype_in)
-                    elif np.can_cast(self.dtype_in, oe_dtype):
+                oe_dtype = str(df[str(self.expressions[1])].data_type().index_type)
+                e_dtype = str(self.dtype_in)
+                if e_dtype != oe_dtype:
+                    if np.can_cast(oe_dtype, e_dtype):
+                        df[str(self.expressions[1])] = df[str(self.expressions[1])].astype(e_dtype)
+                    elif np.can_cast(e_dtype, oe_dtype):
                         df[str(self.expressions[0])] = df[str(self.expressions[0])].astype(oe_dtype)
                         self.dtype_in = df[str(self.expressions[0])].data_type().index_type
                     else:
--- a/packages/vaex-core/vaex/agg.py
+++ b/packages/vaex-core/vaex/agg.py
@@ -258,6 +258,20 @@ class AggregatorDescriptorBasic(AggregatorDescriptor):
                 self.dtype_out = DataType(np.dtype('int64'))
             if self.short_name in ['sum', 'summoment']:
                 self.dtype_out = self.dtype_in.upcast()
+            if self.short_name == "first":
+                oe_dtype = df[str(self.expressions[1])].data_type().index_type
+                if self.dtype_in != oe_dtype:
+                    print(f'dtype_in: {self.dtype_in}')
+                    print(type(self.dtype_in))
+                    print(f'oe_dtype: {oe_dtype}')
+                    print(type(oe_dtype))
+                    if np.can_cast(oe_dtype, self.dtype_in):
+                        df[str(self.expressions[1])] = df[str(self.expressions[1])].astype(self.dtype_in)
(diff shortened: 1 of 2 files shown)


yohplala:
Closed by #1848

maartenbreddels:
> 1. Instead of using `pip install -e ".[dev]"`, I used `pip install -e .`. Do you know what can be the difference?

The first one should install extra dependencies, but also @JovanVeljanoski told me it does not work, we should look into that.


> 4\. expression = df[str(self.expression)]

This makes sure that `expression` is of the Expression type (the original expression can be a strings as well, and `str` has no effect on that).

I think the code is just confusing as it is, it's assigning it twice to expression, which might be a copy paste artifact, if you want to clean this up in a PR, you are very welcome.

> * 3.3 Should have been the 1st question: With changes made, I am modifying the dtype of some columns. Is it what you have in mind?

Although it will work, I think we could have done it just be modifying self.expressions in agg.py, but given #1848 this is ok.

Thanks for you PR, I think this was good work, but if we can get #1848 in, that would be even better.

yohplala:
Hi @maartenbreddels ,
I am sorry, I am pursuing the discussion in this thread, we can move it elsewhere.
I wanted to check my vaex install with `pytest` before making any modification (the one we identified at point 4:
`expression = expression_sum = expression = df[str(self.expression)]`

So after installing vaex (branch master):
`pip install -e .`
I test it:
`python -m pytest`

Unfortunately, I get this error message:
```bash
================================== test session starts ==================================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
rootdir: /home/yoh/Documents/code/vaex, configfile: pytest.ini
plugins: anyio-2.2.0
collecting ... Failed to import pybind11_tests from pytest:
  ModuleNotFoundError: No module named 'pybind11_tests'
collected 0 items / 1 error                                                             

======================================== ERRORS =========================================
_____________________________ ERROR collecting test session _____________________________
packages/vaex-core/vendor/pybind11/tests/conftest.py:232: in _test_import_pybind11
    import pybind11_tests  # noqa: F401 imported but unused
E   ModuleNotFoundError: No module named 'pybind11_tests'

During handling of the above exception, another exception occurred:
../../../anaconda3/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1030: in _gcd_import
    ???
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
../../../anaconda3/lib/python3.9/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
    exec(co, module.__dict__)
packages/vaex-core/vendor/pybind11/tests/conftest.py:239: in <module>
    _test_import_pybind11()
packages/vaex-core/vendor/pybind11/tests/conftest.py:236: in _test_import_pybind11
    sys.exit(1)
E   SystemExit: 1
=================================== warnings summary ====================================
```

Do you have any idea what would be wrong here?

PR #1839, opened 2022-01-17
Re-run flaky test

JovanVeljanoski:
Sometimes this one test is flaky because of "chance". I finally found out to do an easy fix :)


JovanVeljanoski:
This was actually fixed via the random seed a while ago. So this PR is not needed.

maartenbreddels:
good to have this in history, didn't know about it.

PR #1317, opened 2021-04-16
Support multiple example datasets

JovanVeljanoski:
With this PR, the `vaex.example(dataset)` command can support multiple datasets. 

At the moment the plan is to provide the following datasets:
- `dataset='astronomy'` - Helmi & De Zeeuw simulation data (default)
- `dataset='taxi'` - Sample of the NYC Yellow Cabs taxi data
- `dataset='iris'` - The Iris flower dataset (via `vaex.ml`)
- `dataset='titanic'` - The Titanic passenger list dataset (via `vaex.ml`)

PR #1868, opened 2022-01-28
🐛 vaex.ml optional import makes vaex-core depend on vaex.ml

maartenbreddels:


maartenbreddels:
@JovanVeljanoski  I think that line in `__init__.py` was a leftover from https://github.com/vaexio/vaex/pull/1317 where we first depended on vaex.ml for the dataset

JovanVeljanoski (approved):

PR #1873, opened 2022-01-31
🐛 to_dask_array broke with most recent dask (2022.1.1)

maartenbreddels:
Due to https://github.com/dask/dask/pull/7417

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -3325,9 +3325,14 @@ class DataFrame(object):
         name = 'vaex-df-%s' % str(uuid.uuid1())
         def getitem(df, item):
             return np.array(df.__getitem__(item).to_arrays(parallel=False)).T
-        dsk = da.core.getem(name, chunks, getitem=getitem, shape=self.shape, dtype=dtype.numpy)
-        dsk[name] = self
-        return da.Array(dsk, name, chunks, dtype=dtype.numpy)
+        # broken since https://github.com/dask/dask/pull/7417
+        if hasattr(da.core, "getem"):
+            dsk = da.core.getem(name, chunks, getitem=getitem, shape=self.shape, dtype=dtype.numpy)
+            dsk[name] = self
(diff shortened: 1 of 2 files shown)


gjoseph92:
@maartenbreddels sorry about this change. Just curious, why not use the public [`da.from_array`](https://docs.dask.org/en/stable/generated/dask.array.from_array.html#dask.array.from_array) function instead of relying on these internals? You can specify your own `getitem`, `chunks`, and `meta` (and `name`), so from a quick skim I'd think it would work fine.

ian-r-rose:
Sorry about this @maartenbreddels! I thought it was safe to change the function name, since (a) I viewed it as a private function (though Dask isn't very good about `_` prefixes or anything), and (b) the signature of the function changed anyways. I'd echo @gjoseph92's question: is there a reason to not use the more public (and stable) `da.from_array`?

ludwigschwardt:
Hi @ian-r-rose and @gjoseph92, I hit the same snag in my library. I asked myself the same question, since it was so long ago 😅 It turned into quite an exploration, as documented in katdal/katdal#351, in case you are interested. Bottom line: I still like `getem` / `graph_from_arraylike` 😄

maartenbreddels:
Good read @ludwigschwardt 

FYI: 2022.4.0 again break how we use dask, see https://github.com/vaexio/vaex/pull/2003
our getter isn't being called

ludwigschwardt:
Thanks! I found that it helped to spell out the options, given that I haven't worked on this code in 4 years...

And thanks for the heads-up too.

The irony is that I converted back to `from_array` to be "future-proof" but then discovered that it only works because dask/dask#7417 / 2022.03.0 didn't optimise away my custom slices anymore, a [bug](https://github.com/dask/dask/issues/8753) for most but a feature for me :joy: I can get around this by specifying an explicit `getitem` that prevents optimisation, though.

I guessed that this behaviour will be short-lived, and here we are in 2022.04.0. Now my getter still works but my putter is bust 🙂

PR #1874, opened 2022-01-31
👷 pip from conda-forge does not unlist python 3.6

maartenbreddels:


maartenbreddels:
Seems like this was picked up https://github.com/conda-forge/pip-feedstock/pull/88 so we can undo this soon

Issue #1883, opened 2022-02-04
[BUG-REPORT]


JovanVeljanoski:
**Description**
This continues the issue raised in this post:
https://github.com/vaexio/vaex/issues/1282#issuecomment-1028264122

**Symmary** 
Left join does not work when the joining is done on a virtual column create via `apply`.

**Software information**
 - Vaex version: 4.7.0
 - Vaex was installed via: pip / conda-forge / from source

JovanVeljanoski:
I was able to reproduce the issue, and will make a test for it. I hope we can fix this soon.

PR #1888, opened 2022-02-08
test(core): test for join on virtual col created via apply

JovanVeljanoski:
Fixes https://github.com/vaexio/vaex/issues/1883

- [x] Make a unit test exposing the issue
- [ ] Make test pass


maartenbreddels:
The test has an issue, it doesn't expose a bug I thinl.

PR #1899, opened 2022-02-09
fix deprecated `np.bool`

Ben-Epstein:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


JovanVeljanoski:
This is nice! 

I think it would be nice if this is propagated throughout the whole project, not just in one module, what do you think?

Ben-Epstein:
@JovanVeljanoski definitely. Migrated to https://github.com/vaexio/vaex/pull/1900

PR #1936, opened 2022-02-17
🐛 apply with filtered data could give wrong dtypes

JovanVeljanoski:
This is quite an exotic bug, discovered by https://github.com/vaexio/vaex/discussions/1933.

Sometimes doing `df.y.dt.strftime('%Y')` results in a dtype of `null` (?). However, this only happens if the dataframe contains another column, say `x` of dtype `string` that has at least two missing values, of which one _has to be in the first position_.  If one does `dropna()` in that column `x`, the result of doing `strftime` on `y` gives weird dtype (see unit test for better explanation). 

See https://github.com/vaexio/vaex/discussions/1933 for a temporary workaround.

Checklist 
- [x] make unit test
- [ ] make test pass

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -6411,7 +6411,7 @@ class DataFrameLocal(DataFrame):
                 deps |= filter_deps
             columns = {k: dataset[k][:] for k in deps if k in dataset}
 
-            if self.filtered:
+            if self.filtered and filtered:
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Thanks @JovanVeljanoski , those last examples were really useful. I think in the end we 

Since we filtered the evaluated data, we ran apply on an empty
array, which by default (of numpy) will give a float64 array.

I think we may hit this issue again on really empty dataframes, and we may want to take another look at https://github.com/vaexio/vaex/pull/1177 to implement stable dtypes for apply.

PR #1950, opened 2022-02-28
🐛 pyarrow 7 seems to have changed .nbytes for dict encoded arrays

maartenbreddels:
We're tracking this consistently for testing


maartenbreddels:
Included in #1920

PR #1958, opened 2022-03-04
Update expression.py

Ben-Epstein:
handle chunked arrays


maartenbreddels:
Could you write a test that exposes the issue, so we don't hit this issue in the future again?

maartenbreddels:
Don't worry about the code quality of the test, it just saves me time to reproduce the exact issue.

Ben-Epstein:
@maartenbreddels added a test! It will fail currently

maartenbreddels:
Thanks Ben, do you mean the tests fails against master? (since now it is green)

Ben-Epstein:
> Thanks Ben, do you mean the tests fails against master? (since now it is green)

@maartenbreddels exactly, this will fail against master without my change

maartenbreddels:
perfect, thanks!

PR #1967, opened 2022-03-10
enable proper `in` checking

Ben-Epstein:
current behavior
```
import vaex
df = vaex.from_arrays(
    id=vaex.vrange(0, 200_000)
)

299_999 in df.id  # True but wrong
```

proposed
```
299_999 in df.id # False
```

### DIFF
--- a/tests/expression_variables_test.py
+++ b/tests/expression_variables_test.py
@@ -44,3 +44,13 @@ def test_non_identifiers():
     df['valid'] = df['2']
     assert df['valid'].variables(ourself=True) == {'x', 'y', '2', '1', 'valid'}
     assert df['valid'].variables(include_virtual=False) == {'x', 'y'}
+
+    
+def test_expression_in():
+    df = vaex.from_dict({"id":[1,2,3,4,5,6], "y":[5.4,3.5,1.1,2.4,3.7,3.5]})
+    for i in range(1,7):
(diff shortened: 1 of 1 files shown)


JovanVeljanoski:
This materializes the column right? That's.. not ideal...

Ben-Epstein:
@JovanVeljanoski after speaking with @maartenbreddels it was my understanding is that `.values` is 0 mem copy. https://vaexio.slack.com/archives/C017EEHSQ84/p1646850161602389

JovanVeljanoski:
Yes because in that example, you created an in memory dataset. So your data (and that column) is already in memory. 
But if you read an hdf5/arrow/parquet file, you first will put it to memory (since otherwise you are reading from disk) then you gonna do the stuff that you like.

Ben-Epstein:
@JovanVeljanoski thanks for the explanation, I didn't understand that. So is there a better approach here? Currently returning `True` always could lead to many side-effects for end users

Ben-Epstein:
@JovanVeljanoski how about now?

maartenbreddels:
I like it, I just don't know why `7 in df["id"]` evaluates to True 🤷

maartenbreddels:
https://vaex.io/docs/installing.html#for-developers should help!

PR #1971, opened 2022-03-14
add `drop_empty_columns` function to DataFrame

Ben-Epstein:


maartenbreddels:
Hi Ben,

this looks useful, but I'm wondering if we start adding many of these helper function, the API is gonna explore (like Pandas).

@Ben-Epstein  @JovanVeljanoski what about adding this to vaex-contrib, and have a accessor, such as
`df.helpers.drop_empty_columns()`

cheers,

Maarten

PR #1979, opened 2022-03-18
Fix: add pool locking in multiprocessing

JovanVeljanoski:


packages/vaex-core/vaex/multiprocessing.py
```diff
@@ -1,3 +1,5 @@
+import threading
+
```
Ben-Epstein, line 2:
```suggestion
from multiprocessing import Pool

```

packages/vaex-core/vaex/multiprocessing.py
```diff
@@ -1,3 +1,5 @@
+import threading
```
Ben-Epstein, line 1:
```suggestion
from multiprocessing import Pool
import threading
```

maartenbreddels (approved):
Thanks Jovan and Ben!
I think as it is, it's good to go.
I agree with Ben that the import in the function isn't nice, but assuming it had a reason (my bad for not commenting on it), lets keep it as it is.

PR #1981, opened 2022-03-18
Pin setup-gcloud to v0 instead of master

sethvargo:
setup-gcloud will be updating the branch name from master to main in a future release. Even though GitHub will establish redirects, this will break any GitHub Actions workflows that pin to master. This PR updates your GitHub Actions workflows to pin to v0, which is the recommended best practice.


maartenbreddels:
Thanks!

PR #1988, opened 2022-03-24
Fix: Support for isnan, isna, ismissing when the underlying data is arrow

JovanVeljanoski:
Added unit tests as well.


maartenbreddels:
Nice work, to bad CI is not so green :) (not your fault)

PR #1990, opened 2022-03-25
✨(core): all groupers/binner have a sort and ascending argument

maartenbreddels:
This means the resulting dataframe can be presorted, this is useful
if we already know we want the aggregated dataframe to be
sorted afterwards, which is much faster now.


maartenbreddels:
failure is unrelated

PR #1993, opened 2022-03-28
🐛 write to cached filesystem when metadata argument is needed

JovanVeljanoski:


packages/vaex-core/vaex/file/__init__.py
```diff
@@ -341,7 +342,7 @@ def _make_argument_optional(f, **defaults):
     def wrapper(*args, **kwargs):
         kwargs = kwargs.copy()
         for name, value in defaults.items():
-            if name in params and name not in kwargs:
+            if (name not in params) or (name not in kwargs):
```
maartenbreddels, line 345: I dont understand, if `name not in params` means that the argument does not exists for `f`, how can this work then? We cannot pass argument to f that is does not support.

packages/vaex-core/vaex/file/__init__.py
```diff
@@ -1,4 +1,5 @@
 __author__ = 'breddels'
+from ast import arg
```
maartenbreddels, line 2:
```suggestion
```

PR #1999, opened 2022-04-01
✨ add unique aggregator

maartenbreddels:
Split off from #1987


armandgurgu23:
Is there an update on the timeline for when this feature will be shipped? This aggregator can be handy in many EDA usecases :).

maartenbreddels:
Hi,

no sorry, no timeline yet :), there is no funding for this feature request currently, so that usually means the priority is lower.
Would love to work on it though if we had some more time.

Regards,

Maarten

PR #2003, opened 2022-04-04
fix: avoid dask 2022.4.0

maartenbreddels:
Again breakage, followup of https://github.com/vaexio/vaex/pull/1873

I think this is taking too much time, and we should drop this, what do you think @JovanVeljanoski ?


maartenbreddels:
I see https://github.com/dask/dask/pull/8827 as a possible candiate for introducing the issue, it seems our getter in Expression.to_dask_array is not getting called at all.

ludwigschwardt:
@maartenbreddels, it looks like the custom getter (`getitem` below) is only used in `da.core.graph_from_arraylike` when `inline_array=True`:
https://github.com/dask/dask/blob/0b36d7fcaf54ee9a78fff4b07f124cb0c8741cdf/dask/array/core.py#L291-L319

Possibly a typo in the latest dask?

PR #2009, opened 2022-04-11
fix: value_counts would give too many null values

maartenbreddels:
internally we wouldn't grow the null bitmask


NickCrews:
@maartenbreddels It looks like `[ci: do not run notebooks for py36](https://github.com/vaexio/vaex/pull/2009/commits/7ba15b1435f6b886def9c8392c98d0ea9dad9c52)` actually makes the notebooks run ONLY for python3.6, is that what you intended?

PR #2023, opened 2022-04-18
Changed deprecated numpy.float to numpy.float64

Nicholas-Schaub:
In some unit tests we discovered that `vaex` is referencing the old `numpy.float` type rather than `numpy.float64` like similar methods. This PR simples changes `numpy.float -> numpy.float64`, which should prevent future errors and silence numpy warnings.


Nicholas-Schaub:
@maartenbreddels I think this should be an easy approval. Is there anyone that should be tagged as a reviewer?

JovanVeljanoski:
We haven't gotten around to it. Just leave it and when we can we will take a look.

maartenbreddels:
Awesome, thanks, sorry for the long wait!

Issue #2038, opened 2022-04-29
[BUG-REPORT] Weird `TypeError` on chained filters


astrojuanlu:
**Description**
I am experiencing a weird `TypeError` that appears when chaining several filters and performing an operation on the resulting dataframe. With this code:

```python
df = vaex.open("data_problematic.parquet")

df_sample = df.head(1_000)
df_sample["DEP_DELAY"].mean(progress="rich")  # Works OK

not_cancelled_sample = df_sample[df_sample["CANCELLED"] == 0.0]
not_cancelled_sample["DEP_DELAY"].mean(progress="rich")  # Works OK

delayed_sample = not_cancelled_sample[not_cancelled_sample["DEP_DELAY"] > 0]
delayed_sample["DEP_DELAY"].mean(progress="rich")  # FAILS
```

I get this error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [13], in <cell line: 10>()
      7 not_cancelled_sample["DEP_DELAY"].mean(progress="rich")  # Works OK
      9 delayed_sample = not_cancelled_sample[not_cancelled_sample["DEP_DELAY"] > 0]
---> 10 delayed_sample["DEP_DELAY"].mean()

File /opt/conda/lib/python3.9/site-packages/vaex/expression.py:919, in Expression.mean(self, binby, limits, shape, selection, delay, progress)
    917 del kwargs['self']
    918 kwargs['expression'] = self.expression
--> 919 return self.ds.mean(**kwargs)

File /opt/conda/lib/python3.9/site-packages/vaex/dataframe.py:1064, in DataFrame.mean(self, expression, binby, limits, shape, selection, delay, progress, edges, array_type)
   1042 @docsubst
   1043 @stat_1d
   1044 def mean(self, expression, binby=[], limits=None, shape=default_shape, selection=False, delay=False, progress=None, edges=False, array_type=None):
   1045     """Calculate the mean for expression, possibly on a grid defined by binby.
   1046 
   1047     Example:
   (...)
   1062     :return: {return_stat_scalar}
   1063     """
-> 1064     return self._compute_agg('mean', expression, binby, limits, shape, selection, delay, edges, progress, array_type=array_type)
   1065     logger.debug("mean of %r, with binby=%r, limits=%r, shape=%r, selection=%r, delay=%r", expression, binby, limits, shape, selection, delay)
   1066     expression = _ensure_strings_from_expressions(expression)

[...]

PR #2043, opened 2022-05-03
Fix: Filtered dataframes after dropping missing/na values

JovanVeljanoski:
This fixes the issue raise in https://github.com/vaexio/vaex/issues/2038

checklist:
- [x]  expose the issue via a unit-test
- [ ] make test pass


maartenbreddels:
Thanks for the test!

PR #2004, opened 2022-04-04
Feat: add `describe` method to `groupby`

JovanVeljanoski:
This PR add/proposes a feature requested in https://github.com/vaexio/vaex/issues/1998

The feature can be best described with an example:
```python
import vaex
df = vaex.datasets.titanic()
res = df.groupby('pclass').describe(['age', df.sex])
print(res)

  #    pclass    age_count    age_count_na    age_mean    age_std    age_min    age_max    sex_count    sex_count_na
  0         1          284              39     39.1599    14.5224     0.9167         80          323               0
  1         2          261              16     29.5067    13.6125     0.6667         70          277               0
  2         3          501             208     24.8164    11.9463     0.1667         74          709               0
```

This is a convenience method, but I think it makes sense to add.

PR #2072, opened 2022-06-03
✨ df.struct.flatten to get rid of arrow structs

maartenbreddels:


maartenbreddels:
This is a split of from #2004 which need this feature

JovanVeljanoski:
Should `df.struct.flatten()` also take a `columns` argument ? Maybe by default none (so apply to all struct cols), but maybe a user can choose a subset of cols to flatten?

maartenbreddels:
Yes, that makes total sense.

maartenbreddels:
OK, argument added, as well as a recursive flag.

JovanVeljanoski (approved):

PR #2076, opened 2022-06-09
support native open json

Ben-Epstein:
to support the following symmetricly

```
import vaex

df = vaex.example()
df.export("file.json")
vaex.open("file.json")
```


packages/vaex-core/vaex/__init__.py
```diff
@@ -219,6 +219,8 @@ def open(path, convert=False, progress=None, shuffle=False, fs_options={}, fs=No
             _, ext, _ = vaex.file.split_ext(path)
             if ext == '.csv':  # special case for csv
                 return vaex.from_csv(path, fs_options=fs_options, fs=fs, convert=convert, progress=progress, **kwargs)
+            if ext == ".json":  # special case for csv
```
NickCrews, line 222: A bogus comment got copy pastad

maartenbreddels, line 222:
```suggestion
            if ext == ".json":  # special case for json
```

PR #2101, opened 2022-06-24
fix: correct $VAEX_PATH_HOME -> $VAEX_PATH

maartenbreddels:
For some reason we used $VAEX_PATH_HOME, while documented it as $VAEX_HOME

It should have been $VAEX_HOME. For backward compatibility we still
use $VAEX_PATH_HOME if is is set.

cc @eff-kay

### DIFF
--- a/packages/vaex-core/vaex/utils.py
+++ b/packages/vaex-core/vaex/utils.py
@@ -247,11 +247,13 @@ class Timer(object):
 def get_vaex_home():
     '''Get vaex home directory, defaults to $HOME/.vaex.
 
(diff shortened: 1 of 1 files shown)


JovanVeljanoski (approved):

eff-kay:
@maartenbreddels if the `VAEX_PATH_HOME` is there just because of us, then you can safely remove it 🙂

maartenbreddels:
Yes, but others might also use it, I've been bitten by backward incompatibilities too many times in my life that I can't help but avoid them :)

Issue #2057, opened 2022-05-20
[BUG-REPORT] Dataframe with virtual columns and specific column names fails export to csv


grafail:
**Description**
Issue when converting to csv with specific column names like "description" or "dtypes". I believe this has to do with the attributes on the DataFrame object being overridden when they coincide with a column name.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: {'vaex': '4.9.1', 'vaex-core': '4.9.1', 'vaex-viz': '0.5.1', 'vaex-hdf5': '0.12.1', 'vaex-server': '0.8.1', 'vaex-astro': '0.9.1', 'vaex-jupyter': '0.7.0', 'vaex-ml': '0.17.0', 'vaex-graphql': '0.2.0'}
 - Vaex was installed via: source
 - OS: Ubuntu 20.04

**Additional information**
Source code to reproduce:
```python
import vaex

df = vaex.open("test.csv")
df = df.as_arrow()
df.export_csv("test2.csv")
```
test.csv:
```
description,test
a,b
```
```python
Traceback (most recent call last):
  File "<Path>/python3.8/site-packages/vaex/scopes.py", line 113, in evaluate
    result = self[expression]
  File "<Path>/python3.8/site-packages/vaex/scopes.py", line 198, in __getitem__
    raise KeyError("Unknown variables or column: %r" % (variable,))
KeyError: "Unknown variables or column: 'as_arrow(____description)'"
```

grafail:
This seems to be resolved when using the CsvWriter from pyarrow

maartenbreddels:
Hi,

thanks for the report, I cannot reproduce this on master now and it also seems to work with tag `core-v4.9.1`.
Strange!
@JovanVeljanoski can you try?

Regards,

Maarten

JovanVeljanoski:
I can confirm this as a bug. 

One must also run the `df = df.as_arrow()` line to trigger the problem, without it things work just fine.

I will try to make a test about this soon.

PR #2117, opened 2022-07-18
Fix: Export csv with protected column names from arrow data structures

JovanVeljanoski:
Fixes https://github.com/vaexio/vaex/issues/2057

- [x] Create a test
- [ ] make test pass


maartenbreddels:
Fix from ci reference https://github.com/pydata/xarray/issues/6818

Issue #2113, opened 2022-07-05
[BUG-REPORT] `describe_categorical` in interchange columns is a tuple, not a dict


honno:
In the [interchange protocol](https://data-apis.org/dataframe-protocol/latest/API.html), `describe_categorical` should return a dict (mind the spec's API type annotation is faulty), but Vaex returns a tuple

https://github.com/vaexio/vaex/blob/35c250d585f889272b8ef1096de6fa5462816f52/packages/vaex-core/vaex/dataframe_protocol.py#L443

This prevents interchanging dataframes with categorical columns, e.g. with https://github.com/pandas-dev/pandas/pull/46141

```python
>>> import numpy as np
>>> import vaex
>>> df = vaex.from_items(("foo", np.asarray([4, 2, 1, 3, 3], dtype="int8")))
>>> df = df.categorize("foo")
>>> from pandas.api.exchange import from_dataframe
>>> from_dataframe(df)
.../pandas/core/exchange/from_dataframe.py:184, in categorical_column_to_series(col)
    169 """
    170 Convert a column holding categorical data to a pandas Series.
    171 
   (...)
    180     that keeps the memory alive.
    181 """
    182 categorical = col.describe_categorical
--> 184 if not categorical["is_dictionary"]:
    185     raise NotImplementedError("Non-dictionary categoricals not supported yet")
    187 mapping = categorical["mapping"]
TypeError: tuple indices must be integers or slices, not str
```

honno:
I'll submit a PR for this

Issue #2093, opened 2022-06-22
[BUG-REPORT] Interchange `Column.size` returns 0d arrays as opposed to Python `int`


honno:
Dataframes via the [interchange protocol](https://data-apis.org/dataframe-protocol/latest) will store return interchange columns which give an erroneous `size`.

```python
>>> import vaex
>>> df = vaex.from_dict({"a": ["foo", "bar"]})
>>> interchange_df = df.__dataframe__()
>>> col = interchange_df.get_column(0)
>>> col.size
array(2)  # should be a Python integer
```

This is due to `Column.size` return the output of a count operation without converting it to a Python integer.

https://github.com/vaexio/vaex/blob/8fada6dd422a82b6f3b50f5c34a46b412536bac4/packages/vaex-core/vaex/dataframe_protocol.py#L318-L323

This prevents interop with https://github.com/pandas-dev/pandas/pull/46141 due to this [line](https://github.com/pandas-dev/pandas/blob/faacb72b96b7d6604e5eff13f0e224cf741ff8d8/pandas/core/exchange/from_dataframe.py#L271) assuming (correctly) that `size` should be a Python integer.

Vaex was built locally from source (upstream `master`) on Ubuntu 20.04.

honno:
Will submit a PR fixing this

PR #2150, opened 2022-08-04
Interchange protocol fixes and updates

honno:
* Resolves https://github.com/vaexio/vaex/issues/2093 (was useful so I could test the updates with `pandas.interchange`)
* `Column.describe_categorical`
  * Resolves https://github.com/vaexio/vaex/issues/2113
  * Per https://github.com/data-apis/dataframe-api/pull/69, updates it to return `categories` as opposed to `mapping`, which is a `Column` containing the labels
  * Update existing tests to use new `Column.describe_categorical` behaviour
* Categorical columns with string labels weren't interchanged correctly (from vaex and other adopting libraries), so I made the necessary fixes to `convert_categorical_column`
* Fixed some internal type hints
* Wrote some xfailed tests from some stirng column interchange things I encountered, which should be useful down-the-line.


tests/dataframe_protocol_test.py
```diff
@@ -369,3 +394,66 @@ def test_smoke_get_buffers_on_categorical_columns(df_factory):
...
+    x = np.array(data)
+    df = df_factory(x=x)
+    if isinstance(df["x"].values, pa.lib.ChunkedArray):
+        pytest.xfail()
+    df = df.categorize('x', labels=labels)
+    interchange_df = df.__dataframe__()
+    interchange_col = interchange_df.get_column_by_name('x')
+    roundtrip_df = _from_dataframe_to_vaex(interchange_df)
+    data_as_labels = [labels[i] for i in data]
+    assert roundtrip_df["x"].values.tolist() in [data, data_as_labels]  # TODO
```
honno, line 458: I wasn't sure if the values should be the codes (`[3, 1, 1, 2, 0]`), or the the codes-as-their-labels (e.g. `["qux", "bar", "bar", "baz", "foo"]`).

honno:
I didn't realise CI has 3.6 and 3.7 jobs. I see https://github.com/vaexio/vaex/pull/1802, so assume these versions want to be generally maintained for? I have used f-stirngs in this PR, although note the df protocol implementation and tests already used them.

packages/vaex-core/vaex/dataframe_protocol.py
```diff
@@ -315,12 +323,17 @@ def __init__(self, column: vaex.expression.Expression, allow_copy: bool = True)
...
         Size of the column, in elements.
+
+        Corresponds to DataFrame.num_rows() if column is a single chunk;
+        equal to size of this current chunk otherwise.
+
+        Is a method rather than a property because it may cause a (potentially
+        expensive) computation for some dataframe implementations.
         """
-        return self._col.df.count("*")
+        return int(self._col.df.count("*"))
```
maartenbreddels, line 336: Just a quick change while we change this line
```suggestion
        return int(len(self._col.df))
```

Btw, just noticed that pandas doesn't have this as a property:
https://github.com/pandas-dev/pandas/blob/f0814bcca41a116794ff926b571ec6c2d15685d7/pandas/core/interchange/column.py#L84

tests/dataframe_protocol_test.py
maartenbreddels: Good question, for vaex v4 it can be inconsistent (dict encoded probably returns the ['foo'], while a 'vaex-our-own-encoding' can return the indices instead. I've checked, and the current code path does seem to go via an arrow array, so this behaves as I want with data_as_labels (I'll push a commit)

maartenbreddels (commented):
Really happy to see this PR, sorry for taking so long to review this.

I'll push two small commits, you only need to accept 1 suggestion (if you agree)

honno:
Forgot to update some refs of `size` (i.e. use `size()` instead), and wrote a test for that. Found a pre-existing bug with `Column.get_chunks()` that now has some xfailed cases... probably something to fix in a future PR I imagine.

Otherwise I think I've addressed your comments now :)

packages/vaex-core/vaex/dataframe_protocol.py
honno: Ah yeah, wrote an issue https://github.com/pandas-dev/pandas/issues/48392

maartenbreddels:
Yeah, Python 3.6 is still being used by clients of a client, so I'm still stuck with that :)
Note that 3.6 has f-strings, just not that `=` syntax in the f-string.

honno:
> > pyarrow.lib.ArrowInvalid: Dictionary array invalid: Invalid: First or last binary offset out of bounds
> 
> This seems like a legit issue :/

I wasn't even scrolling down the failing jobs heh, didn't notice _all_ win jobs were failing. Seems to be a pre-existing problem, looking at the [log for an older CI run](https://github.com/vaexio/vaex/runs/8201187667?check_suite_focus=true), so I'd suggest we leave it for now. Ideally we'd xfail those tests for windows platforms—is there any platform mechanism for that already exists in vaex that I should use, or shall I try come up with something?

maartenbreddels:
Ok, sorry this takes so long, but apart from being very busy, this was difficult to debug.

The issue seems to be that `convert_categorical_column` calls `convert_string_column`, calling `buffer_to_ndarray` which passes the pointer to numpy, but but the nobody has a reference to the `_VaexBuffer` anymore, which releases the reference to the arrow array, which then release the underlying memory.

In this specific case, we get the Arrow buffers via `bitmap_buffer, offsets, string_bytes = self._col.evaluate().buffers()` which returns an arrow array Vaex' doesn't hold the reference to any more (as will be the case for virtual column as well).

So the question is, using the `.ptr` mechanism, who's responsible for holding a reference to the underlying memory?

honno:
Looking at the diff again, this seems like it was an underlying issue that's only come up because this PR handles string labels correctly for categoricals? Mind I'm pretty ignorant when it comes to memory management (seems like a nightmare to debug heh).

So if this isn't an introduced bug and what was working still works, I'd recommend to just xfail the failing test cases if you don't have the bandwidth to fix this right now. I did introduce tests covering string behaviour generally, which should help prevent regressions when attempting to fix this.

maartenbreddels:
Yeah, feel free to xfail that test for now, on all platform

honno:
> Yeah, feel free to xfail that test for now, on all platform

Done!

maartenbreddels (approved):
thanks ! ❤️

Issue #2151, opened 2022-08-04
[BUG-REPORT] df.count error with selections and no limits


arunpersaud:
**Description**
df.count results in an error when called with a list of selections and no limits, but works when limits are given.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`:
   {'vaex': '4.11.1', 'vaex-core': '4.11.1', 'vaex-viz': '0.5.2', 'vaex-hdf5': '0.12.3', 'vaex-server': '0.8.1', 'vaex-astro': '0.9.1', 'vaex-jupyter': '0.8.0', 'vaex-ml': '0.18.0'}
 - Vaex was installed via: pip 
 - OS: OS X 12.4 (M1)

**Additional information**
Here is some example code:

```
import vaex as vx
df = vx.example()

# this works
hist_x = df.count("*", binby="x", shape=1024, selection=None)

# this gives an IndexError (see below
hist_x = df.count("*", binby="x", shape=1024, selection=[None])

# this works again
hist_x = df.count("*", binby="x", shape=1024, selection=[None], limits=(1, 10))

```

The error I'm getting is:
```
>>> hist_x = df.count("*", binby="x", shape=1024, selection=[None])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/dataframe.py", line 965, in count
    return self._compute_agg('count', expression, binby, limits, shape, selection, delay, edges, progress, array_type=array_type)
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/dataframe.py", line 939, in _compute_agg
    return self._delay(delay, progressbar.exit_on(var))
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/dataframe.py", line 1779, in _delay
    return task.get()
  File "/opt/homebrew/lib/python3.9/site-packages/aplus/__init__.py", line 170, in get
    raise self._reason
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/promise.py", line 121, in callAndReject
    ret.fulfill(failure(r))
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/progress.py", line 91, in error
    raise arg
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/promise.py", line 121, in callAndReject
    ret.fulfill(failure(r))
  File "/opt/homebrew/lib/python3.9/site-packages/vaex/delayed.py", line 38, in _wrapped
[...]

PR #2152, opened 2022-08-04
Fix: count with binning using multiple selections

JovanVeljanoski:
Closes: https://github.com/vaexio/vaex/issues/2151

- [x] Implement a test
- [ ] test passes

### DIFF
--- a/tests/count_test.py
+++ b/tests/count_test.py
@@ -23,6 +23,13 @@ def test_count_2d():
     assert list(binned_values.shape) == [32, 32]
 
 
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Ok, had to think about this, but I'm not sure what to do here.

a single aggregation will only be done on a 'fixed grid', so for both selections the same min and max. At some point, we started respecting the selection for the min and max of the binby grid, now I think this was a mistake, because we painted ourselves in a corner. We cannot have two different limits for the two different selections.

maartenbreddels:
@JovanVeljanoski we discussed this privately, do you remember what we decided?

PR #2160, opened 2022-08-07
feature: `get_column_names` accepts a dtypes argument

JovanVeljanoski:
Now we can query column names by data type: 

```python 
import vaex

df = vaex.datasets.titanic()
df.get_column_names(dtypes=[float])

# returns  ['age', 'fare', 'body']
```

Several attempts at this have been done at the past. This is my latest proposal and I think it works quite well. 
Closes:
- https://github.com/vaexio/vaex/issues/240
- https://github.com/vaexio/vaex/pull/251
- https://github.com/vaexio/vaex/pull/1766

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4260,7 +4260,7 @@ class DataFrame(object):
             [k for k in self.variables.keys() if not hidden or not k.startswith('__')] +\
             [k for k in self.functions.keys() if not hidden or not k.startswith('__')]
 
-    def get_column_names(self, virtual=True, strings=True, hidden=False, regex=None):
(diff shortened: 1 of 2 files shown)


maartenbreddels:
See the comments in #1766, mainly
 * dtypes -> dtype
 * Possibly put this in the docs?

JovanVeljanoski:
> See the comments in #1766, mainly
> 
>     * dtypes -> dtype
> 
>     * Possibly put this in the docs?

The idea behind `dtypes` is that you can put more than one, i.e. a list. Or we can indeed change it to `dtype` and be more explicit in the docstring that you can also pass a list. Let's go for that.

maartenbreddels:
Yes, the point is that we always used singular instead of plural.

Issue #2153, opened 2022-08-05
[BUG-REPORT] Filtering breaks negative indexing


karotchykau:
**Description**
The following code
```
import pandas as pd
import vaex

p_df = pd.DataFrame({"A": ["abc"] * 100})
df = vaex.from_pandas(p_df)
f_df = df[df["A"] == "abc"]

f_df[99]  # Works fine.
f_df[-1]  # Throws an error (same for any negative number).
```
throws
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [3], in <cell line: 6>()
      3 f_df = df[df["A"] == "abc"]
      5 f_df[99]  # Works fine.
----> 6 f_df[-1]

File ~/mambaforge/envs/tmp_env/lib/python3.9/site-packages/vaex/dataframe.py:5337, in DataFrame.__getitem__(self, item)
   5335 if isinstance(item, int):
   5336     names = self.get_column_names()
-> 5337     return [self.evaluate(name, item, item+1, array_type='python')[0] for name in names]
   5338 elif isinstance(item, six.string_types):
   5339     if hasattr(self, item) and isinstance(getattr(self, item), Expression):

File ~/mambaforge/envs/tmp_env/lib/python3.9/site-packages/vaex/dataframe.py:5337, in <listcomp>(.0)
   5335 if isinstance(item, int):
   5336     names = self.get_column_names()
-> 5337     return [self.evaluate(name, item, item+1, array_type='python')[0] for name in names]
   5338 elif isinstance(item, six.string_types):
   5339     if hasattr(self, item) and isinstance(getattr(self, item), Expression):

File ~/mambaforge/envs/tmp_env/lib/python3.9/site-packages/vaex/dataframe.py:3090, in DataFrame.evaluate(self, expression, i1, i2, out, selection, filtered, array_type, parallel, chunk_size, progress)
   3088     return self.evaluate_iterator(expression, s1=i1, s2=i2, out=out, selection=selection, filtered=filtered, array_type=array_type, parallel=parallel, chunk_size=chunk_size, progress=progress)
   3089 else:
-> 3090     return self._evaluate_implementation(expression, i1=i1, i2=i2, out=out, selection=selection, filtered=filtered, array_type=array_type, parallel=parallel, chunk_size=chunk_size, progress=progress)

[...]

PR #2163, opened 2022-08-08
Fix: negative index of a filtered dataframe

JovanVeljanoski:
Closes https://github.com/vaexio/vaex/issues/2153

- [x] unit test
- [ ] tests pass


maartenbreddels:
I think you didn't test with master, this was already fixed in #2129

JovanVeljanoski:
I did, and i just double checked it -  it fails on master, i don't know why it does not fail on CI (try in your notebooks)

maartenbreddels:
i guess locally you didn't rebase, but a PR will run against the merged with master.

Issue #2071, opened 2022-06-02
[BUG-REPORT] 'BinnerTime' object has not attribute 'combine_expressions'


thisismeleaving:
**Description**
After upgrading from vaex v4.7.0 This call now throws the following error:
```python
grouped = vx.groupby(by=[vaex.BinnerTime.per_day(vx.dt), vx.tmp_idx])
```

```bash
AttributeError: 'BinnerTime' object has not attribute 'combine_expressions'
```

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.9.1
 - Vaex was installed via: pip 
 - OS: Ubuntu

**Additional information**
Please state any supplementary information or provide additional context for the problem (e.g. screenshots, data, etc..).

thisismeleaving:
This is still an issue in `v4.9.2`

Vicar-of-AI:
Still an issues as of `v4.11.1`

thisismeleaving:
still an issue as of `v4.11.1`

fllara:
I'm experiencing this issue with 4.12

nelmondo:
I'm experiencing this issue with 4.14

JovanVeljanoski:
We've look at this in some detail.. unfortunately it clashes with other features and it might not be fixable (at least not in the short term). @maartenbreddels can explain this better.

As a work-around, i would suggest casting the datatime column to string in the required format, and then do the groupby. Does this make sense? I know it is annoying..

nelmondo:
the same as 4.16

nelmondo:
> We've look at this in some detail.. unfortunately it clashes with other features and it might not be fixable (at least not in the short term). @maartenbreddels can explain this better.
> 
> As a work-around, i would suggest casting the datatime column to string in the required format, and then do the groupby. Does this make sense? I know it is annoying..

i change the format from datetime to string, but I can't solve this issue??? please check this... (I'm using with 4.16.0)

PR #2167, opened 2022-08-12
Fix: multi key groupby when one of the keys is `BinnerTime`

JovanVeljanoski:
Closes: https://github.com/vaexio/vaex/issues/2071

Checklist:
- [x] Make unit-tests
- [ ] Make tests pass


armandgurgu23:
Are there any updates on this bug-fix?

maartenbreddels:
We had some private conversations should this, and it's gonna be tricky to do this in vaex.
Maybe we can think about a workaround @JovanVeljanoski ?

JovanVeljanoski:
Yeah this is discussed the in the parent issue: https://github.com/vaexio/vaex/issues/2071

PR #2176, opened 2022-08-18
CI: Do not run the notebooks on p36

JovanVeljanoski:
We want to run the notebooks in CI on every python version save for 3.6


JovanVeljanoski:
Yeah let;s do that!

Issue #2115, opened 2022-07-11
[BUG-REPORT] mysterious `NameError: name '__filter__' is not defined`


Ben-Epstein:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Please provide a clear and concise description of the problem. This should contain all the steps needed to reproduce the problem. A minimal code example that exposes the problem is very appreciated.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.9.1
 - Vaex was installed via: pip / conda-forge / from source pip
 - OS: ubuntu

**Additional information**
I unfortunately cannot reliably reproduce this error. 

It happens in a fastAPI server when the server is being hit with many requests. Here is the stacktrace (comes from a filter on the data followed by a `df.extract()`

```
  df_copy = df_copy.extract()
  File "/usr/local/lib/python3.9/site-packages/vaex/dataframe.py", line 4468, in extract
    df._push_down_filter()
  File "/usr/local/lib/python3.9/site-packages/vaex/dataframe.py", line 4474, in _push_down_filter
    self._fill_filter_mask()  # make sure the mask is filled
  File "/usr/local/lib/python3.9/site-packages/vaex/dataframe.py", line 5688, in _fill_filter_mask
    self.execute()
  File "/usr/local/lib/python3.9/site-packages/vaex/dataframe.py", line 417, in execute
    self.executor.execute()
  File "/usr/local/lib/python3.9/site-packages/vaex/execution.py", line 308, in execute
    for _ in self.execute_generator():
NameError: name '__filter__' is not defined
```

@maartenbreddels mentioned that 
> I did a benchmark with ab (apache bench) a while ago, triggering a LOT of requests per second, and that triggered a bug. maybe I can trigger this bug as well using the same method

PR #2182, opened 2022-08-23
fix: df.extract() was not thread safe

maartenbreddels:
This caused errors like:
    "NameError: name '__filter__' is not defined"

This happened before we checked in we had a filter, and then removed
the filter in a non-atomic way.


maartenbreddels:
Fixes #2115

PR #2186, opened 2022-08-29
Expose vaex.DataFrame and vaex.Expression for typing

### LINKED ISSUE #2086: [FEATURE-REQUEST]: Expose `vaex.DataFrame` for type hinting
I would like to add type annotations to my code. Currently I can do `from vaex.dataframe import DataFrame`, but it would be nice if  the DataFrame was exposed at the top level so I could just do `import vaex` and then `df: vaex.DataFrame = ...`

I like being able to namespace it like this to differentiate between other DataFrames, eg `import pandas as pd` and then `pdf: pd.DataFrame = ...`

I can submit a PR if you want. Thank you for this awesome library, I'm excited to get started with it! Better string processing is a big requirement of mine and this is 💯

NickCrews:
Closes #2086

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -56,6 +56,8 @@ from .groupby import *
 from . import agg
 import vaex.datasets
 
+# Re-export these so users can type hint with eg vaex.DataFrame
+from vaex.dataframe import DataFrame as DataFrame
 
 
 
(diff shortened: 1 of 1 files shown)


NickCrews:
IDK I didn't figure this required a test?

maartenbreddels:
No, this is fine. I tested this with vscode, and I already love it :)

Issue #2187, opened 2022-08-29
[BUG-REPORT] DataFrame.astype(pyarrow.DictionaryArray(...)) is broken


NickCrews:
**Description**
I was trying to cast a DF to a pyarrow schema and ran across this issue. `DataFrame[column].astype(pyarrow_type)` works fine for many pyarrow types such as float, string, bool. But it doesn't work for `pyarrow.lib.DictionaryType`. However, if I use vaex's "wrapped" version, it works fine.

See the xfail-ing test PR

(PS is there hope for a future `DataFrame.astype()` similar to [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html)? I'm writing this myself and feels like I'm reinventing the wheel.)

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: {'vaex': '4.10.0', 'vaex-core': '4.10.0', 'vaex-viz': '0.5.2', 'vaex-hdf5': '0.12.2', 'vaex-server': '0.8.1', 'vaex-astro': '0.9.1', 'vaex-jupyter': '0.8.0', 'vaex-ml': '0.18.0', 'vaex-graphql': '0.2.0'}
 - Vaex was installed via: both in my application via pip, and on `main` in dev environment
 - OS: MacOS

maartenbreddels:
I like the idea of having a a dtype for the dataframe, we already kind of do checks for that in `DataFrame.__array__`, and we may possibly have some code for this in https://github.com/vaexio/vaex/pull/415 (where we want to know if a dataframe is of a homogenous type, so it can be treated as a 'matrix').

What should `astype(some_dict_type)` do actually?

PR #2188, opened 2022-08-29
Dictionary_casting

NickCrews:
Demonstration of
https://github.com/vaexio/vaex/issues/2187


maartenbreddels:
So, it works for some cases, because our dtype wrapper compared true to `str` when it's an encoded type that maps to str. 
Next problem is, that because of the expression system, we would need to add the types passed in to the variables, because a `repr(arrow_dtype)` cannot be evaluated again.
This would need to use a similar ideas as https://github.com/vaexio/vaex/pull/2089 where we don't add types multiple times.

For the rest, I'll first wait for your response to https://github.com/vaexio/vaex/issues/2187#issuecomment-1232709280

PR #2194, opened 2022-08-30
Extract pre filter test

Ben-Epstein:
@maartenbreddels 

i would expect this to fail on `extract` because there is one pre_filtered task (0) and one not pre_filtered task (1) but this seems to work fine for some reason.


maartenbreddels:
ok, i pushed a failing test, the situation where it failed is.. a filtered dataframe, where 1 of the tasks does pre-filtering (like sum), and the other does not (nop uses mapreduce).

PR #2196, opened 2022-08-31
Fix: histogram sometimes does not obey selections

JovanVeljanoski:
This issue could in partly due to `count` something being buggy when handling arrow data that has missing values. Not sure tho, but there are tests implemented that cover these things in both sides for safety (asserting the histogram output and the `count` output). 

_Workaround:_
If somebody is struggling with this issue, there is a simple workaround. Say column `x` is arrow column with missing values. All you need to do is:
```python
df['x'] = df.x.as_numpy()
```
and everything should work as expected.

_Checklist:_
 - [x] Make unit-tests exposing the issues
 - [x] Make unit-tests pass
 - [x] Party 🎉

### DIFF
--- a/packages/vaex-core/vaex/cpu.py
+++ b/packages/vaex-core/vaex/cpu.py
@@ -11,6 +11,7 @@ import pyarrow as pa
 
 import vaex
 import vaex.encoding
+import vaex.array_types
 from vaex.hash import counter_type_from_dtype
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Party 🎉

JovanVeljanoski:
Party!! 🎉 

Awesome, thank you!

Issue #989, opened 2020-10-01
[BUG-REPORT] add_virtual_columns_eq2gal not working properly


balbinot:
**Description**
add_virtual_columns_eq2gal is not using the correct rotation matrix. An example of the issue can be found here where I use Astropy to check vaex transformations

https://gist.github.com/balbinot/2f13711a747ffc6cf7b32c48dede5ad3

A workaround is to use df.add_virtual_columns_celestial(..., _matrix='eq2gal')

**Software information**
```
'vaex': '3.0.0',
 'vaex-core': '2.0.3',
 'vaex-viz': '0.4.0',
 'vaex-hdf5': '0.6.0',
 'vaex-server': '0.3.1',
 'vaex-astro': '0.7.0',
 'vaex-jupyter': '0.5.2',
 'vaex-ml': '0.9.0',
 'vaex-arrow': '0.5.1'
```
 - Vaex was installed via: pip on anaconda3 2019.10
 - OS: Linux

PR #2206, opened 2022-09-13
Fix Typo in eq2gal

TomCallingham:
Small fix in add_virtual_columns_eq2gal calling the eq2ecl function instead of the eq2gal function. This is an old bug (see #989 https://github.com/vaexio/vaex/issues/989) with a very simple fix.


TomCallingham:
I think the failed tests are down to some connection problems in the test suite, not the fix?

PR #2223, opened 2022-10-01
Docs: Add missing docstrings to `df.export_csv` and `df.export_csv_arrow`

JovanVeljanoski:
The documentation will be updated in a separate PR once we stabilize the methods etc..

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -6959,6 +6959,15 @@ class DataFrameLocal(DataFrame):
 
     @docsubst
     def export_csv(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, backend="pandas", **kwargs):
+        """Exports the DataFrame to a CSV file.
+
+        :param str path: path to the file
+        :param progress: {progress}
+        :param int chunk_size: {chunk_size_export}
+        :param bool parallel: {evaluate_parallel}
+        :param str backend: Which backend to use, either 'pandas' or 'arrow'. Arrow is considerably faster, but pandas is more flexible.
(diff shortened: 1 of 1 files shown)


packages/vaex-core/vaex/dataframe.py
```diff
@@ -6959,6 +6959,15 @@ def export_fits(self, path, progress=None):
...
     @docsubst
     def export_csv(self, path, progress=None, chunk_size=default_chunk_size, parallel=True, backend="pandas", **kwargs):
+        """Exports the DataFrame to a CSV file.
+
+        :param str path: path to the file
+        :param progress: {progress}
+        :param int chunk_size: {chunk_size_export}
+        :param bool parallel: {evaluate_parallel}
+        :param str backend: Which backend to use, either 'pandas' or 'arrow'. Arrow is considerably faster, but pandas is more flexible.
+        :param kwargs: additional keyword arguments are passed to the the backends. See `export_csv_pandas` and `export_csv_arrow` for more details.
```
maartenbreddels, line 6969:
```suggestion
        :param kwargs: additional keyword arguments are passed to the the backends. See :meth:`DataFrame.export_csv_pandas` and :meth:`DataFrame.export_csv_arrow` for more details.
```
I always forget the syntax, but I think this turns it into a link

JovanVeljanoski: Oh I think.. what you suggested turns the formatting in a more fancy way, but from other examples I see we've done `:func:` to get the link.. so I'll try with that (i think the link was your intent)

maartenbreddels:
Links don't work :/ i thought you need to fiddle with this locally

JovanVeljanoski:
I tried many things.. basically copy-pasted the examples that do work and it is still the same

PR #2225, opened 2022-10-10
Docs: update i/o guide for with the new CSV methods backed by arrow.

JovanVeljanoski:


maartenbreddels:
can you rebase this? the version issue is fixed in master

docs/source/guides/io.ipynb
```diff
@@ -346,12 +346,64 @@
...
-    "Datasets are still commonly stored in text-based file formats such as CSV. Since text-based file formats are not memory-mappable, they have to be read in memory. If the contents of a CSV file fits into the available RAM, one can simply do:"
+    "Datasets are still commonly stored in text-based file formats such as CSV and JSON. Vaex supports various methods for reading such datasets.\n",
+    "\n",
+    "<div class=\"alert alert-block alert-info\"><b>New in 4.14: </b> \n",
+    "\n",
+    "The `vaex.open` method can also be used to read a CSV file. With this method Vaex will *lazily* read the CSV file, i.e. the data from the CSV file will be streamed when computations need to be executed. This is powered by Apache Arrow under the hood. In this way you can work with arbitraruly large CSV files without caring about RAM!\n",
+    "\n",
+    "Note: When opening a CSV file in this way, Vaex will first quickly scan the file to determine some basic metadata such as the number of rows, column names and their data types. The duration of this can vary depending on the number of rows, columns, your disk read speed, and settings telling Vaex thoroughly it should scan the CSV file. \n",
```
maartenbreddels, line 355: Should we at mention the `schema_infer_fraction` here?

maartenbreddels (commented):
nice work! I do think we should hint about schema_infer_fraction, what do you think?

docs/source/guides/io.ipynb
JovanVeljanoski: Made some changes, now it is explicitly mentioned. What do you think?

docs/source/faq.ipynb
```diff
@@ -40,13 +40,37 @@
...
+    "\n",
+    "<div class=\"alert alert-block alert-warning\"><b>Prior to 4.14: </b>\n",
     "\n",
     "```\n",
     "df = vaex.from_csv('./my_data/my_big_file.csv', convert=True, chunk_size=5_000_000)\n",
     "```\n",
     "\n",
     "When the above line is executed, Vaex will read the CSV in chunks, and convert each chunk to a temporary HDF5 file on disk. All temporary will files are then concatenated into a single HDF5, and the temporary files deleted. The size of the individual chunks to be read can be specified via the `chunk_size` argument. \n",
+    "    \n",
+    "Note: this is still possible in the newer version of Vaex, but it not the most performant approach.\n",
```
Ben-Epstein, line 71:
```suggestion
    "Note: this is still possible in the newer version of Vaex, but it is not the most performant approach.\n",
```

JovanVeljanoski: thanks!

Issue #2211, opened 2022-09-20
[FEATURE-REQUEST] Support Google Colab installation without runtime restart


franz101:
**Description**
When running `pip install vaex`, the runtime of the colab is needed to restarted (see attachments)

**Is your feature request related to a problem? Please describe.**
The problem is some dependencies update matplotlib to larger then 3.22 (which runs on Google Colab)
**Additional context**
[Here is a Google Colab that shows the problem and a workaround](https://colab.research.google.com/drive/19O8H2n947R4g5PEMKrOHvIb7pvec3VPg?usp=sharing)

Example of runtime restart needed:
<img width="1197" alt="Screenshot 2022-09-20 at 15 27 03" src="https://user-images.githubusercontent.com/18228395/191348197-88ec9c97-d03f-4eb5-b612-2aa3522aa1c7.png">

Example of installing the packages in a way matplotlib is not updated:
<img width="1547" alt="Screenshot 2022-09-20 at 15 27 56" src="https://user-images.githubusercontent.com/18228395/191348378-4da6ee8a-1eb5-4b33-b920-6df407a55052.png">

franz101:
@maartenbreddels

JovanVeljanoski:
I don't think we can do anything about that. It is their environment. 

If somebody has an idea on how to fix this - please get in touch.

maartenbreddels:
Strange issue, thanks for letting us know.

Maybe by using the solution here https://github.com/geoalchemy/geoalchemy2/pull/392 in vaex (instead of the slow pkg_resources) we can make this faster and also working on colab without requiring a restart.

I think this could be a good first issue for someone.

franz101:
Hey @maartenbreddels ,

thanks for the input. I have forked vaex and am experimenting with the different version implementation:
https://github.com/franz101/vaex-colab/blob/f70f4a4e2ef6c4d2902239e183f5d7abd4bcc4d3/packages/vaex-core/vaex/version.py

But running the installation it fails during the vaex-meta setup:
https://colab.research.google.com/drive/1WC-ROEunPKaxHD50b_vqxHS_3FT9METM#scrollTo=apSprOWRQ2qy

is this related to python 3.7 and not a venv?

franz101:
Running setup.py develop for vaex-meta
[...]

PR #2226, opened 2022-10-11
[FEATURE-REQUEST] Import and version checking improvements  #2211

franz101:
Changing to importlib according to this issue:
https://github.com/vaexio/vaex/issues/2211

I was not able to build vex locally yet. Curios about feedback

### DIFF
--- a/packages/vaex-core/vaex/version.py
+++ b/packages/vaex-core/vaex/version.py
@@ -4,6 +4,7 @@ packages = ['vaex', 'vaex-core', 'vaex-viz', 'vaex-hdf5', 'vaex-server', 'vaex-a
 
 def get_versions():
     def is_installed(p):
+        #  Concept from https://github.com/geoalchemy/geoalchemy2/pull/392/files#diff-208c493ee84f787a0a966b235a31f86a5d382ff76e1e5533b5b1da712996b7e2
         try:
             import importlib.metadata
         except ImportError:
--- a/packages/vaex-core/vaex/version.py
+++ b/packages/vaex-core/vaex/version.py
@@ -1,6 +1,3 @@
-import pkg_resources
-
-
 packages = ['vaex', 'vaex-core', 'vaex-viz', 'vaex-hdf5', 'vaex-server', 'vaex-astro',
             'vaex-ui', 'vaex-jupyter', 'vaex-ml', 'vaex-arrow', 'vaex-graphql', 'vaex-enterprise']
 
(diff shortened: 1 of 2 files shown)


packages/vaex-core/vaex/version.py
```diff
@@ -1,15 +1,38 @@
...
     def is_installed(p):
+        #  Concept from https://github.com/geoalchemy/geoalchemy2/pull/392/files#diff-208c493ee84f787a0a966b235a31f86a5d382ff76e1e5533b5b1da712996b7e2
+        try:
+            import importlib.metadata
+        except ImportError:
+            try:
+                from pkg_resources import DistributionNotFound
+                from pkg_resources import get_distribution
+            except ImportError:  # pragma: no cover
+                pass
```
maartenbreddels, line 14: why would this import fail?

packages/vaex-core/vaex/version.py
```diff
@@ -1,15 +1,38 @@
...
+            else:
+                try:
+                    return get_distribution(p).version
+                except DistributionNotFound:  # pragma: no cover
+                    raise Exception(f"Package not found: {p}")
```
maartenbreddels, line 19: Isn't this much simpler?
```suggestion
                    return False
```

packages/vaex-core/vaex/version.py
```diff
@@ -1,15 +1,38 @@
...
+        else:
+            try:
+                return importlib.metadata.version(p)
+
+            except importlib.metadata.PackageNotFoundError:  # pragma: no cover
+                raise Exception(f"Package not found: {p}")
```
maartenbreddels, line 25:
```suggestion
                return False
```

maartenbreddels (changes requested):
Hi,

thanks for you contribution!
I do think the code can use a cleanup, the `is_installed` can simply return a boolean, no need to throw exceptions.
Or, call the function `get_version()`, but I don't like to do string comparisons on exception handling.
I think the cleanest approach is a `is_installed` and `get_version`.
Does that make sense?

Regards,

Maarten

franz101:
I agree with your changes. Currently I found another issue with colab. When you install a package that changes a subdependency it requires  restart. but if you just run the versions command it's fine SMH:
https://colab.research.google.com/drive/1yu0Fkszz7UR5zLYZdWUt_v1net5ewVvQ?usp=sharing

packages/vaex-core/vaex/version.py
franz101: removed

packages/vaex-core/vaex/version.py
franz101: changed the logic to just save the version

franz101:
@maartenbreddels please have a look at the new implementation

maartenbreddels (approved):
Yeah, i like this a lot!

PR #2245, opened 2022-10-25
Fix: export to hdf5 when there are numeric missing values

JovanVeljanoski:
When we have pyarrow numeric data with missing values, the missing data are not correctly exported to hdf5.

- [x] Make unit-test
- [x] Make test pass
- [x] Party


JovanVeljanoski:
Awesome! 🎉

PR #2254, opened 2022-11-01
Update fits.py

balbinot:
Fix bug while plotting and manipulating df coming from FITS files

### DIFF
--- a/packages/vaex-astro/vaex/astro/fits.py
+++ b/packages/vaex-astro/vaex/astro/fits.py
@@ -18,8 +18,8 @@ logger = logging.getLogger("vaex.astro.fits")
 
 
 class FitsBinTable(DatasetMemoryMapped):
+    snake_name='fits'
     def __init__(self, filename, write=False, fs_options={}, fs=None):
-	snake_name='fits'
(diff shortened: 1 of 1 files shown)


packages/vaex-astro/vaex/astro/fits.py
```diff
@@ -19,6 +19,7 @@
 
 class FitsBinTable(DatasetMemoryMapped):
     def __init__(self, filename, write=False, fs_options={}, fs=None):
+	snake_name='fits'
```
maartenbreddels, line 22:
```suggestion
    snake_name='fits'
    def __init__(self, filename, write=False, fs_options={}, fs=None):
```
The snake name should be at the class level

maartenbreddels: I think it also needs the @vaex.dataset.register decorator.

balbinot:
It seems to work well with or without the decorator.

The missing decorator is also in the Gadget class, which I am not sure it is used still. 

https://github.com/vaexio/vaex/commit/b171bcdce68e72589996f2af568a4bb0374b49c9

maartenbreddels:
Ok, this is good enough for now, thanks!

PR #2233, opened 2022-10-17
Fix: `extract()` on an empty dataframe due to a filter

JovanVeljanoski:
Fixes https://github.com/vaexio/vaex/issues/2232

- [x] Test exposing an issue
- [ ] Making the test pass

Issue #2232, opened 2022-10-17
[BUG-REPORT] to_pandas_df throws exception on filtered data frame on virtual column, with empty result


vladmihaisima:
**Description**
to_pandas_df throws exception when invoked on empty filter. This happens only when there is a virtual column with a transformation (it will not happen if col1 is not modified see code below)

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: {'vaex': '4.12.0', 'vaex-core': '4.12.0', 'vaex-viz': '0.5.3', 'vaex-hdf5': '0.12.3', 'vaex-server': '0.8.1', 'vaex-astro': '0.9.1', 'vaex-jupyter': '0.8.0', 'vaex-ml': '0.18.0'}
 - Vaex was installed via: pip 
 - OS: Arch (similar behavior on Ubuntu)

**Additional information**
Code to reproduce. First two prints work, third one throws exception.
```
import pandas as pd
import vaex as vx

df = vx.from_pandas(pd.DataFrame(data={'col1':['chr1','chr2'],'col2':[3,4]}))
print(df[df['col1']=='3'].extract().to_pandas_df()) # Works
df['col1'] = df.col1.astype('str').str.replace('^chr','',regex=True)
print(df[df['col1']=='3'].extract()) # Works
print(df[df['col1']=='3'].extract().to_pandas_df()) # Throws exception
```
results in: AssertionError for assert filter.sum() == expected_length, in dataset.py, line 1018, in slice

JovanVeljanoski:
Thanks for the report. This seems to be an edge case. I needed to convince myself that it is a bug. 

Do you understand why it happens?

Basically, you do not need to call `.extract()` if you plan to export the data to pandas or any other in-memory representation. So if you remove the `.extract()` part of your examples, things will work as expected. 

Still we need to address this. 
I wonder whether doing a `.extract()` on an empty dataframe makes sense and whether it should be allowed at all..? 
@maartenbreddels

maartenbreddels:
This is difficult to fix, unless we refactor how we handle the schema. It is implicit now, and on demand. I think if we passed down the schema we could fix this, but I'm not sure what that would do.
If would say, the workaround for now is
```python
df = df.extract() if len(df) else df
```

maartenbreddels:
FYI, we will raise an error in the future: https://github.com/vaexio/vaex/pull/2267

Ben-Epstein:
[...]

PR #2267, opened 2022-11-16
feat: raise when extract cannot run, instead of failing later.

maartenbreddels:
Alternative to #2233
Adresses: #2232


JovanVeljanoski:
I think we can potentially revisit this in v5 if a larger refactor is possible (to propagate the schema etc.). But for not I'd say it is good enough.

JovanVeljanoski (approved):

Issue #2272, opened 2022-11-18
[FEATURE-REQUEST] Aggregate dict on a groupby


vikalfc:
When  #2032 is done, it would be useful to have something like this

```python
import vaex

df = vaex.datasets.titanic()

def custom_f(x,y):
    return {"name": x,"cabin": y}

df["content"]=df.apply(custom_f,
    arguments=[df["name"], df["cabin"]])
df_pandas=df.to_pandas_df()

df_pandas.groupby(by=['pclass', 'survived']).agg({
    'content': list}).reset_index()

```
<img width="481" alt="Screenshot 2022-11-18 at 12 28 13" src="https://user-images.githubusercontent.com/23177147/202695292-1f9867b1-26c5-49ca-bf32-d5fd31a33aae.png">

PR #2273, opened 2022-11-18
Fix: list aggregator in 

JovanVeljanoski:
Discovered by and partially addresses https://github.com/vaexio/vaex/issues/2272

- [x] Unit-test
- [ ] Unit-tests passes


maartenbreddels:
ok, maybe we should always assume sparse when having arrow like aggregators, that would be 1 solution. The other is that I find a way to flatten the arrays before... which is hard :)

PR #2284, opened 2022-11-29
Better handle empty extract

Ben-Epstein:
I think this is a better way to handle extracting an empty dataframe than https://github.com/vaexio/vaex/pull/2267

relates to https://github.com/vaexio/vaex/issues/2232


maartenbreddels:
Thanks Ben!

Issue #2287, opened 2022-11-30
[BUG-REPORT] .sort() crashes on DFs filtered down to 0 rows


NickCrews:
**Description**

```python
import vaex

df = vaex.from_arrays(x=[0,1])
df = df[df.x > 0]  # no error as long as at least one row remains
df = df[df.x > 10]  # error if all rows filtered out
df.sort("x")
```

I get

```Python traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/Users/nickcrews/Documents/projects/scg/alatna/.scratch/link.ipynb Cell 20 in <cell line: 6>()
      [4](vscode-notebook-cell:/Users/nickcrews/Documents/projects/scg/alatna/.scratch/link.ipynb#Y245sZmlsZQ%3D%3D?line=3) df = df[df.x > 0] # works
      [5](vscode-notebook-cell:/Users/nickcrews/Documents/projects/scg/alatna/.scratch/link.ipynb#Y245sZmlsZQ%3D%3D?line=4) df = df[df.x > 10] # does not work
----> [6](vscode-notebook-cell:/Users/nickcrews/Documents/projects/scg/alatna/.scratch/link.ipynb#Y245sZmlsZQ%3D%3D?line=5) df.sort("x")

File ~/Library/Application Support/hatch/env/virtual/alatna-U4QVXPcf/alatna/lib/python3.9/site-packages/vaex/dataframe.py:4727, in DataFrame.sort(self, by, ascending)
   4725 # if we don't cast to int64, we get uint64 scalars, which when adding numbers to will auto case to float (numpy)
   4726 indices = vaex.array_types.to_numpy(indices).astype('int64')
-> 4727 return self.take(indices)

File ~/Library/Application Support/hatch/env/virtual/alatna-U4QVXPcf/alatna/lib/python3.9/site-packages/vaex/dataframe.py:4458, in DataFrame.take(self, indices, filtered, dropfilter)
   4454 if df.filtered and filtered:
   4455     # we translate the indices that refer to filters row indices to
   4456     # indices of the unfiltered row indices
   4457     df._fill_filter_mask()
-> 4458     max_index = indices.max()
   4459     mask = df._selection_masks[FILTER_SELECTION_NAME]
   4460     filtered_indices = mask.first(max_index+1)

[...]

PR #2290, opened 2022-12-01
Fix: sorting of an empty df returns an empty df with a warning

JovanVeljanoski:
Addresses https://github.com/vaexio/vaex/issues/2287

@maartenbreddels I wrapped this up in `try/expect` block as to now check for length every time a df is sorted. If you think this should be done the other way around (i.e. first check length) then try to sort, lemme know!

Thanks to @NickCrews for bringing this up.

- [x] Unit test
- [x] Tests pass
- [ ] Review

### DIFF
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4709,7 +4709,6 @@ class DataFrame(object):
         '''
 
         if len(self) == 0:
-            warnings.warn('Cannot sort empty DataFrame, returning empty DataFrame.')
             return self.copy()
         self = self.trim()
         # Ensure "by" is in the proper format
--- a/packages/vaex-core/vaex/dataframe.py
+++ b/packages/vaex-core/vaex/dataframe.py
@@ -4707,24 +4707,32 @@ class DataFrame(object):
         :param str or expression or list of str/expressions by: expression to sort by.
         :param bool or list of bools ascending: ascending (default, True) or descending (False).
         '''
-        self = self.trim()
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Yeah, if think we can simply do, `if len(df) == 0: return df.copy()` what do you think about that solution?
This solution can accidentally do more .

JovanVeljanoski:
yeah sounds good, i just didn't want to check length each time without needing to.. but if you think that is not too muc overhead (for heavily filtered dfs) I am happy, and the code will be cleaner

maartenbreddels:
It might trigger a compute if the filter is not yet computed, but that will happen anyway for the sort.

packages/vaex-core/vaex/dataframe.py
```diff
@@ -4707,6 +4707,10 @@ def sort(self, by, ascending=True):
         :param str or expression or list of str/expressions by: expression to sort by.
         :param bool or list of bools ascending: ascending (default, True) or descending (False).
         '''
+
+        if len(self) == 0:
+            warnings.warn('Cannot sort empty DataFrame, returning empty DataFrame.')
```
maartenbreddels, line 4712:
```suggestion
```
Do we need to warn here? I think we can just leave it out.

JovanVeljanoski: Done in new commit (needed to update the test as well, i was checking for the warning.. )

NickCrews: I would +1 to there being no warning.

Issue #2283, opened 2022-11-29
[BUG-REPORT] Fresh install of vaex cannot open files written to disk


Ben-Epstein:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**

On a fresh notebook without vaex installed, run the following 
```
!pip install vaex-core==4.15.0 vaex-hdf5==0.12.3

import vaex
import numpy as np

df = vaex.from_arrays(id=list(range(100_000)), emb=np.random.rand(100_000, 768))
df.export('file.hdf5')
df.export('file1.hdf5')

vaex.open("file*.hdf5")
```

You see the following

<img width="782" alt="image" src="https://user-images.githubusercontent.com/22605641/204547272-1988c2e0-75e1-4ef8-a353-b116f9557484.png">

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: {'vaex-core': '4.15.0', 'vaex-hdf5': '0.12.3'}
 - Vaex was installed via: pip / conda-forge / from source 
 - OS:

**Additional information**
Please state any supplementary information or provide additional context for the problem (e.g. screenshots, data, etc..).

franz101:
Reproducing the error it seems to me related to `pip install vaex ` vs `pip install vaex-core`

franz101:
Maybe related:
https://medium.com/google-colab/colab-updated-to-python-3-8-4922f9970a72

franz101:
For reproduction
https://colab.research.google.com/drive/1EG9898VtmO19FwfZKd_LJzqkz_YVWwlE?usp=sharing

pip freeze diff check shows the packages that additionally get installed:
```python
aplus==0.11.0
blake3==0.3.1
commonmark==0.9.1
frozendict==2.3.4
nest-asyncio==1.5.6
rich==12.6.0
```

franz101:
Found the issue:
in file `dataset.py` line `57`
`for entry in pkg_resources.iter_entry_points(group='vaex.dataset.opener')`

is empty except if you restart the kernel

franz101:
here is a fix

JovanVeljanoski:
@Ben-Epstein is that a bug for a specific version of the vaex-hdf5 package? what happens if you update to 0.13 or 0.14?

Ben-Epstein:
[...]

PR #2293, opened 2022-12-01
Remove pkg_resources from direct imports

franz101:
This is a hot fix for the IO read issues caused by using the legacy pkg_resources.iter_entry_points

legacy `pkg_resources.iter_entry_points` is replaced with `importlib.metadata.entry_points`
```python
try:
  if sys.version_info[:2] >= (3, 10):
      # pylint: disable=no-name-in-module
      from importlib.metadata import entry_points
  else:
      from importlib_metadata import entry_points
except ImportError:
  import pkg_resources
  entry_points = pkg_resources.iter_entry_points
```

### DIFF
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -66,7 +66,9 @@ try:
     if version_info[:2] >= (3, 10):
         from importlib.metadata import entry_points
     else:
-        from importlib_metadata import entry_points
+        from importlib_metadata import entry_points, __version__ as importlib_metadata_version
+        if int(importlib_metadata_version.split(".")[0]) < 4.0:
+            raise ImportError("vaex requires importlib_metadata >= 4.0 when installed")
 except ImportError:
     import pkg_resources
     entry_points = pkg_resources.iter_entry_points
--- a/packages/vaex-core/vaex/__init__.py
+++ b/packages/vaex-core/vaex/__init__.py
@@ -801,7 +801,11 @@ for entry in entry_points(group='vaex.expression.accessor'):
 
 
 for entry in entry_points(group='vaex.plugin'):
-    if entry.module_name == 'vaex_arrow.opener':
+    try:
+        module_name = entry.module 
+    except AttributeError:
+        module_name = entry.module_name
+    if module_name == 'vaex_arrow.opener':
         # if vaex_arrow package is installed, we ignore it
         continue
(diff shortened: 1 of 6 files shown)


franz101:
The discussion of the issue can be found here:
https://github.com/vaexio/vaex/issues/2283

franz101:
@JovanVeljanoski @maartenbreddels 

This should fix the reading file error:
https://colab.research.google.com/drive/1EG9898VtmO19FwfZKd_LJzqkz_YVWwlE?usp=sharing#scrollTo=S-T0Wjd3B_YU

Screenshot 2022-12-01 at 17 10 29


The reason is due to pkg_resources unable to find openers without kernel restart:
https://git.smhi.se/climix/climix/-/merge_requests/165/diffs

maartenbreddels:
Ok, just to make it clear I understand. 
 * pkg_resources.iter_entry_points is slow
 * Python 3.8 has `importlib.metadata` which is faster
 * If on Python < 3.8 and importlib_metadata is installed (it's not a dependency right now) you also get a speedup by not using pkg_resources

maartenbreddels:
Man, I can't believe this is all so flakey, `importlib_metadata==4.0` seems to work, 3.0 doesn't seem to have the groups argument.

packages/vaex-core/vaex/__init__.py
```diff
@@ -60,10 +59,18 @@
...
-
-
 import vaex.progress
 
+try:
+    from sys import version_info
+    if version_info[:2] >= (3, 10):
+        from importlib.metadata import entry_points
+    else:
+        from importlib_metadata import entry_points
```
maartenbreddels, line 69:
```suggestion
        from importlib_metadata import entry_points, __version__ as importlib_metadata_version
        if int(importlib_metadata_version.split(".")[0]) < 4.0:
            raise ImportError("vaex requires importlib_metadata >= 4.0 when installed")
```
should we do this?

franz101:
Actually it is not about the speed here. In Google colab after a fresh installation without a restart, pkg_resources entrypoints are an empty list. Therefore it won't even try any openers and fail.

I agree can't believe how bad python handles lazy loads and has API changes across releases.

maartenbreddels:
Many thanks for helping us sort out this mess! :)

franz101:
its my first large open source project contribution. happy to contribute where I can. ❤️

Issue #2334, opened 2023-02-09
[BUG-REPORT] Potential memory leak when exporting large strings to hdf5


Ben-Epstein:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Please provide a clear and concise description of the problem. This should contain all the steps needed to reproduce the problem. A minimal code example that exposes the problem is very appreciated.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.16.1
 - Vaex was installed via: pip / conda-forge / from source pip
 - OS: Linux (colab)

**Additional information**
If you run this on a limited machine like google colab free, you will get a OOM crash when exporting to hdf5, even though it works fine exporting to arrow. We need to convert the string to a large_string because of pyarrow issues https://issues.apache.org/jira/browse/ARROW-17828

```
import vaex
import pyarrow as pa

df = vaex.example()
df["text"] = vaex.vconstant("OHYEA"*10000, len(df))

@vaex.register_function()
def to_large(arr):
    return arr.cast(pa.large_string())

df["text"] = df["text"].to_large()
#OOM
df.export("file.hdf5")
```

Ben-Epstein:
You should be able to reproduce it here https://colab.research.google.com/drive/1J085UZolLNcaL8zhVKY0LQzbgFMXnYur?usp=sharing

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/22605641/217858040-984f9304-c02b-4b14-889d-01304e156b91.png">

Ben-Epstein:
(Also in the notebook above) When you export to arrow, it works fine, takes about 1 minute, and the memory stays very low 
<img width="1460" alt="image" src="https://user-images.githubusercontent.com/22605641/217858795-822f45d4-229c-44e4-b219-994612eb6aaf.png">

Issue #2335, opened 2023-02-09
[BUG-REPORT] Offset Overflow when joining dataframes with large strings


Ben-Epstein:
Thank you for reaching out and helping us improve Vaex!

Before you submit a new Issue, please read through the [documentation](https://docs.vaex.io/en/latest/). Also, make sure you search through the Open and Closed Issues - your problem may already be discussed or addressed.

**Description**
Please provide a clear and concise description of the problem. This should contain all the steps needed to reproduce the problem. A minimal code example that exposes the problem is very appreciated.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.16.1
 - Vaex was installed via: pip / conda-forge / from source  pip
 - OS: mac/linux

**Additional information**
This likely has to do with both https://issues.apache.org/jira/browse/ARROW-17828 and is probably related to https://github.com/vaexio/vaex/issues/2334

### Part 1
We create a dataframe where the text column is larger than the max size for a string. We write it to an arrow file, and that's fine. **Note here that if we wrote this to an hdf5 file without converting to a large string, we'd have an error... see [2334](https://github.com/vaexio/vaex/issues/2334)**
```python
import vaex
import numpy as np

# from https://issues.apache.org/jira/browse/ARROW-17828
# arrow string maxsize is 2GB. Then you need large string
maxsize_string = 2e9

df = vaex.example()
df["text"] = vaex.vconstant("OHYEA"*2000, len(df))
df["id"] = np.array(list(range(len(df))))

assert df.text.str.len().sum() > maxsize_string

df[["x", "y", "z", "text", "id"]].export("part1.arrow")
```
### Part 2
If we join this dataframe with another, we see that there are no issues
```python
# Works fine
newdf = vaex.example()[["id"]]
newdf["col1"] = vaex.vconstant("foo", len(newdf))

join_df = newdf.join(df, on="id")
join_df
```
<img width="1058" alt="image" src="https://user-images.githubusercontent.com/22605641/217871072-c0ea87ea-0094-402a-9e99-5abcd04c746b.png">

### Part 3 (issue)
BUT, if we instead read that file from disk where it's memory mapped, joining to another dataframe causes an issue
```python
# Reading from disk will break the join
newdf2 = vaex.example()[["id"]]
newdf3 = vaex.open("part1.arrow")

newdf2["col1"] = vaex.vconstant("foo", len(newdf))

join_df2 = newdf2.join(newdf3, on="id")
join_df2
```
[...]

PR #2336, opened 2023-02-09
dont use take with arrow

Ben-Epstein:
This directly addresses https://github.com/vaexio/vaex/issues/2335
And is directly the fix for https://issues.apache.org/jira/browse/ARROW-9773

I believe fixing all of the `.take`s to `.slice` would also fix https://github.com/vaexio/vaex/issues/2334 because `.take` uses memory, but `.slice` is zero-copy. The memory is exploding going to hdf5 because we keep `.take`ing 

You can see huggingface datasets does the same thing: https://github.com/huggingface/datasets/pull/645/files

That being said, there are a number of other places that vaex uses `.take` which should be fixed. But because of the lack of typing in the vaex repo, it's hard for me to know which ones are pyarrow arrays, which are pyarrow tables, and which are numpy arrays. I'm happy to help move the rest over, but I would need some guidance.

Here are all of the places `.take` is used
```
./packages/vaex-core/vaex/functions.py:    values = choices.take(indices)
./packages/vaex-core/vaex/dataframe.py:                    used_keys = keys.take(codes)
./packages/vaex-core/vaex/dataframe.py:                        keys = keys.take(indices)
./packages/vaex-core/vaex/dataframe.py:    def take(self, indices, filtered=True, dropfilter=True):
./packages/vaex-core/vaex/dataframe.py:        >>> df.take([0,2])
./packages/vaex-core/vaex/dataframe.py:        df.dataset = df.dataset.take(indices)
./packages/vaex-core/vaex/dataframe.py:        return self.take(indices)
./packages/vaex-core/vaex/dataframe.py:        return self.take(indices).split(into)
./packages/vaex-core/vaex/dataframe.py:        return self.take(indices)
./packages/vaex-core/vaex/cpu.py:                keys = keys.take(indices)
./packages/vaex-core/vaex/cpu.py:        keys = keys.take(order)
./packages/vaex-core/vaex/join.py:            left = left.concat(left.take(lookup_left))
./packages/vaex-core/vaex/join.py:            left = left.take(left_indices_matched, filtered=False, dropfilter=False)
./packages/vaex-core/vaex/join.py:            right_dataset = right_dataset.take(lookup, masked=any(lookup_masked))
./packages/vaex-core/vaex/dataset.py:    def take(self, indices, masked=False):
./packages/vaex-core/vaex/dataset.py:        return DatasetTake(self, indices, masked=masked)
./packages/vaex-core/vaex/dataset.py:class DatasetTake(DatasetDecorator):
./packages/vaex-core/vaex/hash.py:        keys = pa.compute.take(keys, indices)
./packages/vaex-core/vaex/array_types.py:def take(ar, indices):
./packages/vaex-core/vaex/array_types.py:    return ar.take(indices)
./packages/vaex-core/vaex/groupby.py:                        self.bin_values = pa.compute.take(self.bin_values, self.sort_indices)
./packages/vaex-core/vaex/groupby.py:                        bin_values[field.name] = ar.take(indices)
./packages/vaex-core/vaex/groupby.py:                    bin_values[parent.label] = parent.bin_values.take(indices)
./packages/vaex-core/vaex/groupby.py:            self.bin_values = pa.compute.take(self.bin_values, sort_indices)
./packages/vaex-core/vaex/groupby.py:            values = pa.compute.take(values, indices)
./packages/vaex-core/vaex/groupby.py:                        ar = np.take(ar, sort_indices, axis=i)
./packages/vaex-core/vaex/groupby.py:                            columns[by.label] = vaex.array_types.take(by.bin_values, index)
./packages/vaex-core/vaex/groupby.py:                        columns[by.label] = by.bin_values.take(indices)
./packages/vaex-core/vaex/column.py:            ar = ar_unfiltered.take(vaex.array_types.to_arrow(take_indices))
./packages/vaex-core/vaex/column.py:        x = x.dictionary.take(x.indices)  # equivalent to PyArrow 5.0.0's dictionary_decode() but backwards compatible
```

### DIFF
--- a/packages/vaex-core/vaex/column.py
+++ b/packages/vaex-core/vaex/column.py
@@ -382,8 +382,17 @@ class ColumnIndexed(Column):
             # arrow and numpy do not like the negative indices, so we set them to 0
             take_indices = indices.copy()
             take_indices[mask] = 0
-        if isinstance(ar_unfiltered, supported_arrow_array_types):
-            ar = ar_unfiltered.take(vaex.array_types.to_arrow(take_indices))
+        # Don't use .take in arrow anymore
+        # https://issues.apache.org/jira/browse/ARROW-9773
+        # slice is zero-copy
+        if isinstance(ar_unfiltered, pa.Array):
+            ar = pa.concat_arrays(
+                [ar_unfiltered.slice(i, 1) for i in take_indices]
+            )
+        elif isinstance(ar_unfiltered, pa.ChunkedArray):
+            ar = pa.concat_arrays(
+                [ar_unfiltered.slice(i, 1).combine_chunks() for i in take_indices]
+            )
         else:
             ar = ar_unfiltered[take_indices]
         assert not np.ma.isMaskedArray(indices)


anthonycorletti (approved):

maartenbreddels:
> The memory is exploding going to hdf5 because we keep `.take`ing

Yeah, when using sliced arrays, that seems to be the case. It will try to concatenate them first, which will explode the memory use!

This is quite a bad arrow situation... :/ digesting this

packages/vaex-core/vaex/column.py
```diff
@@ -382,8 +382,17 @@ def __getitem__(self, slice):
...
+        # https://issues.apache.org/jira/browse/ARROW-9773
+        # slice is zero-copy
+        if isinstance(ar_unfiltered, pa.Array):
+            ar = pa.concat_arrays(
+                [ar_unfiltered.slice(i, 1) for i in take_indices]
+            )
+        elif isinstance(ar_unfiltered, pa.ChunkedArray):
+            ar = pa.concat_arrays(
+                [ar_unfiltered.slice(i, 1).combine_chunks() for i in take_indices]
+            )
```
maartenbreddels, line 395:
```suggestion
        if isinstance(ar_unfiltered, pa.Array):
            ar = ar_unfiltered.take(vaex.array_types.to_arrow(take_indices))
        elif isinstance(ar_unfiltered, pa.ChunkedArray):
            # Don't use .take in arrow for chunked arrays 
            # https://issues.apache.org/jira/browse/ARROW-9773
            # slice is zero-copy
            ar = pa.concat_arrays(
                [ar_unfiltered.slice(i, 1).combine_chunks() for i in take_indices]
            )
```
it is only an issue for chunked arrays right?
Maybe the best place to put this is vaex.array_types.take ?
There is already a take function there, and if all places in vaex uses that function, we have a nice central place where we can solve this.

maartenbreddels (commented):
Thanks for looking into this Ben, please take a look at my comment.

packages/vaex-core/vaex/column.py
Ben-Epstein: @maartenbreddels i do not think this is only an issue for chunked_arrays, I think it's an issue with `take` broadly. I definitely like the idea of moving it down to a centralized place, but I think no matter the array type, you need to replace `take` with `slice`

maartenbreddels: https://issues.apache.org/jira/browse/ARROW-9773 only talks about chunked arrays, and I also think it's only a problem for that (since it's more complex).

Ben-Epstein: Okay i'll take your word on this one, you certainly know more here. I'll move this into the vaex `array_types.take`

Ben-Epstein:
@maartenbreddels i updated the function to use the array_types `take` function, but i wasn't able to update the others. `take` is used in a bunch of places and they aren't always from a pyarrow array. Without typing, it's not clear to me when it's on a dataframe, or a dataset, or something else.

I tried to dig into the dataframe `take` which led to the dataset `take` and then this `DatasetTake` [class](https://github.com/Ben-Epstein/vaex/blob/6b363a11b295c5d91a25d48b7312d2bf2bac151b/packages/vaex-core/vaex/dataset.py#L879) but I can't really understand what it's doing. Would you mind helping me out with that part?

maartenbreddels:
I'll scan over it tomorrow!

PR #2339, opened 2023-02-14
ci: fix CI, use macOS-11 and remove pydantic

maartenbreddels:
fixing CI rot

### DIFF
--- a/packages/vaex-server/vaex/server/fastapi.py
+++ b/packages/vaex-server/vaex/server/fastapi.py
@@ -44,8 +44,8 @@ class HistogramInput(BaseModel):
     shape: int = 128
     min: Optional[Union[float, int, str]] = None
     max: Optional[Union[float, int, str]] = None
-    filter: str = None
-    virtual_columns: Dict[str, str] = None
+    filter: Optional[str] = None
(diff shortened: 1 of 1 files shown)


ddelange:
@maartenbreddels any hunches regarding the failing random_projections tests?

`TypeError: '>' not supported between instances of 'NoneType' and 'int'`

if I can support, let me know :)

maartenbreddels:
I was actually surprised it never triggered.

maartenbreddels:
https://github.com/vaexio/vaex/pull/2339/commits/5da11555246e29508081ec9d3b0c566b10f00093 should fix it

ci/conda-env.yml
```diff
@@ -15,6 +15,7 @@ dependencies:
 - gcsfs
 - geopandas
 - graphviz
+- joblib<1.3  # 1.3 is broken for py36
```
ddelange, line 18:
```suggestion
- joblib<1.3; python_version <= '3.6'
```

dunno if its valid syntax in the conda world, but without the marker, people will most likely start running into dephell in their more modern envs

ddelange (approved):

PR #2331, opened 2023-01-20
Support CPython 3.11, 3.12, and aarch64 processors

ddelange:
Hoi 👋

linux-aarch64 makes up for almost 10% of all platforms ref https://github.com/giampaolo/psutil/pull/2103
> aarch64 has already surpassed windows in terms of downloads for this package. Oracle, Amazon, Google, and Microsoft are all offering aarch64 cloud instances at an undeniable price point compared to amd/intel, so the demand will undoubtedly only grow

- this PR is adapted from https://github.com/MagicStack/asyncpg/pull/954
- uses QEMU emulation for linux arm64 wheels: manylinux takes [around 2.5hrs](https://github.com/ddelange/vaex/actions/runs/3965720337/jobs/6795729021) per wheel and alpine arm64 up to 4 hrs 😅
- manylinux2014 wheels are built with GCC 10, which I think does not guarantee proper functioning of pybind11 ([docs](https://pybind11.readthedocs.io/en/stable/compiling.html#configuration-variables)).
  - so with this PR, linux wheels are built with GCC 12 ([`manylinux_2_28`](https://github.com/pypa/manylinux#docker-images)).
  - pip will only install these wheels on linux operating systems with glibc >= 2.28 (mostly all 2020+ linux distributions like debian 10 buster, ubuntu 20.04 focal, almalinux/rhel 8, ...).

#### the wheels from this PR can be installed with:
```sh
# comma separated list for --find-links
export PIP_FIND_LINKS=https://github.com/ddelange/vaex/releases/expanded_assets/core-v4.17.1.post4
pip install --force-reinstall vaex
```

fixes #2366, fixes #2368, fixes #2397, closes #2427, fixes #2384

PR #2356, opened 2023-03-24
Broken Graphql Query conditions

BhaskarChezhiyan:
## Overview
graphQL queries with conditions that have an equal or not equal comparison with 0 as operand value are not evaluated properly. 

## Reproduce
```python

import vaex
import pandas as pd
from decimal import Decimal

df = vaex.from_pandas(pd.Dataframe(data=[[7]], columns=['dummy'] ))

query = """
query($where: BoolExp!, $limit: Int!){
          df(where: $where) { row(limit: $limit){
                               dummy
                               }
                        }
                  }
"""
variables = {'limit': Decimal('10'), 'where': {'dummy': {'_eq': Decimal('0')}}}
output = df.graphql.execute(query, variable_values=variables)

print(output.errors)
```

## Changes
1. Added null check to Compare class _eq and _neq checks.
2. Fixed BooleanCompare class to inherit from Compare class rather than NumberCompare.

### DIFF
--- a/packages/vaex-graphql/vaex/graphql/__init__.py
+++ b/packages/vaex-graphql/vaex/graphql/__init__.py
@@ -72,9 +72,9 @@ def map_to_field(df, name):
 class Compare(graphene.InputObjectType):
     def filter(self, df, name):
         expression = vaex.expression.Expression(df, '(1==1)')
-        if self._eq:
+        if self._eq is not None:
             expression = expression & (df[name] == self._eq)
-        if self._neq:
+        if self._neq is not None:
             expression = expression & (df[name] != self._neq)
         return expression
 
@@ -108,7 +108,7 @@ class FloatCompare(NumberCompare):
     _gte = graphene.Field(graphene.Float)
     _lte = graphene.Field(graphene.Float)
 
-class BooleanCompare(NumberCompare):
+class BooleanCompare(Compare):
     _eq = graphene.Field(graphene.Boolean)
     _neq = graphene.Field(graphene.Boolean)
 


fernando-freitas-alves (approved):

ddelange:
hey 👋 

in #2331 we are encountering an issue with graphql syntax when upgrading to graphene v3: https://github.com/graphql-python/graphene-tornado/commit/a63677c6f463bd082186b77ece450a222e0b7f46

graphene-tornado v2 (currently required by vaex-graphql) pins werkzeug to an incompatible version for python 3.12, forcing us to upgrade to `graphene-tornado==3.0.0b2`, which unpins werkzeug but also requires graphene v3.

do you have any idea how to fix this one?

```pytb
    def test_aggregates(df, schema):
        result = schema.execute("""
        {
            df {
                count
                min {
                    x
                    y
                }
                mean {
                    x
                    y
                }
                max {
                    x
                    y
                }
            }
        }
        """)
>       assert not result.errors
E       assert not [GraphQLError("Aggregation.__init__() got an unexpected keyword argument 'df'", locations=[SourceLocation(line=3, column=9)], path=['df'])]
E        +  where [GraphQLError("Aggregation.__init__() got an unexpected keyword argument 'df'", locations=[SourceLocation(line=3, column=9)], path=['df'])] = ExecutionResult(data={'df': None}, errors=[GraphQLError("Aggregation.__init__() got an unexpected keyword argument 'df'", locations=[SourceLocation(line=3, column=9)], path=['df'])]).errors
```

PR #2436, opened 2024-09-23
fix: avoid timeout due to long blake3 build from source

ddelange:


ddelange:
cc @maartenbreddels saw that master was still flaky

.github/workflows/wheel.yml
```diff
@@ -21,7 +21,7 @@ concurrency:
 
 jobs:
   build_wheels_matrix:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ddelange, line 24: pinning OS versions in `wheel.yml` hoping to get less unexpected failues in CI (due to edits to the runner image like removing libmagic)

ddelange:
@maartenbreddels ready for merge & release

maartenbreddels:
Awesome. Will check if we can do a release now.

PR #2437, opened 2024-09-27
Fix publishing

ddelange:


ddelange:
@maartenbreddels this should do it

ddelange:
after merge, you can re-tag latest master with something like this:
```
git tag -s -f -a "1.0.0" -m "Retag" && git push -f --tags
```

PR #2438, opened 2024-09-30
Update vaex version ranges

ddelange:
awaiting upcoming cp312 support releases ref https://github.com/vaexio/vaex/pull/2331#discussion_r1778800199

this PR also prevents dephell for users who go `pip install vaex` into an existing environment, leading pip to potentially choose older versions of the universal packages

### DIFF
--- a/packages/vaex/setup.py
+++ b/packages/vaex/setup.py
@@ -17,13 +17,13 @@ url = 'https://www.github.com/vaexio/vaex'
 
 install_requires = [
       'vaex-core~=4.18.0',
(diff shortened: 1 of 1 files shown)


ddelange:
cc @maartenbreddels

maartenbreddels:
Ah, I don't think this is needed. I've build this release script (which already noticed vaex-hdf5 requires releasing) which will then auto bump this range. But now I understand what you meant.
image

(Yes, vaex-core should be redone, but locally it looks like it was released)

ddelange:
ah nice!

Issue #2439, opened 2024-10-02
[BUG-REPORT] Hitting `pybind11::handle::inc_ref()` when using NumPy 1.26 with vaex 4.18.0


setu4993:
**Description**

All our unit tests and validations which worked fine on v4.17.1 started failing upon updating to v4.18.0. Curiously, this is occurring only on GitHub Actions so far (`ubuntu-latest` images) and not reproducible on M-series MacBooks. I suspect this is limited to some instruction sets.

I don't have a great sample right now (but can get one tomorrow) to reproduce this, but the CI run [here](https://github.com/rungalileo/dataquality/pull/872) might be helpful to compare the diff. This is occurring across repos but this public one was the easiest to link to.

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 4.18.0
 - Vaex was installed via: pip / conda-forge / from source: `pip`
 - OS: GitHub Actions, `ubuntu-latest`

iisakkirotko:
I think we're running into the same issue in [this run](https://github.com/widgetti/solara/actions/runs/11138736498/job/30955787528) at [widgetti/solara](https://github.com/widgetti/solara). We also see an error on windows: `Windows fatal exception: access violation` resulting in a seg fault.
<details><summary>The full log for Windows</summary>
```bash
Thread 0x00000694 (most recent call first):
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\vaex\hash.py", line 171 in add
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\vaex\cpu.py", line 344 in process
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\vaex\execution.py", line 564 in process_tasks
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\vaex\execution.py", line 500 in process_part
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\vaex\multithreading.py", line 80 in wrapped
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\concurrent\futures\thread.py", line 58 in run
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\concurrent\futures\thread.py", line 83 in _worker
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\threading.py", line 917 in run
[...]

PR #2440, opened 2024-10-07
fix: don't release the gil to early

maartenbreddels:
This is only to check if we now catch the error in CI, the fix will come in a later commit.

See https://github.com/vaexio/vaex/issues/2439

### DIFF
--- a/packages/vaex-core/setup.py
+++ b/packages/vaex-core/setup.py
@@ -121,8 +121,14 @@ else:
     ]
     extra_compile_args.append("-g")
     extra_compile_args += extra_dev_options
+    if os.environ.get("CI"):
+        # this makes sure we catch error in using pybind11, see https://github.com/vaexio/vaex/issues/2439
+        extra_compile_args += ["-UNDEBUG"]
+    else:
+        extra_compile_args += ["-DNDEBUG"]
 if sys.platform == "darwin":
     extra_compile_args.append("-mmacosx-version-min=10.9")
+
 
 
 # on windows (Conda-forge builds), the dirname is an absolute path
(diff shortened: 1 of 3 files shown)


maartenbreddels:
Not sure why the error message is not that clear.

maartenbreddels:
Whow, super happy that this appears to be it.

ddelange (approved):

packages/vaex-core/setup.py
```diff
@@ -121,10 +121,13 @@ def __str__(self):
     ]
     extra_compile_args.append("-g")
     extra_compile_args += extra_dev_options
+    # this makes sure we catch error in using pybind11, see https://github.com/vaexio/vaex/issues/2439
+    extra_compile_args += ["-UNDEBUG"]
```
ddelange, line 125: I wonder if #2441 will override this one, or whether this one is somehow passed alongside the CXXFLAGS (and which behaviour we end up with when both are merged)

ddelange: because this PR causes cibuildwheel to run with `-UNDEBUG`:
```
creating build/temp.linux-x86_64-cpython-38/src
    g++ -pthread -Wl,-strip-all -fPIC -I/tmp/pip-build-env-p52uv4h5/overlay/lib/python3.8/site-packages/numpy/core/include -I/opt/_internal/cpython-3.8.18/include/python3.8 -c src/vaexfast.cpp -o build/temp.linux-x86_64-cpython-38/src/vaexfast.o -std=c++11 -O3 -funroll-loops -Werror=return-type -Wno-unused-parameter -g -UNDEBUG
```
https://github.com/vaexio/vaex/actions/runs/11233046337/job/31225954823?pr=2440#step:7:1403

maartenbreddels: Yes, this is perfect!

PR #2465, opened 2025-09-02
Bump wheel-universal.yml

ddelange:
a different error in `wheel-universal.yml` -- looks like this yml file has not yet been added to the PyPA trusted publishing settings. can you double check?


ddelange:
superceded by #2466

PR #2466, opened 2025-09-02
Fix and prevent CI rot

ddelange:
- dependabot should prevent the CI rot we saw in #2464
- - bump versions (#2465) and add lower pin to s3fs to nudge resolver in the right direction


ddelange:
@maartenbreddels ready for release of the universal packages 👍

PR #2470, opened 2025-10-02
Bump macos-13 to macos-15-intel

ddelange:
https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/


ddelange:
@maartenbreddels by the way did you already add not only `wheel.yml` but also `wheel-universal.yml` as trusted publisher in PyPI? that's the last step before releasing `vaex`, `vaex-ml` etc

ddelange:
@maartenbreddels kind reminder!

ddelange:
> @maartenbreddels by the way did you already add `wheel-universal.yml` as trusted publisher in PyPI for all the `vaex`, `vaex-ml` etc packages? i think until now we had `wheel.yml` as trusted publisher for `vaex-core` only.
> 
> that's the last step before releasing `vaex`, `vaex-ml` etc

@maartenbreddels can you please wrap up the release of python 3.11 and 3.12 support? by now, both 3.13 and 3.14 are already in general availability...

if you need me to support here, im happy to become a maintainer of this github repo (and of the pypi packages if you need me to do the above)

ddelange:
fyi @maartenbreddels this needs to merge for vaex-core builds because macos-13 runner is no longer available since december 2025

maartenbreddels:
Thank you!

PR #2477, opened 2026-02-06
Fix object dtype to string conversion

### LINKED ISSUE #1686: [BUG-REPORT] Issue converting object to string
**Description**
Trying to convert an "object" column to string fails.

Example code:
```python
import vaex
import numpy as np

if __name__ == "__main__":
    arr = np.array([123, "test", None], dtype=object)
    df = vaex.from_arrays(test=arr)
    df['test'] = df['test'].astype('str')
    print(df.head())
```

**Software information**
 - Vaex version (`import vaex; vaex.__version__)`: 
  ```
{'vaex': '4.5.0', 'vaex-core': '4.6.0a5', 'vaex-viz': '0.5.0', 'vaex-hdf5': '0.10.0', 'vaex-server': '0.6.1', 'vaex-astro': '0.9.0', 'vaex-jupyter': '0.6.0', 'vaex-ml': '0.14.0', 'vaex-arrow': '0.4.2'}
```
 - Vaex was installed via: pip
 - OS: Ubuntu 20.04

**Additional information**
This makes it impossible to perform an export to HDF5 or other operations that do not support the object type. This is especially blocking when loading CSV files.
Should be relevant to #1568.

### PR DESCRIPTION
**Pull request by RamiNoodle733**
## Description

This PR fixes the issue where converting object dtype columns to string fails with a TypeError.

When calling `.astype('str')` on an object dtype column (e.g., numpy arrays with mixed types like `[123, "test", None]`), the `to_string()` function would fail because `vaex.strings.to_string()` doesn't have an overload for object dtype arrays.

## Fix

Modified the `to_string()` function in `packages/vaex-core/vaex/functions.py` to handle object dtype arrays by converting them to string arrays using `ar.astype(str)` before passing to `vaex.strings.to_string()`.

## Changes

- Added dtype check for object arrays in `to_string()`
- Convert object arrays to string arrays before processing
- Handles both regular and masked arrays

## Example

```python
import vaex
import numpy as np

# This now works correctly
arr = np.array([123, "test", None], dtype=object)
df = vaex.from_arrays(test=arr)
df['test'] = df['test'].astype('str')  # No longer raises TypeError
```

Fixes #1686

### DIFF
--- a/packages/vaex-core/vaex/functions.py
+++ b/packages/vaex-core/vaex/functions.py
@@ -2407,8 +2407,14 @@ def to_string(x):
     # don't change the dtype, otherwise for each block the dtype may be different (string length)
     ar = vaex.array_types.to_numpy(x)
     if np.ma.isMaskedArray(ar):
+        # Handle object dtype by converting to string array first
+        if ar.dtype == object:
+            ar = ar.astype(str)
         sl = vaex.strings.to_string(ar.data, ar.mask)
     else:
+        # Handle object dtype by converting to string array first
+        if ar.dtype == object:
+            ar = ar.astype(str)
         sl = vaex.strings.to_string(ar)
     return column.ColumnStringArrow.from_string_sequence(sl)
 


packages/vaex-core/vaex/functions.py
```diff
@@ -2407,8 +2407,14 @@ def to_string(x):
...
     ar = vaex.array_types.to_numpy(x)
     if np.ma.isMaskedArray(ar):
+        # Handle object dtype by converting to string array first
+        if ar.dtype == object:
+            ar = ar.astype(str)
         sl = vaex.strings.to_string(ar.data, ar.mask)
     else:
+        # Handle object dtype by converting to string array first
+        if ar.dtype == object:
+            ar = ar.astype(str)
```
ddelange, line 2417: if the same thing happens for both if and else branches, it can be moved out of the if else entirely

PR #2478, opened 2026-02-07
fix: correct typos throughout codebase

thecaptain789:
Fixed common spelling mistakes across 17 files:

- `occured` → `occurred`
- `seperated`/`seperator` → `separated`/`separator`
- `asychronous` → `asynchronous`

All changes are spelling corrections in comments, docstrings, and user-facing strings.


maartenbreddels (changes requested):
Nice, thanks! There is just 1 change that would be breaking, if you tell an agent, it can make it work backwards comparible using kwargs.

